PR #5, opened 2020-01-08
FM2-20: Added telecom translator

corneliouzbett:

Implemented telecom translator

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-20

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
@@ -28,5 +28,8 @@ public class FhirConstants {
 	public static String OPENMRS_FHIR_EXT_NAME = OPENMRS_FHIR_EXT_PREFIX + "/name";
 	
 	public static String PATIENT = "Patient";
+
+	@Value("${project.parent.artifactId}.personAttributeTypeUuid")
+	public static String PERSON_ATTRIBUTE_TYPE_PROPERTY;
 	
 }
--- a/api/src/main/java/org/openmrs/module/fhir2/api/FhirGlobalPropertyService.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/FhirGlobalPropertyService.java
@@ -1,0 +1,18 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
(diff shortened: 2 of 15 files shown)


api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
```diff
@@ -28,5 +28,7 @@ private FhirConstants() {
 	public static String OPENMRS_FHIR_EXT_NAME = OPENMRS_FHIR_EXT_PREFIX + "/name";
 	
 	public static String PATIENT = "Patient";
+
+	public static String PERSON_ATTRIBUTE_TYPE_PROPERTY = "fhir2.personAttributeTypeUuid";
```
ibacher: I'm thinking about this from a maintenance perspective and I think it might be better if we constructed the name of the global property the same everywhere, so here and in `config.xml`. The way to do this would be to add a line to `fhir.properties` like so:

```properties
project.parent.artifactId=@project.parent.artifactId@
```

Then annotating this like this:

```java
@Value("${project.parent.artifactId}.personAttributeTypeUuid")
public static String PERSON_ATTRIBUTE_TYPE_PROPERTY;
```

Does this seem like a good idea to you?

api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
```diff
@@ -28,5 +28,7 @@ private FhirConstants() {
 	public static String OPENMRS_FHIR_EXT_NAME = OPENMRS_FHIR_EXT_PREFIX + "/name";
```
dkayiwa: If these are constants, is there a reason why they are not declared as `final`?

ibacher: This is my fault. The initial "constants" I created were using Spring property injection a la `OPENMRS_FHIR_SERVER_VERSION`, which cannot be declared final. I'll patch these up in a separate PR.

ibacher: This is my fault. The initial constants in this module where all set via Spring property injection, as we're doing for `OPENMRS_FHIR_SERVER_VERSION` and such "constants" can't be declared final for obvious reasons. I'll be patching these up in a separate PR.

api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
corneliouzbett: Yeah. It is ☺️

api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
```diff
@@ -28,5 +28,8 @@ private FhirConstants() {
...
 	
 	public static String PATIENT = "Patient";
+
+	@Value("${project.parent.artifactId}.personAttributeTypeUuid")
+	public static String PERSON_ATTRIBUTE_TYPE_PROPERTY;
```
dkayiwa: How about renaming this class to some thing like `FhirProperties`. Otherwise, looking at constants which can be modified (are not final) makes me think otherwise.

PR #9, opened 2020-01-08
FM2-2 Task Resource CREATE and UPDATE implementation work

pmanko:

Implement CREATE and UPDATE functionality for the FHIR Task resource.

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-2

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/FhirTaskService.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/FhirTaskService.java
@@ -22,7 +22,7 @@ public interface FhirTaskService {
 	 * @param uuid
 	 * @return task with given internal identifier
 	 */
-	public Task getTaskByUuid(String uuid);
+	Task getTaskByUuid(String uuid);
 	
 	/**
 	 * Save task to the DB
@@ -30,6 +30,13 @@ public interface FhirTaskService {
 	 * @param task the task to save
 	 * @return the saved task
 	 */
-	public Task saveTask(Task task);
+	Task saveTask(Task task);
 	
+	/**
+	 * Save task to the DB
+	 * 
+	 * @param task the task to save
+	 * @return the saved task
+	 */
+	Task updateTask(String uuid, Task task);
 }
--- a/api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirTaskServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirTaskServiceImpl.java
@@ -40,9 +40,25 @@ public class FhirTaskServiceImpl implements FhirTaskService {
 	 */
 	@Override
 	public Task saveTask(Task task) {
-		org.openmrs.module.fhir2.Task savedOpenmrsTask = dao.saveTask(translator.toOpenmrsType(task));
+		return translator.toFhirResource(dao.saveTask(translator.toOpenmrsType(task)));
+	}
+	
+	/**
+	 * Save task to the DB, or update task if one exists with given UUID
+	 * 
+	 * @param uuid the uuid of the task to update
+	 * @param task the task to save
+	 * @return the saved task
+	 */
+	@Override
+	public Task updateTask(String uuid, Task task) {
+		org.openmrs.module.fhir2.Task openmrsTask = null;
 		
-		return null;
+		if (uuid != null) {
+			openmrsTask = dao.getTaskByUuid(task.getId());
+		}
+		
+		return translator.toFhirResource(dao.saveTask(translator.toOpenmrsType(openmrsTask, task)));
 	}
 	
 }
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImpl.java
@@ -2,8 +2,6 @@ package org.openmrs.module.fhir2.api.translators.impl;
 
 import lombok.AccessLevel;
 import lombok.Setter;
-import org.hl7.fhir.r4.model.Enumeration;
-import org.hl7.fhir.r4.model.Identifier;
 import org.hl7.fhir.r4.model.Task;
 import org.openmrs.module.fhir2.api.translators.TaskTranslator;
 import org.springframework.stereotype.Component;
@@ -16,10 +14,8 @@ public class TaskTranslatorImpl implements TaskTranslator {
 	public Task toFhirResource(org.openmrs.module.fhir2.Task openmrsTask) {
 		Task fhirTask = new Task();
 		
-		if (openmrsTask != null && openmrsTask.getStatus() != null && openmrsTask.getIntent() != null) {
-			fhirTask.setId(openmrsTask.getUuid());
-			fhirTask.setStatus(Task.TaskStatus.valueOf(openmrsTask.getStatus().name()));
-			fhirTask.setIntent(Task.TaskIntent.valueOf(openmrsTask.getIntent().name()));
+		if (openmrsTask != null) {
+			setFhirTaskFields(openmrsTask, fhirTask);
 		}
 		
 		return fhirTask;
@@ -29,17 +25,45 @@ public class TaskTranslatorImpl implements TaskTranslator {
 	public org.openmrs.module.fhir2.Task toOpenmrsType(Task fhirTask) {
 		org.openmrs.module.fhir2.Task openmrsTask = new org.openmrs.module.fhir2.Task();
 		
-		if (fhirTask != null && fhirTask.getStatus() != null && fhirTask.getIntent() != null) {
-			openmrsTask.setUuid(fhirTask.getId());
-			openmrsTask.setStatus(org.openmrs.module.fhir2.Task.TaskStatus.valueOf(fhirTask.getStatus().name()));
-			openmrsTask.setIntent(org.openmrs.module.fhir2.Task.TaskIntent.valueOf(fhirTask.getIntent().name()));
+		if (fhirTask != null) {
+			setOpenmrsTaskFields(openmrsTask, fhirTask);
 		}
 		
 		return openmrsTask;
 	}
 	
 	@Override
-	public org.openmrs.module.fhir2.Task toOpenmrsType(org.openmrs.module.fhir2.Task existingObject, Task resource) {
-		return null;
+	public org.openmrs.module.fhir2.Task toOpenmrsType(org.openmrs.module.fhir2.Task openmrsTask, Task fhirTask) {
+		if (fhirTask != null) {
+			if (openmrsTask == null) {
+				openmrsTask = new org.openmrs.module.fhir2.Task();
+			}
+			setOpenmrsTaskFields(openmrsTask, fhirTask);
+		}
+		
+		return openmrsTask;
 	}
+	
+	private void setFhirTaskFields(org.openmrs.module.fhir2.Task openmrsTask, Task fhirTask) {
+		fhirTask.setId(openmrsTask.getUuid());
+		if (openmrsTask.getStatus() != null) {
+			fhirTask.setStatus(Task.TaskStatus.valueOf(openmrsTask.getStatus().name()));
+		}
+		if (openmrsTask.getIntent() != null) {
+			fhirTask.setIntent(Task.TaskIntent.valueOf(openmrsTask.getIntent().name()));
+		}
+	}
+	
+	private void setOpenmrsTaskFields(org.openmrs.module.fhir2.Task openmrsTask, Task fhirTask) {
+		if (openmrsTask.getUuid() == null) {
+			openmrsTask.setUuid(fhirTask.getId());
+		}
+		if (fhirTask.getStatus() != null) {
+			openmrsTask.setStatus(org.openmrs.module.fhir2.Task.TaskStatus.valueOf(fhirTask.getStatus().name()));
+		}
+		if (fhirTask.getIntent() != null) {
+			openmrsTask.setIntent(org.openmrs.module.fhir2.Task.TaskIntent.valueOf(fhirTask.getIntent().name()));
+		}
+	}
+	
 }
--- a/api/src/test/java/org/openmrs/module/fhir2/api/impl/FhirTaskServiceImplTest.java
+++ b/api/src/test/java/org/openmrs/module/fhir2/api/impl/FhirTaskServiceImplTest.java
@@ -1,0 +1,126 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.impl;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.when;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+import org.openmrs.module.fhir2.Task;
+import org.openmrs.module.fhir2.api.dao.FhirTaskDao;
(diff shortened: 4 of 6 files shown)


api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirTaskServiceImpl.java
```diff
@@ -40,9 +40,32 @@ public Task getTaskByUuid(String uuid) {
...
+	 * Save task to the DB, or update task if one exists with given UUID
+	 * 
+	 * @param uuid the uuid of the task to update
+	 * @param task the task to save
+	 * @return the saved task
+	 */
+	@Override
+	public Task saveTask(String uuid, Task task) {
+		org.openmrs.module.fhir2.Task openmrsTask;
+		openmrsTask = dao.getTaskByUuid(task.getId());
```
ibacher: What happens if the taskId is not supplied? I expect that to be the case with a new resource (since the id should be provided by the source system and not the calling system). In other words, I think the implementation here should be:

```java
org.openmrs.module.fhir2.Task openmrsTask = null;
if (task.getId() != null) {
    openmrsTask = dao.getTaskByUuid(task.getId());
}
```

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirTaskServiceImpl.java
```diff
@@ -40,9 +40,32 @@ public Task getTaskByUuid(String uuid) {
...
+		
+		if (openmrsTask == null) {
```
ibacher: Maybe we should move this logic to the `toOpenmrsType(openmrsTask, task)` implementation, so that here we can just do:

```java
translator.toOpenmrsType(openmrsTask, task);
```

This is different from how I implemented the other translators, but I actually think it might be a better pattern than throwing an error.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImpl.java
```diff
@@ -29,17 +27,40 @@ public Task toFhirResource(org.openmrs.module.fhir2.Task openmrsTask) {
...
 		}
 		
 		return openmrsTask;
 	}
 	
 	@Override
-	public org.openmrs.module.fhir2.Task toOpenmrsType(org.openmrs.module.fhir2.Task existingObject, Task resource) {
-		return null;
+	public org.openmrs.module.fhir2.Task toOpenmrsType(org.openmrs.module.fhir2.Task openmrsTask, Task fhirTask) {
+		if (fhirTask != null & openmrsTask != null) {
```
ibacher: Basically, with the changes I mentioned above, this should look something like:

```java
if (fhirTask != null) {
    if (openmrsTask == null) {
        openmrsTask = new org.openmrs.module.fhir2.Task();
    }
    setOpenmrsTaskFields(openmrsTask, fhirTask);
}

return openmrsTask;

File: api/src/test/java/org/openmrs/module/fhir2/api/impl/FhirTaskServiceImplTest.java [hunk 4]
```diff
@@ -0,0 +1,122 @@
...
+		org.hl7.fhir.r4.model.Task result = fhirTaskService.getTaskByUuid(TASK_UUID);
+		assertNotNull(result);
+	}
+	
+	@Test
+	public void shouldSaveNewTask() {
+		org.hl7.fhir.r4.model.Task fhirTask = new org.hl7.fhir.r4.model.Task();
+		Task openmrsTask = new Task();
+		
+		fhirTask.setId(TASK_UUID);
```
ibacher: As I mentioned above, I think we should not be supplying the taskId here.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImpl.java
```diff
@@ -29,17 +27,40 @@ public Task toFhirResource(org.openmrs.module.fhir2.Task openmrsTask) {
...
+		if (openmrsTask.getStatus() != null) {
+			fhirTask.setStatus(Task.TaskStatus.valueOf(openmrsTask.getStatus().name()));
+		}
+		if (openmrsTask.getIntent() != null) {
+			fhirTask.setIntent(Task.TaskIntent.valueOf(openmrsTask.getIntent().name()));
+		}
+	}
+	
+	private void setOpenmrsTaskFields(org.openmrs.module.fhir2.Task openmrsTask, Task fhirTask) {
+		openmrsTask.setUuid(fhirTask.getId());
```
ibacher: I would suggest changing this to:

```java
if (openmrsTask.getUuid() == null) {
    openmrsTask.setUuid(fhirTask.getId());
}
```

Since we don't want to allow a FHIR client to change the UUID of an existing task. I should also rethink this in the other translators.

api/src/main/java/org/openmrs/module/fhir2/api/FhirTaskService.java
```diff
@@ -32,4 +32,12 @@
 	 */
 	public Task saveTask(Task task);
 	
+	/**
+	 * Save task to the DB
+	 * 
+	 * @param task the task to save
+	 * @return the saved task
+	 */
+	public Task saveTask(String uuid, Task task);
```
ibacher: We don't need the `public` modifier in an interface. Can you also change this for the other two methods here?

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirTaskServiceImpl.java
```diff
@@ -40,9 +40,31 @@ public Task getTaskByUuid(String uuid) {
 	 */
 	@Override
 	public Task saveTask(Task task) {
-		org.openmrs.module.fhir2.Task savedOpenmrsTask = dao.saveTask(translator.toOpenmrsType(task));
```
ibacher: This can be implemented as a one-liner:

```java
return translator.toFhirResource(dao.saveTask(translator.toOpenmrsType(task)));
```

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirTaskServiceImpl.java
```diff
@@ -40,9 +40,31 @@ public Task getTaskByUuid(String uuid) {
...
+	public Task saveTask(String uuid, Task task) {
+		org.openmrs.module.fhir2.Task openmrsTask = null;
+		
+		if (uuid != null) {
+			openmrsTask = dao.getTaskByUuid(task.getId());
+		}
+		
+		openmrsTask = translator.toOpenmrsType(openmrsTask, task);
+		
+		return translator.toFhirResource(dao.saveTask(openmrsTask));
```
ibacher:
```java
return translator.toFhirResource(dao.saveTask(translator.toOpenmrsType(openmrsTask, task)));
```

api/src/test/java/org/openmrs/module/fhir2/api/impl/FhirTaskServiceImplTest.java
```diff
@@ -0,0 +1,121 @@
...
+	
+	@Before
+	public void setUp() {
+		fhirTaskService = new FhirTaskServiceImpl();
+		fhirTaskService.setDao(dao);
+		fhirTaskService.setTranslator(translator);
+	}
+	
+	@Test
+	public void shouldGetTaskByUuid() {
```
ibacher: I think we'd prefer the "shouldRetrieveTaskByUuid" terminology to stay consistent with other tests.

api/src/test/java/org/openmrs/module/fhir2/api/impl/FhirTaskServiceImplTest.java
```diff
@@ -0,0 +1,121 @@
...
+		org.hl7.fhir.r4.model.Task translatedTask = new org.hl7.fhir.r4.model.Task();
+		
+		task.setUuid(TASK_UUID);
+		translatedTask.setId(TASK_UUID);
+		
+		when(dao.getTaskByUuid(TASK_UUID)).thenReturn(task);
+		when(translator.toFhirResource(task)).thenReturn(translatedTask);
+		
+		org.hl7.fhir.r4.model.Task result = fhirTaskService.getTaskByUuid(TASK_UUID);
+		assertNotNull(result);
```
ibacher: We should also have:

```java
assertEquals(task, result);
```

or (equivalently):

```java
assertThat(result, equalTo(task));
```

api/src/test/java/org/openmrs/module/fhir2/api/impl/FhirTaskServiceImplTest.java
```diff
@@ -0,0 +1,121 @@
...
+		openmrsTask.setStatus(OPENMRS_TASK_STATUS);
+		openmrsTask.setIntent(OPENMRS_TASK_INTENT);
+		
+		when(translator.toOpenmrsType(fhirTask)).thenReturn(openmrsTask);
+		when(dao.saveTask(openmrsTask)).thenReturn(openmrsTask);
+		when(translator.toFhirResource(openmrsTask)).thenReturn(fhirTask);
+		
+		org.hl7.fhir.r4.model.Task result = fhirTaskService.saveTask(fhirTask);
+		
+		assertNotNull(result);
```
ibacher: We should also have:

```java
assertEquals(task, result);
```

or (equivalently):

```java
assertThat(result, equalTo(task));
```

api/src/test/java/org/openmrs/module/fhir2/api/impl/FhirTaskServiceImplTest.java
```diff
@@ -0,0 +1,121 @@
...
+		openmrsTask.setIntent(OPENMRS_TASK_INTENT);
+		
+		when(translator.toOpenmrsType(openmrsTask, fhirTask)).thenReturn(updatedOpenmrsTask);
+		when(dao.saveTask(updatedOpenmrsTask)).thenReturn(updatedOpenmrsTask);
+		when(dao.getTaskByUuid(TASK_UUID)).thenReturn(openmrsTask);
+		when(translator.toFhirResource(updatedOpenmrsTask)).thenReturn(fhirTask);
+		
+		org.hl7.fhir.r4.model.Task result = fhirTaskService.saveTask(TASK_UUID, fhirTask);
+		
+		assertNotNull(result);
```
ibacher: We should also have:

```java
assertEquals(task, result);
```

or (equivalently):

```java
assertThat(result, equalTo(task));
```

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImplTest.java
```diff
@@ -22,10 +23,16 @@
 @RunWith(MockitoJUnitRunner.class)
```
ibacher: Not that it's harmful, but we don't need this if we aren't using the `@Mock` annotation

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImplTest.java
```diff
@@ -12,6 +12,7 @@
 import static org.hamcrest.Matchers.notNullValue;
 import static org.junit.Assert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
```
ibacher: Can we organize these imports so all the hamcrest one's are together followed by the junit one?

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImplTest.java
```diff
@@ -77,4 +84,48 @@ public void shouldTranslateFhirTaskStatusIntentToOpenmrsTaskStatusIntent() {
...
+	public void shouldTranslateNewOpenmrsTask() {
+		Task fhirTask = new Task();
+		fhirTask.setStatus(FHIR_NEW_TASK_STATUS);
+		fhirTask.setIntent(FHIR_TASK_INTENT);
+		
+		org.openmrs.module.fhir2.Task result = taskTranslator.toOpenmrsType(fhirTask);
+		
+		assertThat(result, notNullValue());
+		assertThat(result.getStatus(), equalTo(OPENMRS_NEW_TASK_STATUS));
+		assertThat(result.getUuid(), notNullValue());
```
ibacher: We should also assert the value of intent.

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImplTest.java
```diff
@@ -77,4 +84,48 @@ public void shouldTranslateFhirTaskStatusIntentToOpenmrsTaskStatusIntent() {
...
+	}
+	
+	@Test
+	public void shouldIgnoreUUIDForNewOpenmrsTask() {
+		Task fhirTask = new Task();
+		fhirTask.setStatus(FHIR_NEW_TASK_STATUS);
```
ibacher: These two lines shouldn't be necessary for this test.

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImplTest.java
```diff
@@ -77,4 +84,48 @@ public void shouldTranslateFhirTaskStatusIntentToOpenmrsTaskStatusIntent() {
 		assertThat(result.getStatus(), equalTo(OPENMRS_TASK_STATUS));
 		assertThat(result.getIntent(), equalTo(OPENMRS_TASK_INTENT));
 	}
+	
+	@Test
+	public void shouldTranslateNewOpenmrsTask() {
```
ibacher: We should add a similar test without having a status or intent set on the task.

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImplTest.java
```diff
@@ -77,4 +84,48 @@ public void shouldTranslateFhirTaskStatusIntentToOpenmrsTaskStatusIntent() {
...
+		
+		org.openmrs.module.fhir2.Task result = taskTranslator.toOpenmrsType(fhirTask);
+		
+		assertThat(result, notNullValue());
+		assertThat(result.getStatus(), equalTo(OPENMRS_NEW_TASK_STATUS));
+		assertThat(result.getUuid(), not(equalTo(TASK_UUID)));
+	}
+	
+	@Test
+	public void shouldUpdateExistingOpenmrsTask() {
```
ibacher: We should add a similar test without having a status or intent set on the task.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirTaskServiceImpl.java
```diff
@@ -40,9 +40,25 @@ public Task getTaskByUuid(String uuid) {
...
+	public Task saveTask(String uuid, Task task) {
+		org.openmrs.module.fhir2.Task openmrsTask = null;
 		
-		return null;
+		if (uuid != null) {
+			openmrsTask = dao.getTaskByUuid(task.getId());
+		}
+
+		return translator.toFhirResource(dao.saveTask(translator.toOpenmrsType(openmrsTask, task)));
 	}
```
corneliouzbett: I think this should be **`UpdateTask`** instead
Regardless of using `dao.saveTask()` the method does update the existing Task

api/src/main/java/org/openmrs/module/fhir2/api/FhirTaskService.java
```diff
@@ -22,14 +22,21 @@
...
-	public Task saveTask(Task task);
+	Task saveTask(Task task);
 	
+	/**
+	 * Save task to the DB
+	 * 
+	 * @param task the task to save
+	 * @return the saved task
+	 */
+	Task saveTask(String uuid, Task task);
```
corneliouzbett: You should also change also here

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirTaskServiceImpl.java
```diff
@@ -40,9 +40,25 @@ public Task getTaskByUuid(String uuid) {
...
+	
+	/**
+	 * Save task to the DB, or update task if one exists with given UUID
+	 * 
+	 * @param uuid the uuid of the task to update
+	 * @param task the task to save
+	 * @return the saved task
+	 */
+	@Override
+	public Task updateTask(String uuid, Task task) {
```
jecihjoy: Don't  you guys think that maybe we can stick to the same approach being used in openmrs core and openmrs modules where same save method is used for both save and update?

corneliouzbett: It's okay to have the same save method for saveOrUpdate in the **`DAO`** layer. But we should try to distinguish in the **`service layer`**. Although it's using the same dao method.

PR #15, opened 2020-01-10
FM2-6 Implement the ServiceRequest Resource

pmanko:

Basic implementation of the ServiceRequest resource with simple mappings from Order.Active to ServiceRequest.Status.

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-6

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirServiceRequestDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirServiceRequestDaoImpl.java
@@ -1,0 +1,42 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.dao.impl;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+import lombok.AccessLevel;
+import lombok.Setter;
+import org.hibernate.SessionFactory;
+import org.hibernate.criterion.Restrictions;
+import org.openmrs.OrderType;
+import org.openmrs.TestOrder;
+import org.openmrs.api.OrderService;
+import org.openmrs.module.fhir2.FhirException;
+import org.openmrs.module.fhir2.api.dao.FhirServiceRequestDao;
+import org.springframework.stereotype.Component;
+
+@Component
+@Setter(AccessLevel.PACKAGE)
+public class FhirServiceRequestDaoImpl implements FhirServiceRequestDao {
+	
+	@Inject
+	private OrderService orderService;
+	
+	@Inject
+	@Named("sessionFactory")
+	SessionFactory sessionFactory;
+	
+	@Override
+	public TestOrder getTestOrderByUuid(String uuid) {
+		return (TestOrder) sessionFactory.getCurrentSession().createCriteria(TestOrder.class)
+		        .add(Restrictions.eq("uuid", uuid)).uniqueResult();
+	}
+}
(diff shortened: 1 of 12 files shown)


api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/ServiceRequestTranslatorImplTest.java
```diff
@@ -0,0 +1,71 @@
+package org.openmrs.module.fhir2.api.translators.impl;
+
```
corneliouzbett: Add license header

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ServiceRequestTranslatorImpl.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.Setter;
+import org.hl7.fhir.r4.model.ServiceRequest;
+import org.openmrs.Order;
+import org.openmrs.module.fhir2.api.translators.ServiceRequestTranslator;
+import org.springframework.stereotype.Component;
+
+@Component
+@Setter(AccessLevel.PACKAGE)
+public class ServiceRequestTranslatorImpl implements ServiceRequestTranslator {
+
```
corneliouzbett: Ditto

api/src/test/java/org/openmrs/module/fhir2/api/impl/FhirConceptSourceServiceImplTest.java
```diff
@@ -65,44 +65,44 @@ public void getFhirConceptSources_shouldReturnEmptyCollectionWhenNoSourcesFound(
...
 	public void getFhirConceptSourceByConceptSourceName_shouldReturnEmptyWhenNoConceptSourceFound() {
 		when(dao.getFhirConceptSourceByConceptSourceName("LOINC")).thenReturn(Optional.empty());
-
+		
 		Optional<FhirConceptSource> result = fhirConceptSourceService.getFhirConceptSourceByConceptSourceName("LOINC");
-
+		
 		assertThat(result.isPresent(), is(false));
 	}
 }
```
corneliouzbett: Remove this file from your commit.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirServiceRequestDaoImpl.java
```diff
@@ -0,0 +1,41 @@
...
+public class FhirServiceRequestDaoImpl implements FhirServiceRequestDao {
+	
+	@Inject
+	private OrderService orderService;
+	
+	@Override
+	public TestOrder getTestOrderByUuid(String uuid) {
+		Order result = orderService.getOrderByUuid(uuid);
+		
+		if (result != null && OrderType.TEST_ORDER_TYPE_UUID.equals(result.getOrderType().getUuid())) {
```
ibacher: Thinking about this, I think the better implementation is to do this:

```java
if (result == null) {
   return null;
}
else if (Order.TEST_ORDER_TYPE_UUID.equals(result.getOrderType().getUuid()))) {
    return (TestOrder) result;
}
else {
    throw new FhirException("UUID is not a valid Test Order");
}
```

But we might also get the correct behaviour much more simply by doing this:

```java
return (TestOrder) sessionFactory.getCurrentSession().createCriteria(TestOrder.class).add(Restrictions.eq("uuid", uuid)).uniqueResult();
```

Though I'm not 100% sure that would work given the inheritance.

pmanko: the one-liner would not throw a `FhirException`, and we would have the same behavior (null value) for an order a wrong type as a not-found UUID, correct?

pmanko: also, what should guide the decision between using the `Service` class functionality vs. using `Hibernate` directly?

ibacher: Yes, it would not throw the FhirException, but that's ok because we're limiting things to TestOrders rather than the whole order table.

My thinking is that we should use the OpenMRS service layer as little as possible. This can get complicated though. For example, in the FhirPatientDaoImpl, we rely on the `patientService` because the logic there is hard to duplicate, especially in more recent versions of the platform which use Lucene for full text searching. Similarly, in the FhirConceptDaoImpl, I used the conceptService because the `getConceptsByMapping()` function is a bit convoluted and I wanted to make sure it was done correctly.

ibacher:
@pmanko Can you add the license header to FhirServiceRequestDao?

PR #33, opened 2020-01-21
FM2-10: Initial implementation for encounter resource

corneliouzbett:

Initial work on encounter resource.
Properties implemented:

- Id
- status
- Subject: (patient)

## Issue I worked on

see https://issues.openmrs.org/projects/FM2/issues/FM2-10

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
@@ -1,0 +1,35 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.dao.impl;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+import lombok.AccessLevel;
+import lombok.Setter;
+import org.hibernate.SessionFactory;
+import org.openmrs.Encounter;
+import org.openmrs.module.fhir2.api.dao.FhirEncounterDao;
+import org.springframework.stereotype.Component;
+import static org.hibernate.criterion.Restrictions.eq;
+
+@Component
+@Setter(AccessLevel.PACKAGE)
+public class FhirEncounterDaoImpl implements FhirEncounterDao {
+
+	@Inject
+	@Named("sessionFactory")
+	private SessionFactory sessionFactory;
+	
+	@Override
+	public Encounter getEncounterByUuid(String uuid) {
+		return (Encounter) sessionFactory.getCurrentSession().createCriteria(Encounter.class).add(eq("uuid", uuid)).uniqueResult();
+	}
+}
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
@@ -23,13 +23,14 @@ import static org.hibernate.criterion.Restrictions.eq;
 @Component
 @Setter(AccessLevel.PACKAGE)
 public class FhirEncounterDaoImpl implements FhirEncounterDao {
-
+	
 	@Inject
 	@Named("sessionFactory")
 	private SessionFactory sessionFactory;
 	
 	@Override
 	public Encounter getEncounterByUuid(String uuid) {
-		return (Encounter) sessionFactory.getCurrentSession().createCriteria(Encounter.class).add(eq("uuid", uuid)).uniqueResult();
+		return (Encounter) sessionFactory.getCurrentSession().createCriteria(Encounter.class).add(eq("uuid", uuid))
+		        .uniqueResult();
 	}
 }
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImpl.java
@@ -1,0 +1,64 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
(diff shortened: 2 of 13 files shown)


api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
```diff
@@ -0,0 +1,32 @@
...
+@Component
+@Setter(AccessLevel.PACKAGE)
+public class FhirEncounterDaoImpl implements FhirEncounterDao {
+	
+	@Inject
+	private EncounterService encounterService;
+	
+	@Override
+	public Encounter getEncounterByUuid(String uuid) {
+		return encounterService.getEncounterByUuid(uuid);
```
ibacher: I tend to be in favour of the `sessionFactory.getCurrentSession().createCriteria(Encounter.class).add(eq("uuid", uuid)).uniqueResult()` pattern.

api/src/main/java/org/openmrs/module/fhir2/api/util/FhirUtils.java
```diff
@@ -0,0 +1,52 @@
...
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import org.apache.commons.lang.StringUtils;
+import org.hl7.fhir.r4.model.Identifier;
+import org.hl7.fhir.r4.model.Reference;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class FhirUtils {
+	
+	public static Reference addReference(@NotNull String uri, String givenName, String familyName, String identifier) {
```
ibacher: This should probably be `addPersonReference`, given its reliance on name. Also, shouldn't we just pass the UUID in here?

api/src/main/java/org/openmrs/module/fhir2/api/util/FhirUtils.java
```diff
@@ -0,0 +1,52 @@
...
+		Reference reference = new Reference();
+		Identifier fhirIdentifier = new Identifier();
```
ibacher: I don't think we should be populating the identifier field, as it's primarily intended for [logical references](https://www.hl7.org/fhir/references.html#logical), where as this is a proper resource reference.

api/src/main/java/org/openmrs/module/fhir2/api/util/FhirUtils.java
```diff
@@ -0,0 +1,52 @@
...
+		fhirIdentifier.setValue(identifier);
+		
+		StringBuilder displayName = new StringBuilder();
```
ibacher: Instead of doing the formatting here, we should just pass the result of `PersonName#getFullName()` into this method and use that as the display name.

api/src/main/java/org/openmrs/module/fhir2/api/util/FhirUtils.java
```diff
@@ -0,0 +1,52 @@
...
+			displayName.append(" ");
+		}
+		if (StringUtils.isNotBlank(familyName)) {
+			displayName.append(familyName);
+		}
+		
+		reference.setReference(uri);
+		reference.setDisplay(displayName.toString());
+		reference.setIdentifier(fhirIdentifier);
+		reference.setId(extractUuid(uri));
```
ibacher: The reference Id should be left unset, so that it's set at run-time. Conceptually, this UUID is specific to the Patient resource and the reference to the patient resource is an entirely different object.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImpl.java
```diff
@@ -0,0 +1,71 @@
...
+			return existingEncounter;
+		}
+		existingEncounter.setUuid(encounter.getId());
+		String patientUuid = FhirUtils.extractUuid(encounter.getSubject().getReference());
+		existingEncounter.setPatient(patientService.getPatientByUuid(patientUuid));
+		
+		return existingEncounter;
+	}
+	
+	public Reference getPatientReference(Patient patient) {
```
ibacher: I would move this to the FhirUtils class. We will need it in many places.

api/src/test/java/org/openmrs/module/fhir2/FhirTestConstants.java
```diff
@@ -14,4 +14,7 @@
 	public static final String LOINC_SYSTEM_URL = "http://loinc.org";
 	
 	public static final String CIEL_SYSTEM_URN = "urn:oid:2.16.840.1.113883.3.7201";
+	
+	public static final String PATIENT = "Patient";
```
ibacher: Just use the value from FhirConstants. The purpose of FhirTestConstants is for some constants needed across tests that aren't constants in the application, e.g., the `LOINC_SYSTEM_URL` or `CIEL_SYSTEM_URN` are actually determined at runtime from the database, but for testing purposes, we can use fixed values.

api/src/main/java/org/openmrs/module/fhir2/api/util/FhirUtils.java
```diff
@@ -0,0 +1,52 @@
+/*
```
ibacher: We should probably consider renaming either this file to `FhirReferenceUtils` or the file in the omod to `FhirServerUtils` (I favour the second of these) just to avoid confusion.

api/src/test/java/org/openmrs/module/fhir2/FhirTestConstants.java
corneliouzbett: 👍🏽

api/src/main/java/org/openmrs/module/fhir2/api/util/FhirUtils.java
corneliouzbett: Probably then should just have `addPatientReference(org.openmrs.Patient patient)`

ibacher: That would also make sense!

api/src/main/java/org/openmrs/module/fhir2/api/util/FhirReferenceUtils.java
```diff
@@ -0,0 +1,47 @@
...
+public class FhirReferenceUtils {
+
+	
+	public static Reference addPatientReference(Patient patient) {
+		Reference reference = new Reference();
+		String patientUri = FhirConstants.PATIENT +"/" +patient.getUuid();
+		reference.setReference(patientUri);
+		String nameDisplay = patient.getPersonName().getFullName()
+				+ "("
+				+ FhirConstants.IDENTIFIER
```
ibacher: I think instead of the IDENTIFIER constant, we should use `patient.getPatientIdentifier().getIdentifier().getIdentifierType().getName()`; the reason is, not all implementations use only a single identifier per patient.

corneliouzbett: You're right. Didn't thought of that.

PR #60, opened 2020-02-10
Added mappings for Observation.interpretation and Observation.status for 2.1.0 and newer

jecihjoy:

I added a mapping for Observation.interpretation and Observation.status for 2.1.0 and newer 

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-53

  No? -> execute above command

### DIFF
--- a/api-2.1/pom.xml
+++ b/api-2.1/pom.xml
@@ -1,0 +1,69 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>fhir2</artifactId>
+        <groupId>org.openmrs.module</groupId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>fhir2-api-2.1</artifactId>
+    <packaging>jar</packaging>
+    <name>FHIR2 API 2.1</name>
+    <description>API 2.1 for FHIR2</description>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.parent.groupId}</groupId>
+            <artifactId>${project.parent.artifactId}-api</artifactId>
+            <version>${project.parent.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.openmrs.api</groupId>
+            <artifactId>openmrs-api</artifactId>
+            <version>${openmrsPlatformVersion}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.inject</groupId>
+            <artifactId>javax.inject</artifactId>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>com.mycila</groupId>
+                <artifactId>license-maven-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.projectlombok</groupId>
+                <artifactId>lombok-maven-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>net.revelc.code.formatter</groupId>
+                <artifactId>formatter-maven-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>net.revelc.code</groupId>
+                <artifactId>impsort-maven-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
+
+    <properties>
+        <openmrsPlatformVersion>2.3.0</openmrsPlatformVersion>
+    </properties>
+</project>
+
--- a/api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationInterpretationTranslatorImpl_2_1.java
+++ b/api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationInterpretationTranslatorImpl_2_1.java
@@ -1,0 +1,184 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.translators.impl;
+
+import org.apache.commons.lang.WordUtils;
+import org.hl7.fhir.r4.model.CodeableConcept;
+import org.hl7.fhir.r4.model.Coding;
+import org.openmrs.Obs;
+import org.openmrs.annotation.OpenmrsProfile;
+import org.openmrs.module.fhir2.FhirConstants;
+import org.springframework.context.annotation.Primary;
+import org.springframework.stereotype.Component;
+
+@Primary
+@Component
+@OpenmrsProfile(openmrsPlatformVersion = "2.1.* - 2.3.*")
+public class ObservationInterpretationTranslatorImpl_2_1 extends ObservationInterpretationTranslatorImpl {
+	
+	@Override
+	public CodeableConcept toFhirResource(Obs obs) {
+		if (obs.getInterpretation() == null) {
+			return null;
+		}
+		CodeableConcept interpretation = new CodeableConcept();
+		switch (obs.getInterpretation()) {
+			case NORMAL:
+				interpretation.addCoding(addIntepretationCoding("N", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case ABNORMAL:
+				interpretation.addCoding(addIntepretationCoding("A", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case CRITICALLY_ABNORMAL:
+				interpretation.addCoding(addIntepretationCoding("AA", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case CRITICALLY_HIGH:
+				interpretation.addCoding(addIntepretationCoding("HH", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case CRITICALLY_LOW:
+				interpretation.addCoding(addIntepretationCoding("LL", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case HIGH:
+				interpretation.addCoding(addIntepretationCoding("H", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case LOW:
+				interpretation.addCoding(addIntepretationCoding("L", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case OFF_SCALE_LOW:
+				interpretation.addCoding(addIntepretationCoding("<", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case OFF_SCALE_HIGH:
+				interpretation.addCoding(addIntepretationCoding(">", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case SIGNIFICANT_CHANGE_DOWN:
+				interpretation.addCoding(addIntepretationCoding("D", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case SIGNIFICANT_CHANGE_UP:
+				interpretation.addCoding(addIntepretationCoding("U", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case RESISTANT:
+				interpretation.addCoding(addIntepretationCoding("R", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case SUSCEPTIBLE:
+				interpretation.addCoding(addIntepretationCoding("S", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case INTERMEDIATE:
+				interpretation.addCoding(addIntepretationCoding("I", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case POSITIVE:
+				interpretation.addCoding(addIntepretationCoding("POS", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case NEGATIVE:
+				interpretation.addCoding(addIntepretationCoding("NEG", obs.getInterpretation()));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			case VERY_SUSCEPTIBLE:
+				interpretation.addCoding(new Coding().setSystem(FhirConstants.OPENMRS_FHIR_EXT_VS_INTERPRETATION)
+				        .setCode("VS").setDisplay("Very Susceptible"));
+				setText(interpretation, obs.getInterpretation());
+				break;
+			default:
+				return interpretation;
+			
+		}
+		return interpretation;
+	}
+	
+	@Override
+	public Obs toOpenmrsType(Obs openmrsObs, CodeableConcept interpretation) {
+		if (interpretation.getCoding().size() == 0) {
+			return null;
+		}
+		switch (interpretation.getCoding().get(0).getCode()) {
+			case "N":
+				openmrsObs.setInterpretation(Obs.Interpretation.NORMAL);
+				break;
+			case "A":
+				openmrsObs.setInterpretation(Obs.Interpretation.ABNORMAL);
+				break;
+			case "AA":
+				openmrsObs.setInterpretation(Obs.Interpretation.CRITICALLY_ABNORMAL);
+				break;
+			case "HH":
+				openmrsObs.setInterpretation(Obs.Interpretation.CRITICALLY_HIGH);
+				break;
+			case "LL":
+				openmrsObs.setInterpretation(Obs.Interpretation.CRITICALLY_LOW);
+				break;
+			case "H":
+				openmrsObs.setInterpretation(Obs.Interpretation.HIGH);
+				break;
+			case "L":
+				openmrsObs.setInterpretation(Obs.Interpretation.LOW);
+				break;
+			case "<":
+				openmrsObs.setInterpretation(Obs.Interpretation.OFF_SCALE_LOW);
+				break;
+			case ">":
+				openmrsObs.setInterpretation(Obs.Interpretation.OFF_SCALE_HIGH);
+				break;
+			case "D":
+				openmrsObs.setInterpretation(Obs.Interpretation.SIGNIFICANT_CHANGE_DOWN);
+				break;
+			case "U":
+				openmrsObs.setInterpretation(Obs.Interpretation.SIGNIFICANT_CHANGE_UP);
+				break;
+			case "R":
+				openmrsObs.setInterpretation(Obs.Interpretation.RESISTANT);
+				break;
+			case "S":
+				openmrsObs.setInterpretation(Obs.Interpretation.SUSCEPTIBLE);
+				break;
+			case "I":
+				openmrsObs.setInterpretation(Obs.Interpretation.INTERMEDIATE);
+				break;
+			case "POS":
+				openmrsObs.setInterpretation(Obs.Interpretation.POSITIVE);
+				break;
+			case "NEG":
+				openmrsObs.setInterpretation(Obs.Interpretation.NEGATIVE);
+				break;
+			case "VS":
+				openmrsObs.setInterpretation(Obs.Interpretation.VERY_SUSCEPTIBLE);
+				break;
+			default:
+				openmrsObs.setInterpretation(null);
+				break;
+		}
+		return openmrsObs;
+	}
+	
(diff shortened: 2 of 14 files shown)


api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationTranslatorImpl_2_1.java
```diff
@@ -0,0 +1,54 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.translators.impl;
```
ibacher: I don't *think* we actually need a custom version of the `ObservationTranslator` here, since the default version still calls the `ObservationStatusTranslator`. We just need to ensure that is properly overridden.

pom.xml
```diff
@@ -38,6 +38,7 @@
 		<module>api</module>
         <module>api-2.0</module>
 		<module>omod</module>
+        <module>api-2.1</module>
```
ibacher: Small thing, but this should go under the `api-2.0` line.

api-2.1/src/main/java/org/openmrs/module/fhir2/api/impl/FhirObservationServiceImpl_2_1.java
```diff
@@ -0,0 +1,46 @@
+/**
```
ibacher: I also don't think we need to implement this class for this change.

api-2.1/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+            <groupId>javax.inject</groupId>
+            <artifactId>javax.inject</artifactId>
+        </dependency>
+    </dependencies>
+
+    <properties>
+    <openmrsPlatformVersion>2.3.0</openmrsPlatformVersion>
+    </properties>
+
+
```
ibacher: Could you remove the blank space here?

api-2.1/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+            <artifactId>lombok</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.inject</groupId>
+            <artifactId>javax.inject</artifactId>
+        </dependency>
+    </dependencies>
+
+    <properties>
+    <openmrsPlatformVersion>2.3.0</openmrsPlatformVersion>
```
ibacher: Indent this 1 more tab

api-2.1/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+    </properties>
+
+
+</project>
```
ibacher: Add an empty new line at the end of the file

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationStatusTranslatorImpl_2_1.java
```diff
@@ -0,0 +1,37 @@
...
+public class ObservationStatusTranslatorImpl_2_1 extends ObservationStatusTranslatorImpl {
+
+    @Override
+    public Observation.ObservationStatus toFhirResource(Obs obs) {
+        return Observation.ObservationStatus.valueOf(obs.getStatus().toString());
+    }
+
+    @Override
+    public Obs toOpenmrsType(Obs observation, Observation.ObservationStatus observationStatus) {
+        Obs obs = super.toOpenmrsType(observation, observationStatus);
```
ibacher: We don't need to do this. This implementation can simply be

```java
observation.setStatus(Obs.Status.valueOf(observationStatus.toString()));
return observation;
```

ibacher:
@jecihjoy I know this is only a draft PR, but it's looking good so far!

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationTranslatorImpl_2_1.java
jecihjoy: hey Ian, this might work with status but what about translating the interpretations and it's not part of the origin implementation with platform 2.0.5

omod/pom.xml
```diff
@@ -23,6 +23,11 @@
 			<artifactId>${project.parent.artifactId}-api-2.0</artifactId>
 			<version>${project.version}</version>
 		</dependency>
+		<dependency>
+			<groupId>org.openmrs.module</groupId>
+			<artifactId>fhir2-api-2.1</artifactId>
+			<version>1.0.0-SNAPSHOT</version>
+		</dependency>
```
corneliouzbett: Change the above to:

```java
        <dependency>
            <groupId>${project.parent.groupId}</groupId>
            <artifactId>${project.parent.artifactId}-api-2.1</artifactId>
            <version>${project.parent.version}</version>
        </dependency>
```

api-2.1/src/test/java/org/openmrs/module/fhir2/api/translators/impl/ObservationStatusTranslatorImpl_2_1Test.java
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.openmrs.Obs;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+
+public class ObservationStatusTranslatorImpl_2_1Test {
+
```
corneliouzbett: Can you add license header to this file

api-2.1/src/test/java/org/openmrs/module/fhir2/api/translators/impl/ObservationInterpretationTranslatorImpl_2_1Test.java
```diff
@@ -0,0 +1,55 @@
...
+import org.mockito.junit.MockitoJUnitRunner;
+import org.openmrs.Obs;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ObservationInterpretationTranslatorImpl_2_1Test {
```
corneliouzbett: Can you add license header to this class

api-2.1/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+            <version>${openmrsPlatformVersion}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.openmrs.test</groupId>
+            <artifactId>openmrs-test</artifactId>
+            <version>${openmrsPlatformVersion}</version>
+            <type>pom</type>
+            <scope>test</scope>
+        </dependency>
```
corneliouzbett: Do you really need this dependencies? I think this should work without them.

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/ObservationInterpretationTranslator.java
```diff
@@ -0,0 +1,24 @@
+package org.openmrs.module.fhir2.api.translators;
+
+import org.hl7.fhir.r4.model.CodeableConcept;
+import org.openmrs.Obs;
+
+public interface ObservationInterpretationTranslator extends ToFhirTranslator<Obs, CodeableConcept>, UpdatableOpenmrsTranslator<Obs, CodeableConcept> {
```
corneliouzbett: Add license header to this class

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationInterpretationTranslatorImpl_2_1.java
```diff
@@ -0,0 +1,41 @@
...
+import org.hl7.fhir.r4.model.CodeableConcept;
+import org.openmrs.Obs;
+import org.openmrs.annotation.OpenmrsProfile;
+import org.openmrs.module.fhir2.api.translators.ObservationInterpretationTranslator;
+import org.springframework.context.annotation.Primary;
+import org.springframework.stereotype.Component;
+
+@Primary
+@Component
+@Setter(AccessLevel.PACKAGE)
```
corneliouzbett: Remove this line (Not used)

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationStatusTranslatorImpl_2_1.java
```diff
@@ -0,0 +1,36 @@
...
+import lombok.Setter;
+import org.hl7.fhir.r4.model.Observation;
+import org.openmrs.Obs;
+import org.openmrs.annotation.OpenmrsProfile;
+import org.springframework.context.annotation.Primary;
+import org.springframework.stereotype.Component;
+
+@Primary
+@Component
+@Setter(AccessLevel.PACKAGE)
```
corneliouzbett: Remove this line (Not used at this point)

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/ObservationInterpretationTranslator.java
```diff
@@ -0,0 +1,24 @@
...
+    CodeableConcept toFhirResource(Obs obs);
+
+    /**
+     * Maps an {@link org.hl7.fhir.r4.model.CodeableConcept} to an {@link Obs}
+     * @param existingObject the existingObject to update
+     * @param resource the resource to map
+     * @return
+     */
+    @Override
+    Obs toOpenmrsType(Obs existingObject, CodeableConcept resource);
```
corneliouzbett: I prefer **`existingObs`** instead of **`existingObject`**

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationInterpretationTranslatorImpl_2_1.java
corneliouzbett: > Remove this line (Not used)

referring to `@Setter(AccessLevel.PACKAGE)`

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationTranslatorImpl_2_1.java
ibacher: I think the solution there is to add an translator for interpretations similar to the dummy translator for observation status we have. That way, it's easy for implementors to support interpretations if they're currently capturing that information.

jecihjoy:
Thanks for the comments, check the updated code

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationTranslatorImpl_2_1.java
jecihjoy: That makes sense, I'll look into it

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationInterpretationTranslatorImpl_2_1.java
```diff
@@ -0,0 +1,38 @@
...
+@Primary
+@Component
+@OpenmrsProfile(openmrsPlatformVersion = "2.1.* - 2.3.*")
+public class ObservationInterpretationTranslatorImpl_2_1 implements ObservationInterpretationTranslator {
+	
+	@Override
+	public CodeableConcept toFhirResource(Obs obs) {
+		CodeableConcept interpretation = new CodeableConcept();
+		if (obs.getInterpretation() != null) {
+			interpretation.setText(obs.getInterpretation().toString());
```
ibacher: So what we actually want to do here is to map not to text but to the coding using the code set defined [here](https://www.hl7.org/fhir/valueset-observation-interpretation.html). So, for example, `Obs.Interpretation.NORMAL` should map to a structure that looks like this:

```json
"interpretation": {
	"coding": [
		{
			"system": "http://hl7.org/fhir/ValueSet/observation-interpretation",
			"code": "N",
			"display": "Normal
		}
	],
	"text": "Normal"
}
```

Code-wise this looks something like (with the constants appropriately extracted out):

```java
switch(obs.getInterpretation()) {
    case NORMAL:
		interpretation.addCoding().setSystem("http://hl7.org/fhir/ValueSet/observation-interpretation").setCode("N").setDisplay("Normal");
		interpretation.setText("Normal");
	...
}
```

The mappings should be something like:

|OpenMRS Type|FHIR Interpretation|
|-------|------|
|`NORMAL`|`N`|
|`ABNORMAL`|`A`|
|`CRITICALLY_ABNORMAL`|`AA`|
|`CRITICALLY_HIGH`|`HH`|
|`CRITICALLY_LOW`|`LL`|
|`HIGH`|`H`|
|`LOW`|`L`|
|`OFF_SCALE_LOW`|`<`|
|`OFF_SCALE_HI`|`>`|
|`SIGNIFICANT_CHANGE_DOWN`|`D`|
|`SIGNIFICANT_CHANGE_UP`|`U`|
|`RESISTANT`|`R`|
|`SUSCEPTIBLE`|`S`|
|`INTERMEDIATE`|`I`|
|`POSITIVE`|`POS`|
|`NEGATIVE`|`NEG`|

The display values should be those in the valueset I linked above.

Some special handling is needed for the `VERY_SUSCEPTIBLE` interpretation, since this isn't supported by the valueset. For this case, it might make sense to add the coding for susceptible and then create an additional code with the system `https://fhir.openmrs.org/ext/obs/interpretations`, the code `VS`, and the display "Very Susceptible", which should also be the reported text.

jecihjoy: This is noted, working on it..

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationInterpretationTranslatorImpl_2_1.java
```diff
@@ -0,0 +1,124 @@
...
+	
+	@Override
+	public CodeableConcept toFhirResource(Obs obs) {
+		if (obs.getInterpretation() == null) {
+			return null;
+		}
+		CodeableConcept interpretation = new CodeableConcept();
+		switch (obs.getInterpretation()) {
+			case NORMAL:
+				interpretation.setCoding(Collections.singletonList(addIntepretationCoding("N", obs.getInterpretation())));
```
ibacher: Use `addCoding()` instead of `setCoding()`. The former allows us to have multiple codings where it makes sense.

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationInterpretationTranslatorImpl_2_1.java
```diff
@@ -0,0 +1,124 @@
...
+			default:
+				return interpretation;
+			
+		}
+		return interpretation;
+	}
+	
+	@Override
+	public Obs toOpenmrsType(Obs openmrsObs, CodeableConcept interpretation) {
+		openmrsObs.setInterpretation(Obs.Interpretation.valueOf(interpretation.getText()));
```
ibacher: This actually needs to invert the `toFhirResource()` method. So here we expect a coding of, e.g., "NEG" and want to translate that into `Obs.Interpretation.NEGATIVE`

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationInterpretationTranslatorImpl_2_1.java
```diff
@@ -0,0 +1,124 @@
...
+	@Override
+	public CodeableConcept toFhirResource(Obs obs) {
+		if (obs.getInterpretation() == null) {
+			return null;
+		}
+		CodeableConcept interpretation = new CodeableConcept();
+		switch (obs.getInterpretation()) {
+			case NORMAL:
+				interpretation.setCoding(Collections.singletonList(addIntepretationCoding("N", obs.getInterpretation())));
+				interpretation.setText(obs.getInterpretation().toString().toLowerCase());
```
ibacher: I think we should do the same thing here that we do to translate the interpretation into it's display value.

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationStatusTranslatorImpl_2_1.java
```diff
@@ -0,0 +1,33 @@
...
+public class ObservationStatusTranslatorImpl_2_1 extends ObservationStatusTranslatorImpl {
+	
+	@Override
+	public Observation.ObservationStatus toFhirResource(Obs obs) {
+		return Observation.ObservationStatus.valueOf(obs.getStatus().toString());
+	}
+	
+	@Override
+	public Obs toOpenmrsType(Obs observation, Observation.ObservationStatus observationStatus) {
+		observation.setStatus(Obs.Status.valueOf(observationStatus.toString()));
```
ibacher: The FHIR observation status set is substantially larger than the OpenMRS observation status. What happens, for example, if we get an observation with the status "cancelled"? (I'd suggest writing a unit test that passes that status and try to fix the issue.

api-2.1/src/test/java/org/openmrs/module/fhir2/api/translators/impl/ObservationInterpretationTranslatorImpl_2_1Test.java
```diff
@@ -0,0 +1,233 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+import org.openmrs.Obs;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ObservationInterpretationTranslatorImpl_2_1Test {
+	
+	private static final Obs.Interpretation OBS_INTERPRETATION = Obs.Interpretation.HIGH;
+	
+	ObservationInterpretationTranslatorImpl_2_1 observationInterpretationTranslator;
```
ibacher: Should be `private` (I'm bad about remembering this too)

api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
```diff
@@ -21,6 +21,10 @@
 	
 	public static final String DIAGNOSTIC_SERVICE_SECTIONS_VALUESET_URI = "http://hl7.org/fhir/ValueSet/diagnostic-service-sections";
 	
+	public static final String INTERPRETATION_VALUE_SET_URI = "http://hl7.org/fhir/ValueSet/observation-interpretation";
+	
+	public static final String VERY_SUSCEPTIBLE_SYSTEM_URI = "https://fhir.openmrs.org/ext/obs/interpretations";
```
ibacher: Can we construct this as we construct the other extension URLs? That way if we need to change something, we only have to change it in one place.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationTranslatorImpl.java
```diff
@@ -80,6 +85,8 @@ public Observation toFhirResource(Obs observation) {
 		
 		obs.setValue(observationValueTranslator.toFhirResource(observation));
 		
+		obs.setInterpretation(Collections.singletonList(interpretationTranslator.toFhirResource(observation)));
```
ibacher: `addInterpretation()` rather than `setInterpretation()` here.

jecihjoy: OpenMrs only allows adding a single interpretation to an observation, **addInterpretation()** is not supported

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationInterpretationTranslatorImpl_2_1.java
jecihjoy: Do we need to use **addCoding()** yet OpenMrs only supports a single interpretation per an Observation?

jecihjoy: Also while doing the translations back to openmrs type, I suppose this would be the code

**switch (interpretation.getCoding().get(0).getCode()){
case "N":
	openmrsObs.setInterpretation(Obs.Interpretation.NEGATIVE);
	break;**

well it will not make sense to check the value at 0 if we're expecting more than one integration..

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationTranslatorImpl.java
ibacher: So, I'm not suggesting we support multiple interpretations, just that `obs.addInterpretation(interpretationTranslator.toFhirResource(observation))` does the same thing and is a bit shorter and clearer.

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationInterpretationTranslatorImpl_2_1.java
ibacher: > @ibacher do we need to use **addCoding()** yet OpenMrs only supports a single interpretation to an Observation?

So, `addCoding()` is basically a simpler way of achieving the same result without having to construct a list here.

ibacher: > Also while doing the translations back to openmrs type, I suppose this would be the code
> 
> **switch (interpretation.getCoding().get(0).getCode()){ case "N": openmrsObs.setInterpretation(Obs.Interpretation.NEGATIVE); break;**
> 
> well it will not make sense to check the value at 0 if we're expecting more than one interpretation..

I'm not anticipating more than one interpretation here, so yes, just mapping from "N" to NEGATIVE is the right way to go.

jecihjoy: i understand your perspective now

jecihjoy:
Thanks @ibacher 😄

PR #77, opened 2020-02-20
FM2-75 Address gaps in supported elements for DiagnosticReport, Task, and ServiceRequest 

pmanko:

Implemented support for elements required for the Lab workflow in Task, DiagnosticReport, and ServiceRequest.

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-75

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/FhirReference.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/FhirReference.java
@@ -1,0 +1,50 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
+import org.openmrs.BaseOpenmrsMetadata;
+
+/**
+ * FHIR Reference - https://www.hl7.org/fhir/references.html
+ */
+@Data(staticConstructor = "of")
+@NoArgsConstructor
+@EqualsAndHashCode(callSuper = true)
+@Entity
+@Table(name = "fhir_reference")
+public class FhirReference extends BaseOpenmrsMetadata {
+	
+	private static final long serialVersionUID = 1L;
+	
+	@Id
+	@GeneratedValue(strategy = GenerationType.AUTO)
+	@Column(name = "reference_id")
+	private Integer id;
+	
+	@Column(name = "target_type")
+	private String type;
+	
+	@Column(name = "reference")
+	private String reference;
+	
+	public void setName() {
+		this.setName(id + "/" + type);
+	}
+}
--- a/api/src/main/java/org/openmrs/module/fhir2/FhirTask.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/FhirTask.java
@@ -9,9 +9,8 @@
  */
 package org.openmrs.module.fhir2;
 
-import javax.persistence.CollectionTable;
+import javax.persistence.CascadeType;
 import javax.persistence.Column;
-import javax.persistence.ElementCollection;
 import javax.persistence.Entity;
 import javax.persistence.EnumType;
 import javax.persistence.Enumerated;
@@ -19,24 +18,26 @@ import javax.persistence.GeneratedValue;
 import javax.persistence.GenerationType;
 import javax.persistence.Id;
 import javax.persistence.JoinColumn;
+import javax.persistence.JoinTable;
+import javax.persistence.OneToMany;
+import javax.persistence.OneToOne;
 import javax.persistence.Table;
 
-import java.util.Collection;
+import java.util.Set;
 
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
-import org.openmrs.BaseOpenmrsData;
+import org.openmrs.BaseOpenmrsMetadata;
 
 @Data(staticConstructor = "of")
 @NoArgsConstructor
 @EqualsAndHashCode(callSuper = true)
 @Entity
 @Table(name = "fhir_task")
-public class FhirTask extends BaseOpenmrsData {
+public class FhirTask extends BaseOpenmrsMetadata {
 	
 	// Based on https://www.hl7.org/fhir/task.html v4.0.1
-	// TODO: Support this valueset: https://www.hl7.org/fhir/valueset-task-status.html
 	public enum TaskStatus {
 		REQUESTED,
 		REJECTED,
@@ -44,7 +45,6 @@ public class FhirTask extends BaseOpenmrsData {
 		COMPLETED
 	};
 	
-	// TODO: Support this valueset: https://www.hl7.org/fhir/valueset-task-intent.html
 	public enum TaskIntent {
 		ORDER
 	}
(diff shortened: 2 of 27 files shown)


HerbertYiga:
You do have failing checks here

ibacher:
That's why its in draft mode... sometimes you've gotta collaborate on errors.

api/src/main/java/org/openmrs/module/fhir2/FhirTask.java
```diff
@@ -71,20 +71,47 @@
...
+	@CollectionTable(name = "fhir_task_based_on", joinColumns = @JoinColumn(name = "task_id"))
 	@Column(name = "based_on")
-	private String basedOn;
+	private Collection<String> basedOnReferences;
+	
+	/**
+	 * Referenced resources represented with relative resource identifier string in the format of
+	 * <ResourceName>/<ResourceId>.
+	 */
+	@Column(name = "for")
```
ibacher: Change this to anything else. Even "for_" will work.

api/src/main/java/org/openmrs/module/fhir2/FhirTask.java
```diff
@@ -71,20 +71,47 @@
...
+	private String forReference;
```
ibacher: Might it be worth it to create our own `Reference` type that's marked as `@Embedded` or similar that can handle parsing the reference properly?

api/src/main/java/org/openmrs/module/fhir2/FhirTask.java
```diff
@@ -68,23 +69,33 @@
...
+	private FhirReference encounterReference;
 	
-	@Column(name = "description")
-	private String description;
+	@OneToOne(cascade = CascadeType.ALL)
+	@JoinColumn(name = "owner_reference_id", referencedColumnName = "reference_id")
+	private FhirReference ownerReference;
 	
+	@OneToMany(cascade = CascadeType.ALL)
+	@JoinColumn(name = "task_id")
```
ibacher: I think you need to add `referencedColumnName`

api/src/main/java/org/openmrs/module/fhir2/FhirTaskParam.java
```diff
@@ -0,0 +1,51 @@
...
+import org.openmrs.BaseOpenmrsData;
+import org.openmrs.Concept;
+
+@Data(staticConstructor = "of")
+@NoArgsConstructor
+@EqualsAndHashCode(callSuper = true)
+@MappedSuperclass
+public abstract class FhirTaskParam extends BaseOpenmrsData {
+	
+	@OneToOne
```
ibacher: This should almost certainly be @OneToMany

pmanko:
Still need to add tests, do some more implementation work  - @ibacher feel free to comment on what's there so far, especially for `FhirTask`, `FhirReference`, `FhirTaskInput/Output`, and the Translator layer.

api/src/main/java/org/openmrs/module/fhir2/FhirTask.java
```diff
@@ -56,35 +59,72 @@
...
+	 */
 	@Column(name = "status")
 	@Enumerated(EnumType.STRING)
 	private TaskStatus status;
 	
+	/**
+	 * An explanation as to why this task is held, failed, was refused, etc.
+	 */
 	@Column(name = "status_reason")
 	private String statusReason;
```
ibacher: It might be worth dropping this field. It's supposed to be a `CodeableConcept`, but there's no real terminology binding, so it's hard to assess what we need to store.

api/src/main/java/org/openmrs/module/fhir2/FhirReference.java
```diff
@@ -0,0 +1,46 @@
...
+	
+	@Id
+	@GeneratedValue(strategy = GenerationType.AUTO)
+	@Column(name = "reference_id")
+	private Integer id;
+	
+	@Column(name = "reference_type")
+	private String type;
+	
+	@Column(name = "reference")
```
ibacher: Assuming this is storing the UUID of the reference, we can safely limit this column down to 32 characters (`size = 32`)

api/src/main/java/org/openmrs/module/fhir2/FhirTaskInput.java
```diff
@@ -0,0 +1,50 @@
...
+	
+	private static final long serialVersionUID = 1L;
+	
+	@Id
+	@GeneratedValue(strategy = GenerationType.AUTO)
+	@Column(name = "task_input_id")
+	private Integer id;
+	
+	/**
+	 * FhirTaskInput.type: A code or description indicating how the input is intended to be used as part
```
ibacher: I get what these are doing, but we should probably either 1) add a function here or 2) just add a Javadoc on the `FhirTaskParam` that doesn't explicitly reference input or output.

api/src/main/java/org/openmrs/module/fhir2/FhirTaskParam.java
```diff
@@ -0,0 +1,52 @@
...
+@NoArgsConstructor
+@EqualsAndHashCode(callSuper = true)
+@MappedSuperclass
+public abstract class FhirTaskParam extends BaseOpenmrsMetadata {
+	
+	@ManyToOne
+	@JoinColumn(name = "task_id")
+	protected FhirTask task;
+	
+	@Column(name = "type_id")
```
ibacher: This should probably be marked `nullable = false`

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/BaseServiceRequestTranslatorImpl.java
```diff
@@ -15,6 +15,7 @@
```
ibacher: Not a point for this PR, but maybe we should talk through this class. The more I look at it, the more I feel like its pretty tightly coupling `ServiceRequest` to `Task` which makes sense in the lab workflow, but I don't think is a good, generalizable thing.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ServiceRequestTranslatorImpl.java
```diff
@@ -37,8 +49,20 @@ public ServiceRequest toFhirResource(TestOrder order) {
...
+		
 		serviceRequest.setIntent(ServiceRequest.ServiceRequestIntent.ORDER);
+		
+		serviceRequest.setSubject(patientReferenceTranslator.toFhirResource(order.getPatient()));
+		
+		serviceRequest.setRequester(providerReferenceTranslator.toFhirResource(order.getOrderer()));
+		
+		serviceRequest.setPerformer(Collections.singletonList(determineServiceRequestPerformer(order.getUuid())));
+		
+		serviceRequest
```
ibacher: Does this work if there is no `effectiveStopDate`? I'm not sure if this parameter makes sense in the context of a one-off non-medication order.

pmanko: Yeah, I wasn't sure how exactly to model the effective dates to dateTime/Period/Timing. 

I'll talk about this at the design forum, but for now I'll set a period like this, and have an empty start and/or end. 

The FHIR docs mention this about missing times in period: 

> If the start element is missing, the start of the period is not known. If the end element is missing, it means that the period is ongoing, or the start may be in the past, and the end date in the future, which means that period is expected/planned to end at the specified time

This seems to generally capture how I understand effective start and end dates to work, but we could also model the occurrence as a single Datetime if there's only an effective start date or only an effective end date. 

(https://www.hl7.org/fhir/servicerequest-definitions.html#ServiceRequest.occurrence_x_)

api/src/main/java/org/openmrs/module/fhir2/FhirReference.java
pmanko: If this is to be a true representation of a FHIR Reference (and not just of a FHIR reference that targets system objects by UUID) we might want to relax this limit. 

The `reference` element is defined as:

> A reference to a location at which the other resource is found. The reference may be a relative reference, in which case it is relative to the service base URL, or an absolute URL that resolves to the location where the resource is found. The reference may be version specific or not. If the reference is not to a FHIR RESTful server, then it should be assumed to be version specific. Internal fragment references (start with '#') refer to contained resources.

> Using absolute URLs provides a stable scalable approach suitable for a cloud/web context, while using relative/logical references provides a flexible approach suitable for use when trading across closed eco-system boundaries. Absolute URLs do not need to point to a FHIR RESTful server, though this is the preferred approach. If the URL conforms to the structure "/[type]/[id]" then it should be assumed that the reference is to a FHIR RESTful server.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
```diff
@@ -53,7 +62,53 @@ public FhirTask getTaskByUuid(String uuid) {
 	
 	@Override
 	public Collection<FhirTask> getTasksByBasedOnUuid(Class<? extends DomainResource> clazz, String uuid) {
-		return (Collection<FhirTask>) sessionFactory.getCurrentSession().createCriteria(FhirTask.class)
-		        .add(Restrictions.eq("basedOn", clazz.getSimpleName() + "/" + uuid)).list();
+		Criteria criteria = sessionFactory.getCurrentSession().createCriteria(FhirTask.class, "task");
```
ibacher: Just a suggestion, but:

```java
return (Collection<FhirTask>) sessionFactory.getCurrentSession().createCriteria(FhirTask.class)
    .createAlias("basedOnReferences", "bo").add("bo.type", clazz.getAnnotation(ResourceDef.class).name())
    .add(eq("bo.reference", uuid)).list();
```

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
```diff
@@ -53,7 +62,53 @@ public FhirTask getTaskByUuid(String uuid) {
...
+	        TokenOrListParam status, SortSpec sort) {
+		
+		Criteria criteria = sessionFactory.getCurrentSession().createCriteria(FhirTask.class);
+		
+		// TODO: Refactor with BaseDaoImpl search support
+		// TODO: Handle optional params
+		// Task.basedOn
+		if (!(basedOnReference == null || basedOnReference.getIdPart() == null
+		        || basedOnReference.getResourceType() == null)) {
+			Criteria basedOnCriteria = criteria.createCriteria("basedOnReferences");
```
ibacher: See above

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
```diff
@@ -53,7 +62,53 @@ public FhirTask getTaskByUuid(String uuid) {
...
+			basedOnCriteria.add(Restrictions.eq("reference", basedOnReference.getIdPart()));
+			basedOnCriteria.add(Restrictions.eq("type", basedOnReference.getResourceType()));
+		}
+		
+		// Task.owner
+		if (!(ownerReference == null || ownerReference.getIdPart() == null || ownerReference.getResourceType() == null)) {
+			Criteria ownerCriteria = criteria.createCriteria("ownerReference");
```
ibacher: See above

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ReferenceTranslatorImpl.java
```diff
@@ -0,0 +1,64 @@
...
+	
+	@Override
+	public FhirReference toOpenmrsType(Reference fhirReference) {
+		FhirReference openmrsReference = null;
+		
+		if (fhirReference != null) {
+			openmrsReference = new FhirReference();
+			openmrsReference.setType(fhirReference.getType());
+			openmrsReference.setReference(fhirReference.getReference());
+			// TODO figure out what to use for the non-nullable Name
```
ibacher: I'd use the full reference, e.g., `Patient/00000000000` for the name.

ibacher:
@pmanko: could you remove the commented-out code in FhirTaskDaoImpl. Other than that, I think this is all set.

pmanko:
> @pmanko: could you remove the commented-out code in FhirTaskDaoImpl. Other than that, I think this is all set.

should be set!

PR #85, opened 2020-02-22
FM2-81: Improving Search for Person Resource

CaptainDaVinci:

* Model `FhirPersonResourceProvider`  to provide search  by name, gender, birth date, address.
* Added tests for the Provider class

Todo:
- [x] Implement the search in `FhirPersonDaoImpl`.
- [x] Add tests for `FhirPersonResourceWebTest`, `FhirPersonDao`,  `FhirPersonDaoImpl`

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-81

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/FhirPersonService.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/FhirPersonService.java
@@ -10,19 +10,19 @@
 package org.openmrs.module.fhir2.api;
 
 import java.util.Collection;
-import java.util.Date;
 
+import ca.uhn.fhir.rest.api.SortSpec;
+import ca.uhn.fhir.rest.param.DateRangeParam;
+import ca.uhn.fhir.rest.param.StringOrListParam;
+import ca.uhn.fhir.rest.param.TokenOrListParam;
 import org.hl7.fhir.r4.model.Person;
 
 public interface FhirPersonService {
 	
 	Person getPersonByUuid(String uuid);
 	
-	Collection<Person> findPersonsByName(String name);
+	Collection<Person> searchForPeople(StringOrListParam name, TokenOrListParam gender, DateRangeParam birthDate,
+	        StringOrListParam city, StringOrListParam state, StringOrListParam postalCode, StringOrListParam country,
+	        SortSpec sort);
 	
-	Collection<Person> findSimilarPeople(String name, Integer birthYear, String gender);
-	
-	Collection<Person> findPersonsByGender(String gender);
-	
-	Collection<Person> findPersonsByBirthDate(Date birthDate);
 }
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirPersonDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirPersonDaoImpl.java
@@ -15,11 +15,15 @@ import javax.inject.Inject;
 import javax.inject.Named;
 
 import java.util.Collection;
-import java.util.Date;
 import java.util.List;
 
+import ca.uhn.fhir.rest.api.SortSpec;
+import ca.uhn.fhir.rest.param.DateRangeParam;
+import ca.uhn.fhir.rest.param.StringOrListParam;
+import ca.uhn.fhir.rest.param.TokenOrListParam;
 import lombok.AccessLevel;
 import lombok.Setter;
+import org.hibernate.Criteria;
 import org.hibernate.SessionFactory;
 import org.hibernate.sql.JoinType;
 import org.openmrs.Person;
@@ -30,7 +34,7 @@ import org.springframework.stereotype.Component;
 
 @Component
 @Setter(AccessLevel.PACKAGE)
-public class FhirPersonDaoImpl implements FhirPersonDao {
+public class FhirPersonDaoImpl extends BaseDaoImpl implements FhirPersonDao {
 	
 	@Inject
 	PersonService personService;
@@ -45,28 +49,6 @@ public class FhirPersonDaoImpl implements FhirPersonDao {
 	}
 	
 	@Override
-	public Collection<Person> findPersonsByName(String name) {
-		return personService.getPeople(name, false);
-	}
-	
-	@Override
-	@SuppressWarnings("unchecked")
-	public Collection<Person> findPersonsByBirthDate(Date birthDate) {
-		return sessionFactory.getCurrentSession().createCriteria(Person.class).add(eq("birthdate", birthDate)).list();
-	}
-	
-	@Override
-	public Collection<Person> findSimilarPeople(String name, Integer birthYear, String gender) {
-		return personService.getSimilarPeople(name, birthYear, gender);
-	}
-	
-	@Override
-	@SuppressWarnings("unchecked")
-	public Collection<Person> findPersonsByGender(String gender) {
-		return sessionFactory.getCurrentSession().createCriteria(Person.class).add(eq("gender", gender)).list();
-	}
-	
-	@Override
 	public List<PersonAttribute> getActiveAttributesByPersonAndAttributeTypeUuid(Person person,
 	        String personAttributeTypeUuid) {
 		return (List<PersonAttribute>) sessionFactory.getCurrentSession().createCriteria(PersonAttribute.class)
@@ -74,4 +56,51 @@ public class FhirPersonDaoImpl implements FhirPersonDao {
 		        .createAlias("attributeType", "pat").add(eq("pat.uuid", personAttributeTypeUuid)).add(eq("voided", false))
 		        .list();
 	}
+	
+	@Override
+	public Collection<Person> searchForPeople(StringOrListParam name, TokenOrListParam gender, DateRangeParam birthDate,
+	        StringOrListParam city, StringOrListParam state, StringOrListParam postalCode, StringOrListParam country,
+	        SortSpec sort) {
+		Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Person.class);
+		
+		handleNames(criteria, name, null, null);
+		handleGender("gender", gender).ifPresent(criteria::add);
+		handleDateRange("birthdate", birthDate).ifPresent(criteria::add);
+		handlePersonAddress("pad", city, state, postalCode, country).ifPresent(c -> {
+			criteria.createAlias("addresses", "pad");
+			criteria.add(c);
+		});
+		if (sort != null) {
+			if (sort.getParamName().equals("name") && !containsAlias(criteria, "pn")) {
+				criteria.createAlias("names", "pn");
+			}
+			if (sort.getParamName().startsWith("address") && !containsAlias(criteria, "pad")) {
+				criteria.createAlias("addresses", "pad");
+			}
+		}
+		handleSort(criteria, sort);
+		
+		return criteria.list();
+	}
+	
+	@Override
+	protected String paramToProp(String param) {
+		switch (param) {
+			case "name":
+				return "pn.givenName";
+			case "birthdate":
+				return "birthdate";
+			case "address-city":
+				return "pad.cityVillage";
+			case "address-state":
+				return "pad.stateProvince";
+			case "address-postalCode":
+				return "pad.postalCode";
+			case "address-country":
+				return "pad.country";
+			default:
+				return null;
+		}
+	}
+	
 }
--- a/api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirPersonServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirPersonServiceImpl.java
@@ -12,9 +12,12 @@ package org.openmrs.module.fhir2.api.impl;
 import javax.inject.Inject;
 
 import java.util.Collection;
-import java.util.Date;
 import java.util.stream.Collectors;
 
+import ca.uhn.fhir.rest.api.SortSpec;
+import ca.uhn.fhir.rest.param.DateRangeParam;
+import ca.uhn.fhir.rest.param.StringOrListParam;
+import ca.uhn.fhir.rest.param.TokenOrListParam;
 import lombok.AccessLevel;
 import lombok.Setter;
 import org.hl7.fhir.r4.model.Person;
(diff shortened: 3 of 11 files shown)


CaptainDaVinci:
Creating a draft PR to know if I'm proceeding in the right way with this ticket.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirPersonServiceImpl.java
```diff
@@ -68,4 +72,13 @@ public Person getPersonByUuid(String uuid) {
 		return fhirPersonDao.findPersonsByGender(gender).stream().map(personTranslator::toFhirResource)
 		        .collect(Collectors.toList());
 	}
+	
+	@Override
+	public Collection<Person> searchForPeople(StringOrListParam name, TokenOrListParam identifier, TokenOrListParam gender,
+	        DateRangeParam birthDate, StringOrListParam city, StringOrListParam state, StringOrListParam postalCode,
+	        SortSpec sort) {
+		// TODO Auto-generated method stub
```
VaishSiddharth: @CaptainDaVinci I think we'll have to write something like what's written in `FhirObservationDaoImpl` @ibacher Please guide

```java
@Override
	public Collection<Obs> searchForObservations(ReferenceParam encounterReference, ReferenceParam patientReference,
	        TokenAndListParam code, SortSpec sort) {
		Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Obs.class);
		
		handleEncounterReference("e", encounterReference).ifPresent(c -> criteria.createAlias("encounter", "e").add(c));
		handlePatientReference(criteria, patientReference);
		handleCodedConcept(criteria, code);
		handleSort(criteria, sort);
		
		return criteria.list();
	}
```

CaptainDaVinci: Yes, I'm working on it. I'm using PatientDaoImpl as a reference.

CaptainDaVinci:
@ibacher  I made a comment about this in @VaishSiddharth 's PR as well, can we refactor the `handleNames` and `handleIdentifiers` methods to be present in `BaseDaoImpl`, as it is used by multiple classes.

ibacher:
Yes, please do so.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirPersonDaoImpl.java
```diff
@@ -74,4 +87,61 @@ public Person getPersonByUuid(String uuid) {
 		        .createAlias("attributeType", "pat").add(eq("pat.uuid", personAttributeTypeUuid)).add(eq("voided", false))
 		        .list();
 	}
+	
+	@Override
+	public Collection<Person> searchForPeople(StringOrListParam name, TokenOrListParam identifier, TokenOrListParam gender,
```
VaishSiddharth: @CaptainDaVinci I would suggest you to write test for this as I am facing issues while implementing this for `Encounters` Don't know but my tests are failing

CaptainDaVinci: Thanks! I'll update the tests in sometime.

api/src/main/java/org/openmrs/module/fhir2/api/FhirPersonService.java
```diff
@@ -25,4 +29,9 @@
 	Collection<Person> findPersonsByGender(String gender);
 	
 	Collection<Person> findPersonsByBirthDate(Date birthDate);
+	
+	Collection<Person> searchForPeople(StringOrListParam name, TokenOrListParam identifier, TokenOrListParam gender,
```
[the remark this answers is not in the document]
CaptainDaVinci: Will it be a good idea to return a "Not Implemented" response (not sure how) to a query on `identifier` field? In case we plan to map the field in future.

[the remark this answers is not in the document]
ibacher: So if we just don’t provide an implementation of it at all (i.e., remove it from this method) the HAPI library will do that for us.

CaptainDaVinci:
Sure, I'm working on the tests for `FhirPersonServiceImpl` and `FhirPersonDaoImpl`, I'll work on implementing sort as soon as the tests are done.

CaptainDaVinci:
@ibacher the new tests are failing on travis, but run successfully when run locally. Any idea what could be causing this issue here?
The build shows these as failures,
```
shouldGetPersonByBirthDateGreaterThanOrEqualTo(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
  shouldGetPersonByBirthDateLessThan(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
  shouldGetPersonByName(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
  shouldGetPersonByGender(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
  shouldGetPersonByBirthDateBetween(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
  shouldGetPersonByBirthDateLessThanOrEqualTo(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
  shouldGetPersonByBirthDate(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
  shouldGetPersonByBirthDateGreaterThan(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
```
The [log](https://travis-ci.com/openmrs/openmrs-module-fhir2/builds/150507491#L3556) shows all of these error originate from the `PersonFhirResourceProviderWebTest.verifyUri` method where the response bundle does not have any entries.

P.S: I've had similar tests failing when run on local machine, but they seem to be working now, not exactly sure what the issue is.

CaptainDaVinci:
@ibacher the new tests are failing on travis, but run successfully when run locally. Any idea what could be causing this issue here?
The build shows these as failures,
```
shouldGetPersonByBirthDateGreaterThanOrEqualTo(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
  shouldGetPersonByBirthDateLessThan(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
  shouldGetPersonByName(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
  shouldGetPersonByGender(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
  shouldGetPersonByBirthDateBetween(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
  shouldGetPersonByBirthDateLessThanOrEqualTo(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
  shouldGetPersonByBirthDate(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
  shouldGetPersonByBirthDateGreaterThan(org.openmrs.module.fhir2.providers.PersonFhirResourceProviderWebTest): (..)
```
The [log](https://travis-ci.com/openmrs/openmrs-module-fhir2/builds/150507491#L3556) shows all of these error originate from the `PersonFhirResourceProviderWebTest.verifyUri` method where the response bundle does not have any entries.

P.S: I've had similar tests failing when run on local machine, but they seem to be working now, not exactly sure what the issue is.

ibacher:
@CaptainDaVinci You're probably getting bitten by a non-determinism issue caused by having left those other methods in that allow, e.g., search by name or birthdate. In particular, the `findPersonsByBirthDate()` method probably has priority over your new search method in part because the "birthdate" param is marked as `@Required`. Just remove the old search methods and things should be fine.

CaptainDaVinci:
Thanks! That resolved the issue.

CaptainDaVinci:
The test `FhirPersonDaoImpl.shouldReturnEmptyCollectionForNoMatchOnGender` did not return an empty collection for the parameter "wrong gender" (it returned a list of both male and female person), I changed it to "other" (like in `FhirPatientDaoImplTest`) and it works now, because the data set does not have a person with gender set as other. Is this the intended behavior, or should it test for incorrect parameters given?

ibacher:
@CaptainDaVinci For right now, the behaviour should match that of the Patient class; however, it might be a good idea to start a post on Talk to discuss the design of this.

ibacher:
In any case, this probably isn't a detail that should be sorted out at the DAO layer, but further up, such as in the ResourceProvider or in an Interceptor of some kind.

CaptainDaVinci:
@ibacher I have started a new topic on talk, https://talk.openmrs.org/t/search-test-behavior-for-fhir-person-patient-resource/27025 to discuss the design.

ibacher:
@CaptainDaVinci Thanks for starting the Talk post. When implementing sorting, it's also necessary to override the `paramToProp()` function in BaseDaoImpl, so that it takes the sort parameter name, e.g., "name" and translates it into the correct property in the query, e.g., "p.name" or whatever it is.

ibacher:
Other than that, how close do you think this is to being done?

CaptainDaVinci:
@ibacher  I've updated the tests for all the related classes, `Service`, `Dao`, `Resourceprovider`, `ResourceProviderWeb` after correcting the sort method it should be done.

api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirPersonDaoImplTest.java
```diff
@@ -62,18 +65,24 @@
...
-	private static final int PERSON_BIRTH_YEAR = 1999;
-	
-	private static final int WRONG_BIRTH_YEAR = 1000;
-	
 	private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd");
 	
 	private static final String BIRTH_DATE = "1999-12-20";
 	
-	private static final String NOT_FOUND_BIRTH_DATE = "1000-00-00";
+	private static final String NOT_FOUND_BIRTH_DATE = "0001-01-01";
```
ibacher: Was this change necessary?

api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirPersonDaoImplTest.java
```diff
@@ -187,4 +240,5 @@ public void getActiveAttributesByPersonAndAttributeTypeUuid_shouldReturnPersonAt
 		
 		assertThat(attributeList, notNullValue());
 	}
+
```
ibacher: I don't think we need a new-line here

api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirPersonDaoImplTest.java
```diff
@@ -27,9 +27,12 @@
 import java.text.ParseException;
```
ibacher: Is it possible to add a test/tests here that searches by multiple parameters at the same time?

omod/src/test/java/org/openmrs/module/fhir2/providers/PersonFhirResourceProviderWebTest.java
```diff
@@ -10,15 +10,33 @@
 package org.openmrs.module.fhir2.providers;
```
ibacher: I know this isn't in the model I provided, but it would be nice if this included a test or tests for "complex" queries such as `/Person?given=Harold&gender=male,other`.

api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirPersonDaoImplTest.java
CaptainDaVinci: Yes, else the `DataRangeParam` returned an error on parsing.

CaptainDaVinci:
@ibacher to sort the results by the given name, what should the return value of `paramToProp()` be?
I tried returning `names.givenName` but this results in an error, 

`org.hibernate.QueryException: could not resolve property: names.givenName of: org.openmrs.Person`

I checked in the hibernate mapping files, in the core repo, but couldn't find anything on `Person`.

ibacher:
@CaptainDaVinci So the property should be related to the structure of the Hibernate *query*, so you'd generally want to use something like `pn.givenName`. However, of course, the name may not be part of the attributes already queried so you should end up with something like this:

```java
if (!containsAlias(criteria, "pn") {
	criteria.createAlias("names", "pn");
}

return "pn.given";
```

CaptainDaVinci:
@ibacher in that case, will it be a good idea to set the aliases (names and addresses) in the `FhirPersonDaoImpl.searchForPeople` method instead of setting them in the `BaseDaoImpl` method? 
That way we don't have to set the aliases in `handleSort` again.

ibacher:
@CaptainDaVinci So that was the original setup. I switched to the present one because it seemed best to only create the alias when it was actually required. By default, `createAlias` results in adding an `INNER JOIN`, which is harmful if the relevant data do not exist, but also results in more work on the database end that we don't actually need. Creating the alias only when we need it solves both of those problems, at the cost that we kind of need to be using a naming convention for aliases.

That said, if you can think of a way to implement this in the `searchForPeople` method that won't create unnecessary joins, I'm all in favour of it.

CaptainDaVinci:
So, we would ideally want to create the aliases for sorting within `paramToProp` method instead of creating them in `searchForPeople` like how it's done now.

ibacher:
@CaptainDaVinci Well, I agree its not an ideal place to be creating aliases. My concern is to ensure that we only create aliases when we need an explicit `JOIN`, and this was the best solution that occurred to me so far.

CaptainDaVinci:
Got it! So, in the latest commit , I've just created aliases in `searchForPeople`, this would result in unnecessary `JOIN`s if the sort parameter is neither name nor an address field.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirPersonDaoImpl.java
```diff
@@ -45,33 +49,56 @@ public Person getPersonByUuid(String uuid) {
...
+		Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Person.class);
+		
+		handleNames(criteria, name, null, null);
+		handleGender("gender", gender).ifPresent(criteria::add);
+		handleDateRange("birthdate", birthDate).ifPresent(criteria::add);
+		handlePersonAddress("pad", city, state, postalCode, country).ifPresent(c -> {
+			criteria.createAlias("addresses", "pad");
+			criteria.add(c);
+		});
+		if (!containsAlias(criteria, "pn")) {
```
ibacher: Again... this is creating the alias regardless of what is being queried, which is not what we want to do. I might change your `paramToProp()` to something like this:

```java
protected String paramToProp(String param) {
		String prop = null;

		switch (param) {
			case "name":
				if (!containsAlias(criteria, "pn")) {
					criteria.createAlias("names", "pn");
				}

				return "pn.givenName";
			case "birthDate":
				return "birthdate";
			case "city":
				prop = "pad.cityVillage";
			case "state":
				prop = "pad.stateProvince";
			case "postalCode":
				prop = "pad.postalCode";
			case "country":
				prop = "pad.country";
                
                if (!containsAlias(criteria, "pad")) {
					criteria.createAlias("addresses", "pad");
				}
				
				return prop;
			default:
				return null;
		}
	}
```

CaptainDaVinci: Yes, I'm working on this right now.

CaptainDaVinci:
@ibacher I updated the to sort implementation to check before creating an alias within `searchForPeople` itself, to avoid passing the criteria object or instead making it a data member.

The `PersonDaoImplTest` have also been updated to check for sorting by `givenName` when a person has multiple names.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirPersonDaoImpl.java
```diff
@@ -70,11 +70,13 @@ public Person getPersonByUuid(String uuid) {
 			criteria.createAlias("addresses", "pad");
 			criteria.add(c);
 		});
-		if (!containsAlias(criteria, "pn")) {
-			criteria.createAlias("names", "pn");
-		}
-		if (!containsAlias(criteria, "pad")) {
-			criteria.createAlias("addresses", "pad");
+		if (sort != null) {
```
ibacher: I still don't like this solution. For example, if I were to sort by "birthdate", I'd still get joins against `PersonName` and `PersonAddress`.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirPersonDaoImpl.java
```diff
@@ -88,13 +90,13 @@ protected String paramToProp(String param) {
 				return "pn.givenName";
 			case "birthDate":
```
ibacher: This should be "birthdate" to match the name of the search parameter.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirPersonDaoImpl.java
CaptainDaVinci: But the `sort.getParamName().equals(...)` condition will return false then, right?

ibacher: Whoops! I missed that! Sorry, too much time staring at git diffs.

CaptainDaVinci: It's alright, happens to me all the time! 😄

PR #87, opened 2020-02-25
FM2-84 : Additional properties to allergyIntolerance resource post-commit

jecihjoy:
## Issue I worked on

see https://issues.openmrs.org/browse/FM2-84

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/AllergyIntoleranceTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/AllergyIntoleranceTranslatorImpl.java
@@ -11,16 +11,24 @@ package org.openmrs.module.fhir2.api.translators.impl;
 
 import javax.inject.Inject;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import lombok.AccessLevel;
 import lombok.Setter;
 import org.hl7.fhir.r4.model.AllergyIntolerance;
+import org.hl7.fhir.r4.model.Annotation;
 import org.hl7.fhir.r4.model.CodeableConcept;
 import org.hl7.fhir.r4.model.Coding;
 import org.openmrs.Allergen;
 import org.openmrs.AllergenType;
 import org.openmrs.Allergy;
+import org.openmrs.AllergyReaction;
+import org.openmrs.Concept;
 import org.openmrs.User;
 import org.openmrs.module.fhir2.FhirConstants;
+import org.openmrs.module.fhir2.api.FhirConceptService;
+import org.openmrs.module.fhir2.api.FhirGlobalPropertyService;
 import org.openmrs.module.fhir2.api.translators.AllergyIntoleranceTranslator;
 import org.openmrs.module.fhir2.api.translators.PatientReferenceTranslator;
 import org.openmrs.module.fhir2.api.translators.PractitionerReferenceTranslator;
@@ -35,6 +43,12 @@ public class AllergyIntoleranceTranslatorImpl extends AbstractReferenceHandlingT
 	
 	@Inject
 	private PatientReferenceTranslator patientReferenceTranslator;
+	
+	@Inject
+	private FhirGlobalPropertyService globalPropertyService;
+	
+	@Inject
+	private FhirConceptService conceptService;
 	
 	@Override
 	public AllergyIntolerance toFhirResource(Allergy omrsAllergy) {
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/AllergyIntoleranceTranslatorImpl.java
```diff
@@ -124,4 +167,75 @@ private boolean isAllergyInactive(CodeableConcept status) {
...
+		}
+		
+		return manifestations;
+	}
+	
+	private AllergyIntolerance.AllergyIntoleranceSeverity getFhirSeverity(Concept severityConcept) {
+		if (severityConcept == null) {
+			return null;
+		}
+		if (globalPropertyService.getGlobalProperty("allergy.concept.severity.mild") != null && globalPropertyService
```
ibacher: 1. I'd prefer if this constant was referenced from a constant location, even if it ends up in `FhirConstants`.

2. Instead of calling `getGlobalProperty()` twice, call it once. This could be rewritten as `globalPropertyService.getGlobalProperty(FhirConstants.ALLERGY_MILD, "").equals(severityConcept.getUuid())`. This is because each call to `getGlobalProperty()` results in a query to the database.

jecihjoy: @ibacher i'm working on the tests, also i'll move those values to constant

jecihjoy: @ibacher without this check **globalPropertyService.getGlobalProperty("allergy.concept.severity.mild") != null**
i'm getting  **java.lang.NullPointerException**

ibacher: Right, sorry, I forgot this was our `GlobalProperty` service. Add a method to it like this:

```java
@Override
public String getGlobalProperty(String property, String defaultValue) throws APIException {
	String result = dao.getGlobalProperty(property);
	return result == null ? defaultValue : result;
}
```

and use that. Alternatively, use this pattern:

```java
String gp = globalPropertyService.getGlobalProperty(...);

if (gp != null && gp.equals(...)) {
	// ...
}
```

ibacher: Actually, change that first suggestion. Instead, on the `FhirGlobalProperty` interface add this:

```java
default String getGlobalProperty(String property, String defaultValue) throws APIException {
    String result = this.getGlobalProperty(property);
    return result == null ? defaultValue : result;
}
```

That way it works for any other implementations of `FhirGlobalProperty` without too much hassle.

jecihjoy: @ibacher adding change breaks this test  https://pastebin.com/6JG2P8b0 and I'm not sure how to fix it

here is the [branch](https://github.com/jecihjoy/openmrs-module-fhir2/tree/FM2-84-b) with these changes

How can I change [this](https://github.com/jecihjoy/openmrs-module-fhir2/blob/4450b4cc7d7127211e9c9b3b57f3182b9561d488/omod/src/test/java/org/openmrs/module/fhir2/web/servlet/BaseFhirResourceProviderTest.java#L99) to a normal java function?

ibacher: @jecihjoy I forgot about that. The equivalent is:

```java
servlet.setGlobalPropertyService(new FhirGlobalPropertyServiceImpl() {
	@Override
    public String getGlobalProperty(String property) throws APIException {
        switch (property) {
			case FhirConstants.OPENMRS_FHIR_DEFAULT_PAGE_SIZE:
				return "10";
			case FhirConstants.OPENMRS_FHIR_MAXIMUM_PAGE_SIZE:
				return "100";
		}
    }
});
```

jecihjoy:
Even after adding the changes above I'm getting a null pointer exception in this code

_globalPropertyService.getGlobalProperty(FhirConstants.GLOBAL_PROPERTY_MILD, "")_

ibacher:
That's because when using the mocked `globalPropertyService` you'll need to provide an answer for those calls. Something like this, in addition to the already-existing `when()` clause will work:

```java
when(globalPropertyService.getGlobalProperty(MockitoHamcrest.argThat(not(equalTo(GLOBAL_PROPERTY_SEVERE))), anyString())).thenReturn("");
```

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/AllergyIntoleranceTranslatorImpl.java
jecihjoy: I'm actually doing the mocking
eg  [this is](https://github.com/jecihjoy/openmrs-module-fhir2/blob/36501adfb9f8548f39e606f64e96ef73ef188d7e/api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/AllergyIntoleranceTranslatorImplTest.java#L290) where the tests starts failing if I remove the null check from the translator

The first element in the list, **MILD** works fine but the rest on the tests on severity won't pass

PR #99, opened 2020-03-01
FM2-88 Enhance Observation Search

VaishSiddharth:

FM2-88 Enhanced Observation Search

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-88

- [ ] My pull request only contains **ONE single commit**.

  No? -> execute above command


api/src/main/java/org/openmrs/module/fhir2/api/FhirObservationService.java
```diff
@@ -23,5 +26,7 @@
 	Observation getObservationByUuid(@NotNull String uuid);
 	
 	Collection<Observation> searchForObservations(ReferenceParam encounterReference, ReferenceParam patientReference,
+	        ReferenceParam hasMemberReference, TokenAndListParam valueConcept, DateRangeParam valueDateParam,
```
VaishSiddharth: @ibacher I have a few questions here
1. What should be the data type of hasMemberReference I think that `TokenOrListParam` should be the datatype as we do have a function in `BaseDaoImpl` to handleBoolean.
2. I am a bit confused that how should we handle the `value` parameters, should we have only a single ReferenceParam and include all of these value-concept value-date value-quantity value-string, and then make a function to handle all the value related search queries.

ibacher: @VaishSiddharth

1. So the `hasMember` of [FHIR Observations](https://www.hl7.org/fhir/observation-definitions.html#Observation.hasMember) is not a boolean value. It stores a reference to an observation that is part of this observation group. So this *should* be a `ReferenceParam` and treated similarly to how we treat `Subject`, etc.

2. The [FHIR spec](https://www.hl7.org/fhir/observation.html#search) actually wants separate parameters for each value type, so `value-concept` should map to a `TokenAndListParam`, `value-date` to a `DateRangeParam`, `value-quantity` to a `QuantityAndListParam` and `value-string` to a `StringAndListParam`.

omod/src/main/java/org/openmrs/module/fhir2/providers/ObservationFhirResourceProvider.java
```diff
@@ -61,8 +64,15 @@ public Observation getObservationById(@IdParam @NotNull IdType id) {
 	public Bundle searchObservations(@OptionalParam(name = Observation.SP_ENCOUNTER) ReferenceParam encounterReference,
 	        @OptionalParam(name = Observation.SP_SUBJECT, chainWhitelist = { "", Patient.SP_IDENTIFIER, Patient.SP_GIVEN,
 	                Patient.SP_FAMILY, Patient.SP_NAME }, targetTypes = Patient.class) ReferenceParam patientReference,
+	        @OptionalParam(name = Observation.SP_HAS_MEMBER) ReferenceParam hasMemberReference,
+	        @OptionalParam(name = Observation.SP_VALUE_CONCEPT) TokenAndListParam valueConcept,
+	        @OptionalParam(name = Observation.SP_VALUE_DATE) DateRangeParam valueDateParam,
+	        @OptionalParam(name = Observation.SP_VALUE_QUANTITY) QuantityAndListParam valueQuantityParam,
+	        @OptionalParam(name = Observation.SP_VALUE_STRING) StringAndListParam valueStringParam,
+	        @OptionalParam(name = Observation.SP_DATE) DateRangeParam date,
 	        @OptionalParam(name = Observation.SP_CODE) TokenAndListParam code, @Sort SortSpec sort) {
```
ibacher: I'd actually prefer it if `SP_CODE` came after the `SP_SUBJECT` and before all these other search functions.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
```diff
@@ -63,16 +64,43 @@ public Obs getObsByUuid(String uuid) {
 		
 		handleEncounterReference("e", encounterReference).ifPresent(c -> criteria.createAlias("encounter", "e").add(c));
 		handlePatientReference(criteria, patientReference, "person");
-//		handleBoolean("hasGroupMembers",hasMemberReference);
-
-		handleDateRange("valueDatetime",valueDateParam);
-		handleDateRange("obsDatetime",date);
+		handleHasMemberReference("hm", hasMemberReference).ifPresent(c -> criteria.createAlias("groupMembers", "hm").add(c));
```
VaishSiddharth: @ibacher Please have a look at the handlers. I am still left with the `has-member.code` parameter. Not very sure about how to convert value to codeable concept. I guess something like this 

```java
handleAndListParamBySystem(valueConcept, (system, tokens) -> {
CodeableConcept codeableConcept=new CodeableConcept();
		codeableConcept.addCoding().setCode(Not sure what will come here).setSystem(system);
});
```

omod/src/main/java/org/openmrs/module/fhir2/providers/ObservationFhirResourceProvider.java
VaishSiddharth: will make this change in the next commit

api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImplTest.java
```diff
@@ -394,4 +412,75 @@ public void searchForObs_shouldHandleComplexQuery() {
...
+		
+		assertThat(results, notNullValue());
+		assertThat(results, not(empty()));
+		assertThat(results, hasItem(hasProperty("uuid", equalTo(OBS_UUID))));
+	}
+	
+	//	@Test
+	//	public void searchForObs_shouldSearchForObsByValueQuantity() {
+	//		QuantityAndListParam quantityAndListParam=new QuantityAndListParam();
+	//
```
VaishSiddharth: @ibacher Please help me resolve this issue

ibacher: Change `VALUE_QUANTITY` from "134.0" to 134.0d and change the type to double.

VaishSiddharth: @ibacher Still the same error

ibacher: @VaishSiddharth Can you put the full stack trace somewhere?

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
```diff
@@ -52,17 +57,50 @@ public Obs getObsByUuid(String uuid) {
...
 		return criteria.list();
 	}
 	
+	protected Optional<Criterion> handleValueQuantityParam(@NotNull String propertyName,
+	        QuantityAndListParam valueQuantityParam) {
+		if (valueQuantityParam == null) {
+			return Optional.empty();
+		}
+		return handleAndListParam(valueQuantityParam,
+		    v -> Optional.of(ilike(propertyName, v.getValueAsString(), MatchMode.EXACT)));
```
ibacher: We shouldn't use `ilike` here, which is only for matching strings. We should use `eq`.

api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImplTest.java
VaishSiddharth:
``` java
@Test
	public void searchForObs_shouldSearchForObsByValueQuantity() {
		QuantityAndListParam quantityAndListParam = new QuantityAndListParam();
		
		QuantityOrListParam quantityOrListParam = new QuantityOrListParam();
		
		QuantityParam quantityParam = new QuantityParam();
		quantityParam.setValue(134.0d);
		quantityAndListParam.addAnd(quantityOrListParam.add(quantityParam));
		
		// Error java.lang.ClassCastException: class java.lang.String cannot be cast to class java.lang.Double (java.lang.String and java.lang.Double are in module java.base of loader 'bootstrap')
		Collection<Obs> results = dao.searchForObservations(null, null, null, null, null, quantityAndListParam, null, null,
		    null, null);
		
		assertThat(results, not(empty()));
		assertThat(results, hasItem(hasProperty("uuid", equalTo(OBS_UUID))));
	}
```
```
/usr/lib/jvm/jdk-13.0.1/bin/java -ea -Didea.test.cyclic.buffer.size=1048576 -javaagent:/snap/intellij-idea-community/208/lib/idea_rt.jar=45755:/snap/intellij-idea-community/208/bin -Dfile.encoding=UTF-8 -classpath /snap/intellij-idea-community/208/lib/idea_rt.jar:/snap/intellij-idea-community/208/plugins/junit/lib/junit5-rt.jar:/snap/intellij-idea-community/208/plugins/junit/lib/junit-rt.jar:/home/sid/IdeaProjects/openmrs-module-fhir2/api/target/test-classes:/home/sid/IdeaProjects/openmrs-module-fhir2/api/target/classes:/home/sid/.m2/repository/ca/uhn/hapi/fhir/hapi-fhir-base/4.2.0/hapi-fhir-base-4.2.0.jar:/home/sid/.m2/repository/com/google/code/gson/gson/2.8.5/gson-2.8.5.jar:/home/sid/.m2/repository/org/apache/commons/commons-lang3/3.9/commons-lang3-3.9.jar:/home/sid/.m2/repository/org/apache/commons/commons-text/1.7/commons-text-1.7.jar:/home/sid/.m2/repository/commons-codec/commons-codec/1.12/commons-codec-1.12.jar:/home/sid/.m2/repository/ca/uhn/hapi/fhir/hapi-fhir-structures-r4/4.2.0/hapi-fhir-structures-r4-4.2.0.jar:/home/sid/.m2/repository/ca/uhn/hapi/fhir/org.hl7.fhir.utilities/4.2.0/org.hl7.fhir.utilities-4.2.0.jar:/home/sid/.m2/repository/ca/uhn/hapi/fhir/org.hl7.fhir.r4/4.2.0/org.hl7.fhir.r4-4.2.0.jar:/home/sid/.m2/repository/org/fhir/ucum/1.0.2/ucum-1.0.2.jar:/home/sid/.m2/repository/xpp3/xpp3/1.1.4c/xpp3-1.1.4c.jar:/home/sid/.m2/repository/xpp3/xpp3_xpath/1.1.4c/xpp3_xpath-1.1.4c.jar:/home/sid/.m2/repository/com/google/code/findbugs/jsr305/3.0.2/jsr305-3.0.2.jar:/home/sid/.m2/repository/org/projectlombok/lombok/1.18.10/lombok-1.18.10.jar:/home/sid/.m2/repository/javax/inject/javax.inject/1/javax.inject-1.jar:/home/sid/.m2/repository/com/google/guava/guava/28.1-jre/guava-28.1-jre.jar:/home/sid/.m2/repository/com/google/guava/failureaccess/1.0.1/failureaccess-1.0.1.jar:/home/sid/.m2/repository/com/google/guava/listenablefuture/9999.0-empty-to-avoid-conflict-with-guava/listenablefuture-9999.0-empty-to-avoid-conflict-with-guava.jar:/home/sid/.m2/repository/org/checkerframework/checker-qual/2.8.1/checker-qual-2.8.1.jar:/home/sid/.m2/repository/com/google/errorprone/error_prone_annotations/2.3.2/error_prone_annotations-2.3.2.jar:/home/sid/.m2/repository/com/google/j2objc/j2objc-annotations/1.3/j2objc-annotations-1.3.jar:/home/sid/.m2/repository/org/codehaus/mojo/animal-sniffer-annotations/1.18/animal-sniffer-annotations-1.18.jar:/home/sid/.m2/repository/org/exparity/hamcrest-date/2.0.7/hamcrest-date-2.0.7.jar:/home/sid/.m2/repository/org/openmrs/api/openmrs-api/2.0.5/openmrs-api-2.0.5.jar:/home/sid/.m2/repository/commons-collections/commons-collections/3.2.2/commons-collections-3.2.2.jar:/home/sid/.m2/repository/commons-beanutils/commons-beanutils/1.7.0/commons-beanutils-1.7.0.jar:/home/sid/.m2/repository/commons-io/commons-io/1.4/commons-io-1.4.jar:/home/sid/.m2/repository/org/azeckoski/reflectutils/0.9.14/reflectutils-0.9.14.jar:/home/sid/.m2/repository/org/apache/velocity/velocity/1.6.2/velocity-1.6.2.jar:/home/sid/.m2/repository/commons-lang/commons-lang/2.4/commons-lang-2.4.jar:/home/sid/.m2/repository/log4j/log4j/1.2.15/log4j-1.2.15.jar:/home/sid/.m2/repository/org/springframework/spring-core/4.1.4.RELEASE/spring-core-4.1.4.RELEASE.jar:/home/sid/.m2/repository/org/springframework/spring-beans/4.1.4.RELEASE/spring-beans-4.1.4.RELEASE.jar:/home/sid/.m2/repository/org/springframework/spring-context/4.1.4.RELEASE/spring-context-4.1.4.RELEASE.jar:/home/sid/.m2/repository/org/springframework/spring-expression/4.1.4.RELEASE/spring-expression-4.1.4.RELEASE.jar:/home/sid/.m2/repository/org/springframework/spring-aop/4.1.4.RELEASE/spring-aop-4.1.4.RELEASE.jar:/home/sid/.m2/repository/aopalliance/aopalliance/1.0/aopalliance-1.0.jar:/home/sid/.m2/repository/org/springframework/spring-orm/4.1.4.RELEASE/spring-orm-4.1.4.RELEASE.jar:/home/sid/.m2/repository/org/springframework/spring-tx/4.1.4.RELEASE/spring-tx-4.1.4.RELEASE.jar:/home/sid/.m2/repository/org/springframework/spring-jdbc/4.1.4.RELEASE/spring-jdbc-4.1.4.RELEASE.jar:/home/sid/.m2/repository/asm/asm-commons/2.2.3/asm-commons-2.2.3.jar:/home/sid/.m2/repository/asm/asm-tree/2.2.3/asm-tree-2.2.3.jar:/home/sid/.m2/repository/asm/asm/2.2.3/asm-2.2.3.jar:/home/sid/.m2/repository/asm/asm-util/2.2.3/asm-util-2.2.3.jar:/home/sid/.m2/repository/cglib/cglib-nodep/2.2/cglib-nodep-2.2.jar:/home/sid/.m2/repository/ca/uhn/hapi/hapi-base/2.0/hapi-base-2.0.jar:/home/sid/.m2/repository/commons-cli/commons-cli/1.2/commons-cli-1.2.jar:/home/sid/.m2/repository/org/jdom/jdom/1.1/jdom-1.1.jar:/home/sid/.m2/repository/xalan/xalan/2.7.0/xalan-2.7.0.jar:/home/sid/.m2/repository/ca/uhn/hapi/hapi-structures-v25/2.0/hapi-structures-v25-2.0.jar:/home/sid/.m2/repository/ca/uhn/hapi/hapi-structures-v26/2.0/hapi-structures-v26-2.0.jar:/home/sid/.m2/repository/org/javassist/javassist/3.19.0-GA/javassist-3.19.0-GA.jar:/home/sid/.m2/repository/org/hibernate/hibernate-core/4.3.9.Final/hibernate-core-4.3.9.Final.jar:/home/sid/.m2/repository/org/jboss/logging/jboss-logging/3.1.3.GA/jboss-logging-3.1.3.GA.jar:/home/sid/.m2/repository/org/jboss/logging/jboss-logging-annotations/1.2.0.Beta1/jboss-logging-annotations-1.2.0.Beta1.jar:/home/sid/.m2/repository/org/jboss/spec/javax/transaction/jboss-transaction-api_1.2_spec/1.0.0.Final/jboss-transaction-api_1.2_spec-1.0.0.Final.jar:/home/sid/.m2/repository/org/hibernate/common/hibernate-commons-annotations/4.0.5.Final/hibernate-commons-annotations-4.0.5.Final.jar:/home/sid/.m2/repository/org/hibernate/javax/persistence/hibernate-jpa-2.1-api/1.0.0.Final/hibernate-jpa-2.1-api-1.0.0.Final.jar:/home/sid/.m2/repository/antlr/antlr/2.7.7/antlr-2.7.7.jar:/home/sid/.m2/repository/org/jboss/jandex/1.1.0.Final/jandex-1.1.0.Final.jar:/home/sid/.m2/repository/org/hibernate/hibernate-c3p0/4.3.9.Final/hibernate-c3p0-4.3.9.Final.jar:/home/sid/.m2/repository/com/mchange/c3p0/0.9.2.1/c3p0-0.9.2.1.jar:/home/sid/.m2/repository/com/mchange/mchange-commons-java/0.2.3.4/mchange-commons-java-0.2.3.4.jar:/home/sid/.m2/repository/org/hibernate/hibernate-ehcache/4.3.9.Final/hibernate-ehcache-4.3.9.Final.jar:/home/sid/.m2/repository/org/hibernate/hibernate-search-orm/5.1.2.Final/hibernate-search-orm-5.1.2.Final.jar:/home/sid/.m2/repository/org/hibernate/hibernate-search-engine/5.1.2.Final/hibernate-search-engine-5.1.2.Final.jar:/home/sid/.m2/repository/org/apache/lucene/lucene-analyzers-common/4.10.4/lucene-analyzers-common-4.10.4.jar:/home/sid/.m2/repository/org/apache/lucene/lucene-facet/4.10.4/lucene-facet-4.10.4.jar:/home/sid/.m2/repository/org/apache/lucene/lucene-queryparser/4.10.4/lucene-queryparser-4.10.4.jar:/home/sid/.m2/repository/org/apache/lucene/lucene-core/4.10.4/lucene-core-4.10.4.jar:/home/sid/.m2/repository/org/apache/lucene/lucene-sandbox/4.10.4/lucene-sandbox-4.10.4.jar:/home/sid/.m2/repository/org/apache/lucene/lucene-queries/4.10.4/lucene-queries-4.10.4.jar:/home/sid/.m2/repository/dom4j/dom4j/1.6.1/dom4j-1.6.1.jar:/home/sid/.m2/repository/org/slf4j/slf4j-api/1.6.0/slf4j-api-1.6.0.jar:/home/sid/.m2/repository/org/slf4j/jcl-over-slf4j/1.6.0/jcl-over-slf4j-1.6.0.jar:/home/sid/.m2/repository/org/slf4j/slf4j-log4j12/1.6.0/slf4j-log4j12-1.6.0.jar:/home/sid/.m2/repository/com/thoughtworks/xstream/xstream/1.4.3/xstream-1.4.3.jar:/home/sid/.m2/repository/xmlpull/xmlpull/1.1.3.1/xmlpull-1.1.3.1.jar:/home/sid/.m2/repository/xpp3/xpp3_min/1.1.4c/xpp3_min-1.1.4c.jar:/home/sid/.m2/repository/javax/mail/mail/1.4.1/mail-1.4.1.jar:/home/sid/.m2/repository/javax/activation/activation/1.1/activation-1.1.jar:/home/sid/.m2/repository/org/liquibase/liquibase-core/2.0.5/liquibase-core-2.0.5.jar:/home/sid/.m2/repository/org/openmrs/liquibase/ext/modify-column/2.0.2/modify-column-2.0.2.jar:/home/sid/.m2/repository/org/openmrs/liquibase/ext/identity-insert/1.2.1/identity-insert-1.2.1.jar:/home/sid/.m2/repository/org/openmrs/liquibase/ext/type-converter/1.0.1/type-converter-1.0.1.jar:/home/sid/.m2/repository/xerces/xercesImpl/2.8.0/xercesImpl-2.8.0.jar:/home/sid/.m2/repository/javax/validation/validation-api/1.0.0.GA/validation-api-1.0.0.GA.jar:/home/sid/.m2/repository/org/hibernate/hibernate-validator/4.2.0.Final/hibernate-validator-4.2.0.Final.jar:/home/sid/.m2/repository/org/codehaus/jackson/jackson-core-asl/1.9.13/jackson-core-asl-1.9.13.jar:/home/sid/.m2/repository/org/codehaus/jackson/jackson-mapper-asl/1.9.13/jackson-mapper-asl-1.9.13.jar:/home/sid/.m2/repository/org/codehaus/groovy/groovy-all/2.4.6/groovy-all-2.4.6.jar:/home/sid/.m2/repository/net/sf/ehcache/ehcache/2.10.0/ehcache-2.10.0.jar:/home/sid/.m2/repository/org/openmrs/api/openmrs-api/2.0.5/openmrs-api-2.0.5-tests.jar:/home/sid/.m2/repository/org/springframework/spring-test/4.1.4.RELEASE/spring-test-4.1.4.RELEASE.jar:/home/sid/.m2/repository/com/h2database/h2/1.4.187/h2-1.4.187.jar:/home/sid/.m2/repository/org/databene/databene-benerator/0.5.9/databene-benerator-0.5.9.jar:/home/sid/.m2/repository/org/databene/databene-webdecs/0.4.9/databene-webdecs-0.4.9.jar:/home/sid/.m2/repository/org/apache/poi/poi/3.5-beta5/poi-3.5-beta5.jar:/home/sid/.m2/repository/org/freemarker/freemarker/2.3.9/freemarker-2.3.9.jar:/home/sid/.m2/repository/org/databene/databene-commons/0.4.9/databene-commons-0.4.9.jar:/home/sid/.m2/repository/org/databene/databene-gui/0.1.9/databene-gui-0.1.9.jar:/home/sid/.m2/repository/net/sourceforge/jtds/jtds/1.2.2/jtds-1.2.2.jar:/home/sid/.m2/repository/mysql/mysql-connector-java/5.1.6/mysql-connector-java-5.1.6.jar:/home/sid/.m2/repository/org/apache/derby/derbyclient/10.4.2.0/derbyclient-10.4.2.0.jar:/home/sid/.m2/repository/postgresql/postgresql/8.3-603.jdbc4/postgresql-8.3-603.jdbc4.jar:/home/sid/.m2/repository/org/dbunit/dbunit/2.4.7/dbunit-2.4.7.jar:/home/sid/.m2/repository/xmlunit/xmlunit/1.3/xmlunit-1.3.jar:/home/sid/.m2/repository/com/carrotsearch/junit-benchmarks/0.7.0/junit-benchmarks-0.7.0.jar:/home/sid/.m2/repository/junit/junit/4.13/junit-4.13.jar:/home/sid/.m2/repository/org/hamcrest/hamcrest-core/2.2/hamcrest-core-2.2.jar:/home/sid/.m2/repository/org/hamcrest/hamcrest/2.2/hamcrest-2.2.jar:/home/sid/.m2/repository/org/hamcrest/hamcrest-library/2.2/hamcrest-library-2.2.jar:/home/sid/.m2/repository/org/mockito/mockito-core/3.2.4/mockito-core-3.2.4.jar:/home/sid/.m2/repository/net/bytebuddy/byte-buddy/1.10.5/byte-buddy-1.10.5.jar:/home/sid/.m2/repository/net/bytebuddy/byte-buddy-agent/1.10.5/byte-buddy-agent-1.10.5.jar:/home/sid/.m2/repository/org/objenesis/objenesis/2.6/objenesis-2.6.jar:/home/sid/.m2/repository/org/powermock/powermock-module-junit4/2.0.4/powermock-module-junit4-2.0.4.jar:/home/sid/.m2/repository/org/powermock/powermock-module-junit4-common/2.0.4/powermock-module-junit4-common-2.0.4.jar:/home/sid/.m2/repository/org/powermock/powermock-reflect/2.0.4/powermock-reflect-2.0.4.jar:/home/sid/.m2/repository/org/powermock/powermock-core/2.0.4/powermock-core-2.0.4.jar:/home/sid/.m2/repository/org/powermock/powermock-api-mockito2/2.0.4/powermock-api-mockito2-2.0.4.jar:/home/sid/.m2/repository/org/powermock/powermock-api-support/2.0.4/powermock-api-support-2.0.4.jar com.intellij.rt.junit.JUnitStarter -ideVersion5 -junit4 org.openmrs.module.fhir2.api.dao.impl.FhirObservationDaoImplTest,searchForObs_shouldSearchForObsByValueQuantity
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.springframework.cglib.core.ReflectUtils$2 (file:/home/sid/.m2/repository/org/springframework/spring-core/4.1.4.RELEASE/spring-core-4.1.4.RELEASE.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of org.springframework.cglib.core.ReflectUtils$2
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

java.lang.ClassCastException: class java.lang.String cannot be cast to class java.lang.Double (java.lang.String and java.lang.Double are in module java.base of loader 'bootstrap')

	at org.hibernate.type.descriptor.java.DoubleTypeDescriptor.unwrap(DoubleTypeDescriptor.java:36)
	at org.hibernate.type.descriptor.sql.DoubleTypeDescriptor$1.doBind(DoubleTypeDescriptor.java:63)
	at org.hibernate.type.descriptor.sql.BasicBinder.bind(BasicBinder.java:90)
	at org.hibernate.type.AbstractStandardBasicType.nullSafeSet(AbstractStandardBasicType.java:286)
	at org.hibernate.type.AbstractStandardBasicType.nullSafeSet(AbstractStandardBasicType.java:281)
	at org.hibernate.loader.Loader.bindPositionalParameters(Loader.java:1995)
	at org.hibernate.loader.Loader.bindParameterValues(Loader.java:1966)
	at org.hibernate.loader.Loader.prepareQueryStatement(Loader.java:1901)
	at org.hibernate.loader.Loader.executeQueryStatement(Loader.java:1862)
	at org.hibernate.loader.Loader.executeQueryStatement(Loader.java:1839)
	at org.hibernate.loader.Loader.doQuery(Loader.java:910)
	at org.hibernate.loader.Loader.doQueryAndInitializeNonLazyCollections(Loader.java:355)
	at org.hibernate.loader.Loader.doList(Loader.java:2554)
	at org.hibernate.loader.Loader.doList(Loader.java:2540)
	at org.hibernate.loader.Loader.listIgnoreQueryCache(Loader.java:2370)
	at org.hibernate.loader.Loader.list(Loader.java:2365)
	at org.hibernate.loader.criteria.CriteriaLoader.list(CriteriaLoader.java:126)
	at org.hibernate.internal.SessionImpl.list(SessionImpl.java:1718)
	at org.hibernate.internal.CriteriaImpl.list(CriteriaImpl.java:380)
	at org.openmrs.module.fhir2.api.dao.impl.FhirObservationDaoImpl.searchForObservations(FhirObservationDaoImpl.java:77)
	at org.openmrs.module.fhir2.api.dao.impl.FhirObservationDaoImplTest.searchForObs_shouldSearchForObsByValueQuantity(FhirObservationDaoImplTest.java:467)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:567)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.springframework.test.context.junit4.statements.RunBeforeTestMethodCallbacks.evaluate(RunBeforeTestMethodCallbacks.java:73)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.springframework.test.context.junit4.statements.RunAfterTestMethodCallbacks.evaluate(RunAfterTestMethodCallbacks.java:82)
	at org.springframework.test.context.junit4.statements.SpringRepeat.evaluate(SpringRepeat.java:73)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.runChild(SpringJUnit4ClassRunner.java:217)
	at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.runChild(SpringJUnit4ClassRunner.java:83)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.springframework.test.context.junit4.statements.RunBeforeTestClassCallbacks.evaluate(RunBeforeTestClassCallbacks.java:61)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.springframework.test.context.junit4.statements.RunAfterTestClassCallbacks.evaluate(RunAfterTestClassCallbacks.java:68)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.run(SpringJUnit4ClassRunner.java:163)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:33)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:230)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:58)


Process finished with exit code 255

```

VaishSiddharth: @ibacher Found the solution for this. Thanks for the help

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
VaishSiddharth: @ibacher Ok I will change this. Is everything else fine, should I complete the code & tests and update the PR?

ibacher: I haven't really gone through everything else here, but please do keep this up to date

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
VaishSiddharth: @ibacher Please have a look

ibacher: I actually don't think that you need to turn the `has-member.code` parameter into a codeable concept. Just refactor out the relevant part of the `handleCodedConcept` method from `FhirObservationDaoImpl` into `BaseDaoImpl` and call that.

VaishSiddharth: @ibacher I think that we'll have to handle it by doing something like this

```java
protected void handleHasMemberReference(Criteria criteria, ReferenceParam hasMemberReference) {
		if (hasMemberReference != null) {
			criteria.createAlias("obsGroup", "og");

			if (hasMemberReference.getChain() != null) {
				switch (hasMemberReference.getChain()) {
					case Observation.SP_CODE:
						criteria.add(eq("og.concept", hasMemberReference.getValue()));
						break;
					case "":
						criteria.add(eq("og.uuid", hasMemberReference.getValue()));
						break;
				}
			}
		}
	}
```

but I am not sure that should we use `obsGroup` or `groupMembers`

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
```diff
@@ -73,6 +111,26 @@ protected String paramToProp(@NotNull String paramName) {
 		}
 	}
 	
+	private void handleValueCodedConcept(Criteria criteria, TokenAndListParam valueConcept) {
+		if (valueConcept != null) {
+			criteria.createAlias("valueCoded", "c");
+			
+			handleAndListParamBySystem(valueConcept, (system, tokens) -> {
```
ibacher: Rather than duplicating code all over the place, we should look for ways to create common methods, e.g. adding something like this to `BaseDaoImpl`:

```
protected Optional<Criterion> findMatchingConcepts(@NotNull String conceptAlias, TokenAndListParam concepts) {
    if (concepts == null) {
         return Optional.empty();
    }

    return handleAndListParamBySystem(concepts, (system, tokens) -> {
	    if (system.isEmpty()) {
			return Optional.of(
			    or(in(String.format("%.conceptId", conceptAlias), tokensToParams(tokens).map(NumberUtils::toInt).collect(Collectors.toList())),
			        in(String.format("%.uuid", conceptAlias), tokensToList(tokens))));
		} else {
			if (!containsAlias(criteria, "cm")) {
				criteria.createAlias(String.format("%.conceptMappings", conceptAlias), "cm").createAlias("cm.conceptReferenceTerm", "crt");
			}

			return Optional.of(generateSystemQuery(system, tokensToList(tokens)));
		}
	}
}
```

And then just changing this to:

```java
private void handleValueCodedConcept(Criteria criteria, TokenAndListParam valueConcepts) {
	if (valueConcept != null) {
		criteria.createAlias("valueCoded", "c");
        findMatchingConcepts("c", valueConcepts).ifPresent(criteria::add);
    }
}
```

Do similar things for `FhirObservationDaoImpl` and `FhirAllergyIntoleranceDaoImpl`.

This would also address my comment above about refactoring.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
VaishSiddharth: @ibacher Isn't it better to have only one parameter as `hasMember` and through that we could search by `hasMember.code` as well. Will the above code not be an good option to use?

ibacher: So the property you should use is `groupMembers`. And, you're right that we'll need two branches here, but the handling for `SP_CODE` will need to be a bit more complex to handle:

`/ws/fhir2/Observation?has-member.code=http://loinc.org|8854-6`

😄

ibacher: @VaishSiddharth Do you see what I'm getting at here?

VaishSiddharth: @ibacher Yes you mean to say that the `code` part will be a bit difficult to handle, so no problem in that but can't we include that part in this handle function itself?

``` java 
protected void handleHasMemberReference(Criteria criteria, ReferenceParam hasMemberReference) {
		if (hasMemberReference != null) {
			criteria.createAlias("obsGroup", "og");

			if (hasMemberReference.getChain() != null) {
				switch (hasMemberReference.getChain()) {
					case Observation.SP_CODE:

//Call a function similar to handleValueCodedConcept
						criteria.add(eq("og.concept", hasMemberReference.getValue()));
						break;
					case "":
						criteria.add(eq("og.uuid", hasMemberReference.getValue()));
						break;
				}
			}
		}
	}
```

ibacher: Yes, that's the idea!

varung-31:
@VaishSiddharth  There are some conflicts with the master branch. Please pull the changes before you publish the next revision.

ibacher:
@VaishSiddharth Could you resolve these conflicts so that the CI build can complete? Thanks!

api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImplTest.java
```diff
@@ -387,4 +415,83 @@ public void searchForObs_shouldHandleComplexQuery() {
...
+		assertThat(results, notNullValue());
+		// This gives java.lang.AssertionError:
+		//Expected: not an empty collection
+		//     but: was <[]>
+		//Expected :not an empty collection
+		//Actual   :<[]>,
+		// I tried adding data to xml file but not relevant tags found similar to groupMember
+		//		assertThat(results, not(empty()));
+		//		assertThat(results.iterator().next().getUuid(), equalTo(OBS_UUID));
+		//		assertThat(results.iterator().next().getGroupMembers().iterator().next().getUuid(), equalTo(OBS_UUID));
```
VaishSiddharth: @ibacher Please help me with this error

ibacher: So `MEMBER_UUID` should be referring to the UUID that is the member of an obs group. An `Obs` is a member of an obs group when the `obs_group_id` column points to the id of another observation.

VaishSiddharth: @ibacher I added this and tried pointing to both the ids still the same error
```xml
<obs obs_id="22" person_id="7" concept_id="5092" encounter_id="3" obs_datetime="2008-07-01 00:00:00.0" location_id="1" value_numeric="98.3" comments="" creator="1" date_created="2008-08-18 14:09:35.0" voided="false" value_coded="[NULL]" value_coded_name_id="[NULL]" value_complex="[NULL]" value_text="[NULL]" value_datetime="[NULL]" value_drug="[NULL]" uuid="4efa62d2-6b8b-4803-a8fa-3f32ee54db4f"/>
    <obs obs_id="23" person_id="7" obs_group_id="22" concept_id="5242" encounter_id="3" obs_datetime="2008-07-01 00:00:00.0" location_id="1" value_numeric="14.0" comments="" creator="1" date_created="2008-08-18 14:09:35.0" voided="false" value_coded="[NULL]" value_coded_name_id="[NULL]" value_complex="[NULL]" value_text="[NULL]" value_datetime="[NULL]" value_drug="[NULL]" uuid="744b91f8-bdbc-4950-833b-002244e9fa2b"/>
```

ibacher: When you say "pointing to both ids", could you spell out what you mean?

VaishSiddharth: i tried both

``` java
memberReference.setValue("744b91f8-bdbc-4950-833b-002244e9fa2b");
		memberReference.setChain("");
```
THIS FAILED SO I TRIED THE BELOW ONE AND THAT ALSO FAILED
```java
memberReference.setValue("4efa62d2-6b8b-4803-a8fa-3f32ee54db4f");
		memberReference.setChain("");
```

ibacher: The first one looks right to me. Could you make those changes to this branch and push it out? That way we can both be looking at the same code.

VaishSiddharth: @ibacher Any ideas what's wrong here?

ibacher:
@VaishSiddharth Good job on figuring out that bug! This looks very good. I've got one more wrinkle for you on this:

Per the FHIR spec, quantity can be searched using a number of prefixes, similar to what date supports. See [here](https://www.hl7.org/fhir/search.html#prefix) for the spec.

We should add support for the `eq`, `ne`, `gt`, `lt`, `ge`, `le` prefixes. Note that unlike for date ranges, no prefix might be supplied, in which case, we should assume `eq`.

VaishSiddharth:
@ibacher So I guess this needs to be done universally by that I mean i'll have to change to `handleDateRange` and `handleDate` methods in BaseDAO. Am I right?

ibacher:
@VaishSiddharth I like the initiative, but in this case, no. Dates have some funky handling rules that don't make sense for quantities, e.g. `date=eq2020-03-13` matches 2020-03-13 00:00:00 AM and 2020-03-13 11:57:22 AM, etc.

ibacher:
There are also some separate funky rules for quantities, but I'm ignoring those for now.

ibacher:
@VaishSiddharth This PR is getting a bit hard to follow because it's got changes that are already in the main module. Could you kindly rebase your changes on the latest master so we have a relatively clean history (this may need to involve reordering some commits)

ibacher:
@VaishSiddharth Also two more things to implement:

1. The prefix `ap` for approximate. The search should return values that are within 10% of the value, i.e., a search for `value-quantity=ap2.0` should return any values between 1.8 and 2.2.
2. The really difficult one the implicit range of returned values based on the searched-for value. The FHIR spec says that numbers are interpreted as having an implicit range up to the number of up to half the number of significant digits. The examples they provide are drawn from [here](https://www.hl7.org/fhir/search.html#number):

 * 100 -> [99.5 ... 100.5)
 * 100.00 -> [99.995 ... 100.005]
 * 1e2 -> [95 ... 105]

We should implement this for searches where no prefix is provided. For reference, the HAPI FHIR JPA Server implements both of these requirements in [this interface](https://github.com/jamesagnew/hapi-fhir/blob/578f2b7aef725083cb26bf004df029147f4b8b58/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/SearchFuzzUtil.java), which might serve as inspiration for how this might be coded.

Am-Coder:
@VaishSiddharth Please make sure you squash your commits into a single commit once your PR becomes ready to be merged.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -433,7 +434,25 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
 		
 		if (quantityParam.getPrefix() == null) {
-			return Optional.of(eq(propertyName, quantityParam.getValue().doubleValue()));
+			String plainString = quantityParam.getValue().toPlainString();
+			int dotIdx = plainString.indexOf('.');
+			int eIdx = plainString.indexOf('e');
+			if (dotIdx == -1 && eIdx == -1) {
+				return Optional.of(or(ge(propertyName, quantityParam.getValue().subtract(new BigDecimal(0.1)).doubleValue()),
+				    lt(propertyName, quantityParam.getValue().add(new BigDecimal(0.1)).doubleValue())));
+			} else if (eIdx != -1) {
```
VaishSiddharth: @ibacher Not very sure how to handle 1e2 type of numbers please have a look

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -433,7 +434,25 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
+				int precision = plainString.length() - (eIdx);
+				double mul = Math.pow(10, precision);
+				double val = mul * 5.0d;
+				return Optional.of(or(ge(propertyName, quantityParam.getValue().subtract(new BigDecimal(val)).doubleValue()),
+				    lt(propertyName, quantityParam.getValue().add(new BigDecimal(val)).doubleValue())));
```
VaishSiddharth: @ibacher Not very sure how to handle 1e2 type of numbers please have a look

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -424,6 +428,59 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
+		if (quantityParam == null) {
+			return Optional.empty();
+		}
+		
+		if (quantityParam.getPrefix() == null) {
+			String plainString = quantityParam.getValue().toPlainString();
+			int dotIdx = plainString.indexOf('.');
+			int eIdx = plainString.indexOf('e');
+			if (dotIdx == -1 && eIdx == -1) {
+				return Optional.of(or(ge(propertyName, quantityParam.getValue().subtract(new BigDecimal(0.1)).doubleValue()),
```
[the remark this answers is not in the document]
ibacher: Of course, better than using `and` or `or` here would be to use `between`

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -424,6 +428,59 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
+				    lt(propertyName, quantityParam.getValue().add(new BigDecimal(0.1)).doubleValue())));
+			} else if (eIdx != -1) {
+				int precision = plainString.length() - (eIdx);
+				double mul = Math.pow(10, precision);
+				double val = mul * 5.0d;
+				return Optional.of(or(ge(propertyName, quantityParam.getValue().subtract(new BigDecimal(val)).doubleValue()),
```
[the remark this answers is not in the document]
ibacher: Of course, better than using `and` or `or` here would be to use `between`

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -424,6 +428,59 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
+				    lt(propertyName, quantityParam.getValue().add(new BigDecimal(val)).doubleValue())));
+			} else {
+				int precision = plainString.length() - (dotIdx);
+				double mul = Math.pow(10, -precision);
+				double val = mul * 5.0d;
+				return Optional.of(or(ge(propertyName, quantityParam.getValue().subtract(new BigDecimal(val)).doubleValue()),
```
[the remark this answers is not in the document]
ibacher: Of course, better than using `and` or `or` here would be to use `between`

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -424,6 +428,59 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
+				    lt(propertyName, quantityParam.getValue().add(new BigDecimal(val)).doubleValue())));
```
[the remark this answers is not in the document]
VaishSiddharth: @ibacher So [here](https://www.hl7.org/fhir/search.html#number) I saw something like 

[99.995 ... 100.005) 

where ) means not included that's why I used lt and not le
Please let me know what's correct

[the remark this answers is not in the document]
ibacher: Good point

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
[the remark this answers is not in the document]
VaishSiddharth: @ibacher I am not sure not to test this out. The problem that I am facing is that I can't get the range of a quantityParam. Please tell how should I approach this?
> My advice would be to write a test case using "1e2" and see what happens.

``` java
@Test
	public void shouldGetObservationsByValueQuantityWithPrefix() throws Exception {
		verifyUri("/Observation?value-quantity=1e2");
		
		verify(observationService).searchForObservations(isNull(), isNull(), isNull(), isNull(), isNull(),
		    valueQuantityCaptor.capture(), isNull(), isNull(), isNull(), isNull());
		
		assertThat(valueQuantityCaptor.getValue(), notNullValue());
		assertThat(valueQuantityCaptor.getValue().getValuesAsQueryTokens(), not(empty()));
		assertThat(valueQuantityCaptor.getAllValues().iterator().next().getValuesAsQueryTokens().iterator().next()
		        .getValuesAsQueryTokens().iterator().next().getValueAsString(),
		    equalTo("100"));
	}
```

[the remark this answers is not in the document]
ibacher: So, for our purposes, I'm less concerned that we map "1e2" to [95 ... 105) than I am that we map 1e-2 to [.005 ... .015), i.e., even if we're not matching the precision implied by the spec, I think we're ok. Properly support 1e2 would, I think, require us to go a little crazy writing a string to number parser.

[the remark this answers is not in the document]
VaishSiddharth: @ibacher I need a way to test these ranges then only I'll be able to code something up for 1e-2 like numbers. Please suggest a way to do so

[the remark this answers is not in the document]
ibacher: So, the quickest way to test is to search (at the DAO level) for an Obs with a quantity of "36". There's exactly one Obs in the dataset that should fall in that range (Obs 7e77d071-1f7d-4394-a316-0a458edf28c3; it represents body temperature). Then I'd do a search a search for 39.5. That should have a single result, (Obs 1a9b5242-0ff9-4c5c-adb3-d517632ea63e). That should show that it finds values within ±10%.

api/src/test/resources/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImplTest_initial_data_suppl.xml
```diff
@@ -61,4 +61,6 @@
...
-</dataset>
+    <obs obs_id="25" person_id="7" concept_id="5242" encounter_id="3" obs_datetime="2008-07-01 00:00:00.0" location_id="1" value_numeric="100.001" comments="" creator="1" date_created="2008-08-18 14:09:35.0" voided="false" value_coded="[NULL]" value_coded_name_id="[NULL]" value_complex="[NULL]" value_text="[NULL]" value_datetime="[NULL]" value_drug="[NULL]" uuid="89fg071-1f7d-4394-a316-0a458edf28c3"/>
+    <obs obs_id="26" person_id="7" concept_id="5242" encounter_id="3" obs_datetime="2008-07-01 00:00:00.0" location_id="1" value_numeric="0.007" comments="" creator="1" date_created="2008-08-18 14:09:35.0" voided="false" value_coded="[NULL]" value_coded_name_id="[NULL]" value_complex="[NULL]" value_text="[NULL]" value_datetime="[NULL]" value_drug="[NULL]" uuid="56htgf-1f7d-4394-a316-0a458edf28c3"/>
+</dataset>
```
ibacher: Please leave a new line at the end of the file

api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImplTest.java
```diff
@@ -477,21 +478,94 @@ public void searchForObs_shouldSearchForObsByValueDate() {
...
+	//
+	//		assertThat(results, notNullValue());
+	//		assertThat(results, not(empty()));
+	//		assertThat(results, hasItem(hasProperty("uuid", equalTo("4efa62d2-6b8b-4803-a8fa-3f32ee54db4f"))));
+	//	}
+	//
+	//	@Test
+	//	public void searchForObs_shouldSearchForObsByValueQuantityWithoutPrefixAndNegativeEValue() {
+	//		QuantityAndListParam quantityAndListParam = new QuantityAndListParam();
+	//
```
VaishSiddharth: @ibacher So I have added some tests which don't work at the moment because the prefix coding is not correct for `1e2` and `1e-2` type of numbers.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -440,8 +440,8 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
 			int eIdx = plainString.indexOf('e');
 			if (dotIdx == -1 && eIdx == -1) {
 				return Optional
-				        .of(between(propertyName, quantityParam.getValue().subtract(new BigDecimal(0.1)).doubleValue(),
-				            quantityParam.getValue().add(new BigDecimal(0.1)).doubleValue()));
+				        .of(between(propertyName, quantityParam.getValue().subtract(new BigDecimal(0.5)).doubleValue(),
+				            quantityParam.getValue().add(new BigDecimal(0.5)).doubleValue()));
 			} else if (eIdx != -1) {
```
VaishSiddharth: @ibacher We'll need to correct this part

api/src/test/resources/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImplTest_initial_data_suppl.xml
VaishSiddharth: I'll do that in the next commit

api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImplTest.java
```diff
@@ -477,21 +478,94 @@ public void searchForObs_shouldSearchForObsByValueDate() {
...
+	//		QuantityParam quantityParam = new QuantityParam();
+	//		quantityParam.setValue("1e2");
+	//		quantityAndListParam.addAnd(quantityOrListParam.add(quantityParam));
+	//
+	//		Collection<Obs> results = dao.searchForObservations(null, null, null, null, null, quantityAndListParam, null, null,
+	//		    null, null);
+	//
+	//		assertThat(results, notNullValue());
+	//		assertThat(results, not(empty()));
+	//		assertThat(results, hasItem(hasProperty("uuid", equalTo("4efa62d2-6b8b-4803-a8fa-3f32ee54db4f"))));
```
ibacher: This has the wrong UUID, since we elected not to implement the approximation range without a decimal point, the search range for `1e2` is [99.5 ... 100.5], which is fine, but only UUID 89fg071-1f7d-4394-a316-0a458edf28c3 falls into that range. The other test works for me.

VaishSiddharth: But [here](https://www.hl7.org/fhir/search.html#number) I couldn't find the same range

[parameter]=1e2 | Values that equal 100, to 1 significant figures precision, so this is actually searching for values in the range [95 ... 105)
-- | --

Does 1e-2 works for you?

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
ibacher: So, `new BigDecimal(0.1)` is closer to what we want. What we actually want is this:

```java
// At the top of the file
private static final BigDecimal APPROX_RANGE = new BigDecimal(0.1);

BigDecimal value = quantityParam.getValue();
BigDecimal approxRange = APPROX_RANGE.multiply(value);

return Optional
    .of(between(propertyName, value.subtract(approxRange).doubleValue(), value.add(approxRange).doubleValue()));

File: api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImplTest.java (hunk 17 above)
Comment (ibacher) [reply api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImplTest.java]: I'm aware. This is a point where I'm suggesting we break spec, because the alternative seems to be to write a parser for scientific notation. Maybe we'll reassess later, but I don't think that's necessary at this point in time. And, yes, 1e-2 worked for me without modifications...

File: api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImplTest.java (hunk 15 above)
Comment (ibacher) [reply api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImplTest.java]: So, for prefix coding, I'm happy with `eq` being an exact equality, e.g. `1e2` means only exactly `100` (despite the spec, this is the actual implementation in the HAPI JPA Server). `ap` should still work, since `1e2` should still translate to `[90 ... 100]`. `le`, `ge`, `lt`, and `gt` are all defined as starting from the exact position at infinite precision, so again should be covered by just the normal Hibernate operations.

Comment (VaishSiddharth) [reply api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImplTest.java]: @ibacher I have pushed a commit after this please have a look and tell if anything needs to be fixed

**Comment by ibacher** [comment]
@VaishSiddharth This looks great! We just need some more tests for Quantity. In particular:

1. A Quantity with no decimal a not in scientific notation, e.g. "100"

2. A test for the various prefixes aside from "ap". Note in particular that we want to be sure that "eq" returns only exact matches, so, 100 should match an Obs with a value of 100, but 100.01 shouldn't.

Otherwise, I think this is ready to go. Thanks for your work on this!

**Comment by ibacher** [comment]
@VaishSiddharth Thanks for the update, but please do cover all of the prefixes. I know it's not the most glamorous work, but it needs to be done.

**Comment by VaishSiddharth** [comment]
@ibacher Should I squash all the commits now?

File: api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java [hunk 18]
```diff
@@ -424,6 +431,62 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
+			if (dotIdx == -1 && eIdx == -1) {
+				return Optional.of(
+				    between(propertyName, value.subtract(approxRange).doubleValue(), value.add(approxRange).doubleValue()));
+			} else if (eIdx != -1) {
+				int precision = plainString.length() - (eIdx);
+				double mul = Math.pow(10, precision);
+				double val = mul * 5.0d;
+				return Optional
+				        .of(between(propertyName, quantityParam.getValue().subtract(new BigDecimal(val)).doubleValue(),
+				            quantityParam.getValue().add(new BigDecimal(val)).doubleValue()));
```
ibacher: Remove this whole `else` block

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -424,6 +431,62 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
+			return Optional.empty();
+		}
+		
+		BigDecimal value = quantityParam.getValue();
+		BigDecimal approxRange = APPROX_RANGE.multiply(value);
+		
+		if (quantityParam.getPrefix() == null) {
+			String plainString = quantityParam.getValue().toPlainString();
+			int dotIdx = plainString.indexOf('.');
+			int eIdx = plainString.indexOf('e');
```
ibacher: We don't really need this

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -424,6 +431,62 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
 		return Optional.empty();
 	}
 	
+	protected Optional<Criterion> handleQuantity(String propertyName, QuantityParam quantityParam) {
+		if (quantityParam == null) {
+			return Optional.empty();
+		}
+		
+		BigDecimal value = quantityParam.getValue();
+		BigDecimal approxRange = APPROX_RANGE.multiply(value);
```
ibacher: This should be pushed down to where `dotIdx == -1` since that's the only place we need this value

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -424,6 +431,62 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
+				double val = mul * 5.0d;
+				return Optional
+				        .of(between(propertyName, quantityParam.getValue().subtract(new BigDecimal(val)).doubleValue(),
+				            quantityParam.getValue().add(new BigDecimal(val)).doubleValue()));
+			} else {
+				int precision = plainString.length() - (dotIdx);
+				double mul = Math.pow(10, -precision);
+				double val = mul * 5.0d;
+				return Optional
+				        .of(between(propertyName, quantityParam.getValue().subtract(new BigDecimal(val)).doubleValue(),
```
ibacher: Substitute `value` for `quantityParam.getValue()`

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -424,6 +431,62 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
+				            quantityParam.getValue().add(new BigDecimal(val)).doubleValue()));
+			}
+		} else {
+			switch (quantityParam.getPrefix()) {
```
ibacher: `double val = value.getValue()` then use val below

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -424,6 +431,62 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
+					return Optional.of(lt(propertyName, quantityParam.getValue().doubleValue()));
+				case GREATERTHAN_OR_EQUALS:
+					return Optional.of(ge(propertyName, quantityParam.getValue().doubleValue()));
+				case GREATERTHAN:
+					return Optional.of(gt(propertyName, quantityParam.getValue().doubleValue()));
+				case APPROXIMATE:
+					return Optional.of(between(propertyName,
+					    quantityParam.getValue().subtract(quantityParam.getValue().multiply(new BigDecimal(0.1)))
+					            .doubleValue(),
+					    quantityParam.getValue().add(quantityParam.getValue().multiply(new BigDecimal(0.1))).doubleValue()));
```
ibacher: Remove this block

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -424,6 +431,62 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
 	
+	protected Optional<Criterion> handleQuantity(String propertyName, QuantityParam quantityParam) {
+		if (quantityParam == null) {
+			return Optional.empty();
+		}
+		
+		BigDecimal value = quantityParam.getValue();
+		BigDecimal approxRange = APPROX_RANGE.multiply(value);
+		
+		if (quantityParam.getPrefix() == null) {
```
ibacher: Change this to `quantityParam.getPrefix() == null || quantityParam.getPrefix.equals(APPROXIMATE)`

ibacher:
@VaishSiddharth if you can run `mvn clean package`, commit whatever formatting changes are there and then squash it down, I think this is good to go. (I'm assuming the compilation failures are transitory).

VaishSiddharth:
@ibacher I guess the last commit you made to the master gives a build failure

ibacher:
@VaishSiddharth You're absolutely right! See, I told you we all make stupid mistakes!

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
```diff
@@ -44,17 +50,62 @@ public Obs getObsByUuid(String uuid) {
...
+		}
+	}
+	
+	protected Optional<Criterion> handleValueQuantityParam(@NotNull String propertyName,
+	        QuantityAndListParam valueQuantityParam) {
+		if (valueQuantityParam == null) {
+			return Optional.empty();
+		}
+		return handleAndListParam(valueQuantityParam, quantityParam -> handleQuantity(propertyName, quantityParam));
+	}
```
ibacher: Can we refactor thins so this method is also in BaseDaoImpl

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
```diff
@@ -44,17 +50,62 @@ public Obs getObsByUuid(String uuid) {
...
+	
+	protected Optional<Criterion> handleValueStringParam(@NotNull String propertyName, StringAndListParam valueStringParam) {
+		if (valueStringParam == null) {
+			return Optional.empty();
+		}
+		return handleAndListParam(valueStringParam, v -> Optional.of(ilike(propertyName, v.getValue(), MatchMode.EXACT)));
```
ibacher: Is there any reason to prefer this to `propertyLike`?

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -539,6 +589,29 @@ protected void handleParticipantReference(Criteria criteria, ReferenceAndListPar
 		}
 	}
 	
+	protected Optional<Criterion> findMatchingConcepts(@NotNull String conceptAlias, TokenAndListParam concepts,
```
ibacher: Change this to `handleCodeableConcept`

ibacher: Also change the signature to match `handleResourceCode`

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -539,6 +589,29 @@ protected void handleParticipantReference(Criteria criteria, ReferenceAndListPar
...
+				    in(String.format("%s.uuid", conceptAlias), tokensToList(tokens))));
+			} else {
+				if (!containsAlias(criteria, "cm")) {
+					criteria.createAlias(String.format("%s.conceptMappings", conceptAlias), "cm")
+					        .createAlias("cm.conceptReferenceTerm", "crt");
+				}
+				
+				return Optional.of(generateSystemQuery(system, tokensToList(tokens)));
+			}
+		});
```
ibacher: Can we refactor out the lambda in here (the expression beginning `(system, tokens) ->`) and turn it into a proper method which can be called both here and `handleResourceCode`

ibacher: Also refactor `handleResourceCode` to be called `handleCodeableConcept`. Hopefully we can eventually get rid of the `TokenOrListParam` version of that.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
```diff
@@ -44,17 +50,62 @@ public Obs getObsByUuid(String uuid) {
...
+						break;
+					case "":
+						criteria.add(eq("gm.uuid", hasMemberReference.getIdPart()));
+						break;
+				}
+			}
+		}
+	}
+	
+	protected Optional<Criterion> handleValueQuantityParam(@NotNull String propertyName,
```
ibacher: I'd rename this `handleQuantity()` and move it to BaseDaoImpl. I'd also rename the parameter `quantityAndListParam`.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
```diff
@@ -44,17 +50,62 @@ public Obs getObsByUuid(String uuid) {
...
 		handleSort(criteria, sort);
 		
 		return criteria.list();
 	}
 	
+	protected void handleHasMemberReference(Criteria criteria, ReferenceParam hasMemberReference) {
+		if (hasMemberReference != null) {
+			criteria.createAlias("groupMembers", "gm");
+			
+			if (hasMemberReference.getChain() != null) {
```
ibacher: This is an unnecessary check.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
VaishSiddharth: @ibacher So I guess we'll need to change the parameter either here or in handleResourceCode to make it `TokenOrListParam` or `TokenAndListParam`

ibacher: Well, I'd like to keep them separate for now so that anything that's relying on `TokenOrListParam` doesn't break and we keep the scope of this PR relatively focused. We can have two method with the same name that are distinguished by parameter types, so that shouldn't be an issue. We just need to make these two resemble each other so it's clear that they provide the same functionality.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -792,9 +804,8 @@ protected Criterion generateSystemQuery(String system, List<String> codes) {
 		}
 	}
 	
-	protected void handleResourceCode(Criteria criteria, TokenOrListParam code, @NotNull String conceptAlias,
+	protected void handleCode(Criteria criteria, TokenOrListParam code, @NotNull String conceptAlias,
```
ibacher: Can this also be called `handleCodeableConcept()`? Also, can we move it so it's with the other `handleCodeableConcept` method?

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -812,6 +823,12 @@ protected void handleResourceCode(Criteria criteria, TokenOrListParam code, @Not
 		}).ifPresent(criteria::add);
 	}
 	
+	protected void handleResourceCode(Criteria criteria, TokenOrListParam code, @NotNull String conceptAlias,
+	        @NotNull String conceptMapAlias, @NotNull String conceptReferenceTermAlias) {
+		
+		handleCode(criteria, code, conceptAlias, conceptMapAlias, conceptReferenceTermAlias);
+	}
```
ibacher: Let's get rid of this altogether, even though that means updating more files.

VaishSiddharth: @ibacher I'll do that but it'll take some time

VaishSiddharth:
@ibacher Should I squash all the commits?

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -539,6 +597,33 @@ protected void handleParticipantReference(Criteria criteria, ReferenceAndListPar
 		}
 	}
 	
+	protected Optional<Criterion> handleCode(Criteria criteria, TokenAndListParam concepts, @NotNull String conceptAlias) {
```
ibacher: I don't think we need a separate method for this any more. Just inline this for `handleCodeableConcept`. However, this function *should* still take the `conceptMapAlias` and `conceptReferenceTermAlias` parameters instead of hard-coding them as `cm` and `crt`.

ibacher:
@VaishSiddharth While I'm not upset or anything, that was way more changes than I was intending for you to make. I guess it's a good thing, since all the tests still pass. I would've preferred migrating the other resources to using `TokenAndListParam` one-by-one so we could add appropriate tests for that functionality, however.

That said, I guess this is fine with the change requested. Can you create a ticket to cover adding tests for the `TokenAndListParam`s that have been added?

VaishSiddharth:
@ibacher Oh I am really sorry if I have messed up on that. I should have through about simply making a TokenAndListParam and adding the TokenOrListParam to it and then handling it. If you wish I could revert back

ibacher:
@VaishSiddharth No... I mean we were going to move towards that any way, so we might as well go forward!

ibacher:
@VaishSiddharth I think we're (finally) good here! Thanks for all your work!

PR #102, opened 2020-03-03
FM2-100: Return an empty list when searching by an invalid gender code

CaptainDaVinci:

Return an empty list when searching by an invalid gender code.

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-100

  No? -> execute above command


api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirPatientDaoImplTest.java
```diff
@@ -271,7 +271,7 @@ public void searchForPatients_shouldReturnPatientsByGender() {
 	@Test
 	public void searchForPatients_shouldReturnEmptyCollectionWhenGenderNotMatched() {
 		Collection<Patient> results = dao.searchForPatients(null, null, null, null,
-		    new TokenOrListParam().add(new TokenParam("other")), null, null, null, null, null, null, null, null);
+		    new TokenOrListParam().add(new TokenParam("wrong-gender")), null, null, null, null, null, null, null, null);
```
ibacher: We should probably extract this out into a constant

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDaoImpl.java
```diff
@@ -440,7 +440,7 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
 			}
```
ibacher: We should also address the `Optional.empty()` above this line.

CaptainDaVinci: I have a couple of questions here, 

1. The code `OTHER` has a mapping to `'O'`, as shown in the `GenderTranslatorImplTest`. Keeping that in mind, should we change the Criterion to match `'O'` specifically? Because the code `UNKNOWN` too has a mapping to `'U'`.

2. Should `NULL` match any resource or those which have gender set to empty or null (if that's even possible)?

ibacher: So remember what we're mapping to here: we're querying the OpenMRS database for people matching whatever gender. In OpenMRS, there are really just three possible gender codes `M`, `F` and `null`.  Maybe a better way to support this is to map all of `OTHER`, `UNKNOWN`, and `NULL` to `gender is null`. We should probably fix the gender translator to reflect this as well. I suppose it should just return `M` for "male", `F` for "female" and `null` for "unknown".

ibacher:
@tendomart That was my fault! @CaptainDaVinci please rebase this commit on the current master.

tendomart:
Ok alright thanks for the info @ibacher

ibacher:
@CaptainDaVinci This looks good so far. I think we need tests for the genders `OTHER`, `UNKNOWN` and `NULL` at the DAO level to ensure that they behave properly and this can be merged. Thanks for your work on this!

CaptainDaVinci:
@ibacher I made a few more changes to the `handleGender` method. 

1. The gender code `OTHER` is mapped to resources with gender set to `nulll`.
2. Looking into the `AdministrativeGender.fromCode` method, the code `NULL` is not mapped to `null` value, hence an explicit check is added.
3.  The tests for `PersonDaoImpl` and `PatientDao` now check for if every result item has the matching gender.

P.S: Additional patient data may need to be added to the core data set, to test for `UNKNOWN`, `NULL` and `OTHER` gender code.

PR #103, opened 2020-03-03
FM2-96: Implement additional fields for observations

jecihjoy:
Implemented additional fields for observations

## Issue I worked on

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/AbstractReferenceHandlingTranslator.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/AbstractReferenceHandlingTranslator.java
@@ -21,6 +21,7 @@ import org.openmrs.Drug;
 import org.openmrs.Encounter;
 import org.openmrs.Location;
 import org.openmrs.Obs;
+import org.openmrs.Order;
 import org.openmrs.Patient;
 import org.openmrs.PatientIdentifier;
 import org.openmrs.Person;
@@ -31,6 +32,10 @@ import org.openmrs.module.fhir2.api.translators.PatientIdentifierTranslator;
 
 @Setter(AccessLevel.PACKAGE)
 public abstract class AbstractReferenceHandlingTranslator {
+	
+	public static final String DRUG_ORDER_TYPE_UUID = "131168f4-15f5-102d-96e4-000c29c2a5d7";
+	
+	public static final String TEST_ORDER_TYPE_UUID = "52a447d3-a64a-11e3-9aeb-50e549534c5e";
 	
 	@Inject
 	private PatientIdentifierTranslator patientIdentifierTranslator;
@@ -113,6 +118,22 @@ public abstract class AbstractReferenceHandlingTranslator {
 		}
 		
 		return reference;
+	}
+	
+	protected Reference createOrderReference(@NotNull Order order) {
+		if (order.getOrderType() == null) {
+			return null;
+		}
+		
+		if (order.getOrderType().getUuid().equals(TEST_ORDER_TYPE_UUID)) {
+			return new Reference().setReference(FhirConstants.SERVICE_REQUEST + "/" + order.getUuid())
+			        .setType(FhirConstants.SERVICE_REQUEST);
+		} else if (order.getOrderType().getUuid().equals(DRUG_ORDER_TYPE_UUID)) {
+			return new Reference().setReference(FhirConstants.MEDICATION + "/" + order.getUuid())
+			        .setType(FhirConstants.MEDICATION);
+		} else {
+			throw new IllegalArgumentException("Cannot create reference of undetermined order type");
+		}
 	}
 	
 	protected String getReferenceType(Reference reference) {
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationBasedOnReferenceTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationBasedOnReferenceTranslatorImpl.java
@@ -1,0 +1,66 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.translators.impl;
+
+import javax.inject.Inject;
+
+import lombok.AccessLevel;
+import lombok.Setter;
+import org.hl7.fhir.r4.model.Reference;
+import org.openmrs.Order;
+import org.openmrs.TestOrder;
+import org.openmrs.module.fhir2.FhirConstants;
+import org.openmrs.module.fhir2.api.dao.FhirMedicationRequestDao;
+import org.openmrs.module.fhir2.api.dao.FhirServiceRequestDao;
+import org.openmrs.module.fhir2.api.translators.ObservationBasedOnReferenceTranslator;
+import org.springframework.stereotype.Component;
+
+@Component
+@Setter(AccessLevel.PACKAGE)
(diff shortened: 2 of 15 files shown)


api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/AbstractReferenceHandlingTranslator.java
```diff
@@ -109,6 +111,13 @@ protected Reference createPractitionerReference(@NotNull Provider provider) {
 		return reference;
 	}
 	
+	protected Reference createOrderReference(@NotNull Order order) {
+		if (order instanceof DrugOrder) {
+			return null;
```
ibacher: We actually shouldn't return null if it's a `DrugOrder`. I'd just handle it as usual anyways.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationBasedOnReferenceTranslatorImpl.java
```diff
@@ -0,0 +1,56 @@
...
+import org.openmrs.module.fhir2.FhirConstants;
+import org.openmrs.module.fhir2.api.translators.ObservationBasedOnReferenceTranslator;
+import org.springframework.stereotype.Component;
+
+@Component
+@Setter(AccessLevel.PACKAGE)
+public class ObservationBasedOnReferenceTranslatorImpl extends AbstractReferenceHandlingTranslator implements ObservationBasedOnReferenceTranslator {
+	
+	@Inject
+	private OrderService orderService;
```
ibacher: Please don't use the `org.openmrs.api.OrderService` here. Instead, create a `FhirOrderDao` to get the order from the database. I don't want to depend on the OpenMRS Service API if we can avoid it.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationTranslatorImpl.java
```diff
@@ -137,6 +145,11 @@ public Obs toOpenmrsType(Obs existingObs, Observation observation, Supplier<Obs>
 			interpretationTranslator.toOpenmrsType(existingObs, observation.getInterpretation().get(0));
 		}
 		existingObs.setDateChanged(observation.getMeta().getLastUpdated());
+		existingObs.setDateCreated(observation.getIssued());
```
ibacher: We should not being doing this. Let the `dateCreated` (and `dateChanged`) fields get updated by the system, not by our code.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationTranslatorImpl.java
```diff
@@ -137,6 +145,11 @@ public Obs toOpenmrsType(Obs existingObs, Observation observation, Supplier<Obs>
...
+		existingObs.setObsDatetime(observation.getEffectiveDateTimeType().getValue());
```
ibacher: What happens if there is no `effectiveDateTimeType` set for this resource? We should actually have a translator here that can handle (at least) `effectiveDateTime` and `effectiveInstant`.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationTranslatorImpl.java
```diff
@@ -137,6 +145,11 @@ public Obs toOpenmrsType(Obs existingObs, Observation observation, Supplier<Obs>
...
+		if (observation.hasBasedOn()) {
```
ibacher: Add a new line here

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationTranslatorImpl.java
jecihjoy: @ibacher I'm not sure I understand this, do you mean like we shouldn't be mapping openmrs fields?

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationTranslatorImpl.java
jecihjoy: @ibacher do you mean doing this in something like a **ObservationEffectiveDatetimeTranslator**?

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationTranslatorImpl.java
ibacher: I just mean we shouldn't be updating `dateCreated` and `dateUpdated` based on values supplied via the FHIR interface. They should basically be read-only fields.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationTranslatorImpl.java
ibacher: Yes. It's kind of similar to what we had to do with `ObservationValueTranslator`.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationTranslatorImpl.java
jecihjoy: got it

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/AbstractReferenceHandlingTranslator.java
```diff
@@ -109,6 +110,10 @@ protected Reference createPractitionerReference(@NotNull Provider provider) {
 		return reference;
 	}
 	
+	protected Reference createOrderReference(@NotNull Order order) {
+		return new Reference().setReference(FhirConstants.ORDER + "/" + order.getUuid()).setType(FhirConstants.ORDER);
```
ibacher: This should still be "ServiceRequest" since "Order" isn't a valid FHIR object type.

ibacher: In this method, it may make sense to differentiate `DrugOrder`s, but in that case have them return a reference to a "MedicationRequest". However, that only makes sense once the `MedicationRequest` resource is created.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationBasedOnReferenceTranslatorImpl.java
```diff
@@ -0,0 +1,56 @@
...
+		return createOrderReference(order);
+	}
+	
+	@Override
+	public Order toOpenmrsType(Reference reference) {
+		if (reference == null) {
+			return null;
+		}
+		
+		if (!reference.getType().equals(FhirConstants.ORDER)) {
```
ibacher: Again, reference should be to a ServiceRequest, not an Order

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationEffectiveDatetimeTranslatorImpl.java
```diff
@@ -0,0 +1,49 @@
...
+		return null;
+	}
+	
+	@Override
+	public Obs toOpenmrsType(Obs obs, Type resource) {
+		if (resource == null) {
+			return obs;
+		}
+		
+		if (resource instanceof DateTimeType) {
```
ibacher: In addition to `DateTimeType` we should also support `InstantType` here, which is very similar to `DateTimeType`, but represents a slightly more precise type.

jecihjoy:
@ibacher I have updated the code. Also I have made the based On reference specific to ServiceRequest but once the MedicationRequest is created, we should be able to update this to something like

```
protected Reference createOrderReference(@NotNull Order order) {
		if (order instanceof TestOrder) {
			return new Reference().setReference(FhirConstants.SERVICEREQUEST + "/" + order.getUuid())
			        .setType(FhirConstants.SERVICEREQUEST);
		} else if (order instanceof DrugOrder) {
			return new Reference().setReference(FhirConstants.MEDICATIONREQUEST + "/" + order.getUuid())
			        .setType(FhirConstants.MEDICATIONREQUEST);
		}
		throw new IllegalArgumentException("Cannot create reference of undetermined order type");
	}

PR #119, opened 2020-03-21
Implementation of several parameters for Condition search (api-2.2)

bashir2:
FM2-87 Add search for Condition resource

This is an implementation of several (not all) parameters for Condition search and only for api-2.2.

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-87

  No? -> execute above command


api-2.2/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirConditionDaoImpl_2_2Test.java
```diff
@@ -106,6 +267,7 @@ public void shouldSaveNewCondition() {
 		assertThat(result.getUuid(), equalTo(CONDITION_UUID));
 	}
 	
+	/* TODO comment out and fix before submit, once the conflicts in the initial_data.xml file are resolved.
```
bashir2: I will update this unit-test later (before merge); the problem is that the main input data file (`FhirConditionDaoImplTest_initial_data.xml`) has been changed while I was working on this PR and the change is fairly significant. I also needed to make changes in that file, so I need to figure out how the new save feature works and why this test fails with my version of the input data file.

bashir2: This is resolved now.

api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImplTest.java
```diff
@@ -247,7 +247,7 @@ public void searchForObs_shouldReturnObsByPatientGivenName() {
 		patient.setChain(Patient.SP_GIVEN);
 		
 		patientReference.addValue(new ReferenceOrListParam().add(patient));
-
```
ibacher: This shouldn't be part of this commit

omod/src/main/java/org/openmrs/module/fhir2/providers/ConditionFhirResourceProvider.java
```diff
@@ -70,4 +80,21 @@ public Condition getConditionByUuid(@IdParam @NotNull IdType id) {
 	public MethodOutcome createCondition(@ResourceParam Condition newCondition) {
 		return FhirServerUtils.buildCreate(conditionService.saveCondition(newCondition));
 	}
+	
+	@Search
+	//@SuppressWarnings("unused")
```
ibacher: Just delete this instead of commenting it out.

omod/src/main/java/org/openmrs/module/fhir2/providers/ConditionFhirResourceProvider.java
```diff
@@ -70,4 +80,21 @@ public Condition getConditionByUuid(@IdParam @NotNull IdType id) {
...
+	public Bundle searchConditions(
+	        @OptionalParam(name = Condition.SP_PATIENT, chainWhitelist = { "", Patient.SP_IDENTIFIER, Patient.SP_NAME,
+	                Patient.SP_GIVEN, Patient.SP_FAMILY }) ReferenceAndListParam patientParam,
+	        @OptionalParam(name = Condition.SP_SUBJECT, chainWhitelist = { "", Patient.SP_IDENTIFIER, Patient.SP_NAME,
+	                Patient.SP_GIVEN, Patient.SP_FAMILY }) ReferenceAndListParam subjectParam,
+	        @OptionalParam(name = Condition.SP_CODE) TokenOrListParam code,
+	        @OptionalParam(name = Condition.SP_CLINICAL_STATUS) TokenOrListParam clinicalStatus,
+	        @OptionalParam(name = Condition.SP_ONSET_DATE) DateParam onsetDate,
```
ibacher: I would use a `DateRangeParam` here. `DateParam` only expresses a single date. `DateRangeParam` can express a single date or (more obviously) a point that is between two dates / instants.

api-2.2/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirConditionDaoImpl_2_2Test.java
```diff
@@ -90,14 +116,166 @@ public void shouldReturnNullWhenGetConditionByWrongUuid() {
...
+	public void searchForPatients_shouldReturnConditionByPatientGivenName() {
+		ReferenceParam patientReference = new ReferenceParam(Patient.SP_GIVEN, PATIENT_GIVEN_NAME);
+		ReferenceAndListParam patientList = new ReferenceAndListParam();
+		patientList.addValue(new ReferenceOrListParam().add(patientReference));
+		Collection<Condition> results = dao.searchForConditions(patientList, null, null, null, null, null, null, null);
+		
+		assertThat(results, notNullValue());
+		assertThat(results, not(empty()));
+		// TODO during review: Is returning duplicated Conditions the desired outcome?
+		assertThat(results.size(), equalTo(6));
```
ibacher: > Is returning duplicated Conditions the desired outcome?

Short version: no, we really shouldn't be getting duplicated conditions. Can you say more about why this is happening?

api-2.2/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirConditionDaoImpl_2_2Test.java
```diff
@@ -90,14 +116,166 @@ public void shouldReturnNullWhenGetConditionByWrongUuid() {
...
+		
+		assertThat(results, notNullValue());
+		assertThat(results, not(empty()));
+		assertThat(results.size(), equalTo(6));
+	}
+	
+	@Test
+	public void searchForPatients_shouldReturnConditionByCode() {
+		TokenOrListParam listParam = new TokenOrListParam();
+		listParam.add(new TokenParam("CD41003")); // for concept_id=5497
```
ibacher: Where does this CD41003 value come from?

api-2.2/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConditionDaoImpl_2_2.java
```diff
@@ -42,6 +52,47 @@ public Condition getConditionByUuid(String uuid) {
...
+	private ConditionClinicalStatus convertStatus(String status) {
+		if ("active".equalsIgnoreCase(status)) {
+			return ConditionClinicalStatus.ACTIVE;
+		}
+		if ("inactive".equalsIgnoreCase(status)) {
+			return ConditionClinicalStatus.INACTIVE;
+		}
+		// TODO during review: What value of the spec. should be mapped to HISTORY_OF?
+		// http://www.hl7.org/fhir/valueset-condition-clinical.html
+		return ConditionClinicalStatus.HISTORY_OF;
```
ibacher: We shouldn't be, since `HISTORY_OF` is not semantically equivalent to any of those values. Normally, I'd say we just add a custom valueset, but that obviously makes us non-conformant. I think we need to handle `HISTORY_OF` via an extension. Ugh...

api-2.2/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConditionDaoImpl_2_2.java
```diff
@@ -42,6 +52,47 @@ public Condition getConditionByUuid(String uuid) {
...
+	@Override
+	public Collection<Condition> searchForConditions(ReferenceAndListParam patientParam, ReferenceAndListParam subjectParam,
+	        TokenOrListParam code, TokenOrListParam clinicalStatus, DateParam onsetDate, QuantityParam onsetAge,
+	        DateParam recordedData, SortSpec sort) {
+		Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Condition.class);
+		
+		handlePatientReference(criteria, patientParam);
+		// TODO during review: Do we require a ":Patient" modifier or a "Patient/" type for the "subject" search param?
+		// And do we want to support anything other than "Patient" under "subject" based searches (e.g., "Group")?
+		handlePatientReference(criteria, subjectParam);
```
ibacher: The `patient` search parameter is defined as `Condition.subject.where(resolve() is Patient)`, i.e., if the `patientParam` is valid, we should not handle the `subjectParam` at all.

> Do we require a ":Patient" modifier or a "Patient/" type for the "subject" search param?

No, but we should support these where present

> Do we want to support anything other than "Patient" under "subject" based searches (e.g., "Group")?

We currently haven't implemented the concept of "Group" in the OpenMRS FHIR module, so we needn't support it here.

api-2.2/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConditionDaoImpl_2_2.java
```diff
@@ -42,6 +52,47 @@ public Condition getConditionByUuid(String uuid) {
...
+		handleDate("onsetDate", onsetDate).ifPresent(criteria::add);
+		// TODO: Handle onsetAge as well.
+		handleDate("dateCreated", recordedData).ifPresent(criteria::add);
+		handleOrListParam(clinicalStatus,
+		    tokenParam -> Optional.of(eq("clinicalStatus", convertStatus(tokenParam.getValue())))).ifPresent(criteria::add);
+		if (code != null) {
```
ibacher: Use `handleResourceCode()` for this. We should give that function a more descriptive name... and some docs!

omod/src/main/java/org/openmrs/module/fhir2/providers/ConditionFhirResourceProvider.java
ibacher: I think that's an artifact of the fact that we've been writing code backwards, i.e., tests after the fact, so that the methods look like they're unused. Unfortunately, I'm as guilty of that as the next person.

api-2.2/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirConditionDaoImpl_2_2Test.java
ibacher: Thanks! I haven't looks through the standard dataset in a while, so I forgot about that.

api-2.2/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirConditionDaoImpl_2_2Test.java
ibacher: Yep, that's a good point and probably out of scope for this ticket. I'll create a new ticket to resolve that issue. Small note on coding style, I prefer using the Hamcrest `hasSize()` method instead of `size(), equalTo()` as it results in a more meaningful error message when it fails.

api-2.2/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConditionDaoImpl_2_2.java
ibacher: Thanks! Yeah... `generateSystemQuery()` was written with some different assumptions. Thanks for fixing that.

api-2.2/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConditionDaoImpl_2_2.java
ibacher: Yeah, that's fine.

I don't know whether it's on its way to being deprecated or not, but I hope so. Maybe I can propose that core adopts the second level values from the HL7 value set, since "remission" and "resolved" seem more semantically meaningful than just "history of".

bashir2:
Thanks @ibacher for the review/approval. I have got some non-trivial merge conflicts from `master` which I am resolving now and will update this PR shortly.

bashir2:
@ibacher this is good to go; I resolved merge conflicts and had to clean-up DB to do some manual tests too which were all fine.

PR #150, opened 2020-04-27
Add naming convention documentation and change classes to conform

ibacher:
This is a proposal for naming conventions for this module.

@corneliouzbett @jecihjoy @pmanko Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,19 @@ Concept result = getConcept("expectedConcept");
 assertThat(result, equalTo(expected));
 ```
 
+Class Naming Conventions
+------------------------
+
+* Prefer to create an interface rather than directly creating a concrete type. This allows modules and implementers to swap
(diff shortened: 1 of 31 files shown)


corneliouzbett:
Personally, I am bad at naming classes & methods. Working on improving that. Nevertheless, I think these are good naming conventions and we should adopt them. 

I have some few suggestions:
  * Introduction of a package `base` in the package where you have a base class to make it easy when locating these classes.
 * When naming booleans, we should prefix with `Can`, `Has` or `Is`.

corneliouzbett:
I figure out this has a lot of impact on the PR being raised. Anyway, I am merging into master.

ibacher:
Yeah... It's an issue I only became aware because of the stuff being worked on. Anyways, it more fun to just break everyone's pending changes 😬.

PR #152, opened 2020-04-28
FM2-152: Update Encounter DAO and Service

gitcliff:
link;https://issues.openmrs.org/browse/FM2-152

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
@@ -9,41 +9,28 @@
  */
 package org.openmrs.module.fhir2.api.dao.impl;
 
-import static org.hibernate.criterion.Restrictions.eq;
-
 import java.util.Collection;
+
+import org.hibernate.Criteria;
+import org.openmrs.Encounter;
+import org.openmrs.module.fhir2.api.dao.FhirEncounterDao;
+import org.springframework.stereotype.Component;
 
 import ca.uhn.fhir.rest.param.DateRangeParam;
 import ca.uhn.fhir.rest.param.ReferenceAndListParam;
 import lombok.AccessLevel;
 import lombok.Setter;
-import org.hibernate.Criteria;
-import org.hibernate.SessionFactory;
-import org.openmrs.Encounter;
-import org.openmrs.module.fhir2.api.dao.FhirEncounterDao;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Qualifier;
-import org.springframework.stereotype.Component;
 
 @Component
 @Setter(AccessLevel.PACKAGE)
-public class FhirEncounterDaoImpl extends BaseDao implements FhirEncounterDao {
+public class FhirEncounterDaoImpl extends BaseFhirDao<Encounter> implements FhirEncounterDao {
 	
-	@Autowired
-	@Qualifier("sessionFactory")
-	private SessionFactory sessionFactory;
-	
-	@Override
-	public Encounter getEncounterByUuid(String uuid) {
-		return (Encounter) sessionFactory.getCurrentSession().createCriteria(Encounter.class).add(eq("uuid", uuid))
-		        .uniqueResult();
-	}
 	
 	@Override
 	public Collection<Encounter> searchForEncounters(DateRangeParam date, ReferenceAndListParam location,
 	        ReferenceAndListParam participant, ReferenceAndListParam subject) {
 		
-		Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Encounter.class);
+		Criteria criteria = getSessionFactory().getCurrentSession().createCriteria(Encounter.class);
 		
 		handleDateRange("encounterDatetime", date).ifPresent(criteria::add);
 		handleLocationReference("l", location).ifPresent(l -> criteria.createAlias("location", "l").add(l));
--- a/api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
@@ -15,6 +15,7 @@ import java.util.stream.Collectors;
 import ca.uhn.fhir.rest.param.DateRangeParam;
 import ca.uhn.fhir.rest.param.ReferenceAndListParam;
 import lombok.AccessLevel;
+import lombok.Getter;
 import lombok.Setter;
 import org.hl7.fhir.r4.model.Encounter;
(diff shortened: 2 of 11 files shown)


api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
```diff
@@ -9,41 +9,31 @@
...
-	private SessionFactory sessionFactory;
+public class FhirEncounterDaoImpl extends BaseFhirDaoImpl<Encounter> implements FhirEncounterDao {
 	
 	@Override
-	public Encounter getEncounterByUuid(String uuid) {
-		return (Encounter) sessionFactory.getCurrentSession().createCriteria(Encounter.class).add(eq("uuid", uuid))
-		        .uniqueResult();
+	public Encounter get(String uuid) {
+		return super.get(uuid);
 	}
```
ibacher, line 30: You shouldn't need to implement this method at all.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
```diff
@@ -37,8 +39,8 @@
 	
 	@Override
 	@Transactional(readOnly = true)
-	public Encounter getEncounterByUuid(String uuid) {
-		return translator.toFhirResource(dao.getEncounterByUuid(uuid));
+	public Encounter get(String uuid) {
+		return translator.toFhirResource(dao.get(uuid));
 	}
```
ibacher, line 44: This shouldn't be necessary to have here at all.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
```diff
@@ -48,4 +50,14 @@ public Encounter getEncounterByUuid(String uuid) {
...
+	
+	@Override
+	protected FhirDao<org.openmrs.Encounter> getDao() {
+		return dao;
+	}
+	
+	@Override
+	protected OpenmrsFhirTranslator<org.openmrs.Encounter, Encounter> getTranslator() {
+		return translator;
+	}
```
ibacher, line 62: Rather than explicitly generating these fields, I would add the annotation `@Getter(AccessLevel.PROTECTED)` right above the annotation for `@Setter` at the top of the class.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
gitcliff: @ibacher  thanks for the review ,,
i getting abit confused here why the method should be implemented since its coming from the super class being extended and its only refactoring an already existing method with a change in the method name ..
Just requesting for some more explanation here to help me better understand ..Thanks

ibacher: It doesn't need to be here at all. I think it's left in the `FhirEncounterDao` interface so we can include Javadoc documenting its use, but in the actual implementation, we can rely on the super class version.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
gitcliff: yes sure ,,,thanks

gitcliff: have added it

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
gitcliff: thanks ,,have fixed it ..

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
gitcliff: thanks ,,have removed it

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
```diff
@@ -24,22 +20,26 @@
...
 	@Autowired
 	EncounterTranslator translator;
 	
-	@Override
-	@Transactional(readOnly = true)
-	public Encounter getEncounterByUuid(String uuid) {
-		return translator.toFhirResource(dao.getEncounterByUuid(uuid));
-	}
+
+	//The get(String uuid) method which fetches the encounter by a uuid is handled in the FhirEncounterService interface
```
ibacher, line 42: Can we get rid of this comment?

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
```diff
@@ -9,41 +9,28 @@
...
-	@Qualifier("sessionFactory")
-	private SessionFactory sessionFactory;
+public class FhirEncounterDaoImpl extends BaseFhirDaoImpl<Encounter> implements FhirEncounterDao {
 	
-	@Override
-	public Encounter getEncounterByUuid(String uuid) {
-		return (Encounter) sessionFactory.getCurrentSession().createCriteria(Encounter.class).add(eq("uuid", uuid))
-		        .uniqueResult();
-	}
+	//The get(String uuid) method which fetches the encounter by a uuid is handled in the FhirEncounterDao interface
```
ibacher, line 27: Can we get rid of this comment?

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
gitcliff: i have removed them ..thanks

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
gitcliff: i have removed them,,thanks

HerbertYiga:
hi @gitcliff ,we do have some conflicting files here that need to be resolved

ibacher:
@gitcliff Can you rebase this on the latest master version? In particular `FhirBaseDaoImpl` was just renamed to `FhirBaseDao`. This should then be all set.

gitcliff:
@ibacher i have made the change above to `FhirBaseDao`

ibacher:
@gitcliff Thanks for this!

PR #155, opened 2020-04-28
Add custom strategy for determining URL

ibacher:
@corneliouzbett Here's an implementation of an `IServerAddressStrategy` that solves (I think) the problem. It also adds a global property so you can tell it what it should use.


corneliouzbett:
> @corneliouzbett Here's an implementation of an `IServerAddressStrategy` that solves (I think) the problem. It also adds a global property so you can tell it what it should use.

There's one way to find out 😀.

PR #158, opened 2020-04-30
FM2-183: Add support for DSTU3

ibacher:
@VaishSiddharth I'm opening this as a draft PR as the communication tools are better integrated than what we've been doing.

Anyone else, we're currently building support for DSTU3 into the module. This will also serve as the pattern for integrating new versions of FHIR into the module, though we may not really be looking to support versions *earlier* that DSTU3.


omod-dstu3/src/main/java/org/openmrs/module/fhir2/providers/AllergyIntoleranceDstu3FhirResourceProvider.java
```diff
@@ -0,0 +1,81 @@
...
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.providers;
+
+import javax.validation.constraints.NotNull;
+
+import java.util.List;
+
+import ca.uhn.fhir.rest.annotation.*;
```
ibacher, line 16: Minor point, but please configure your IDE not to introduce `*` imports. OpenMRS doesn't like anyone to use them, and it makes some sense.

omod-dstu3/src/main/java/org/openmrs/module/fhir2/providers/AllergyIntoleranceDstu3FhirResourceProvider.java
```diff
@@ -0,0 +1,81 @@
...
+import ca.uhn.fhir.rest.param.ReferenceAndListParam;
+import ca.uhn.fhir.rest.param.TokenAndListParam;
+import ca.uhn.fhir.rest.param.TokenOrListParam;
+import ca.uhn.fhir.rest.server.IResourceProvider;
+import ca.uhn.fhir.rest.server.exceptions.ResourceNotFoundException;
+import lombok.AccessLevel;
+import lombok.Setter;
+import org.hl7.fhir.convertors.VersionConvertor_30_40;
+import org.hl7.fhir.dstu3.model.*;
```
ibacher, line 25: Remove the `*` import

omod-dstu3/src/main/java/org/openmrs/module/fhir2/providers/ConditionDstu3FhirResourceProvider.java
```diff
@@ -0,0 +1,93 @@
...
+	@SuppressWarnings("unused")
+	public List<Resource> getConditionHistoryById(@IdParam @NotNull IdType id) {
+		org.hl7.fhir.r4.model.Condition condition = conditionService.getConditionByUuid(id.getIdPart());
+		if (condition == null) {
+			throw new ResourceNotFoundException("Could not find condition with Id " + id.getIdPart());
+		}
+		return VersionConvertor_30_40.convertCondition(condition).getContained();
+	}
+	
+	//	// TODO: Required or not?
```
ibacher, line 69: Yes! We should implement this!

omod-dstu3/src/main/java/org/openmrs/module/fhir2/providers/DiagnosticReportDstu3FhirResourceProvider.java
```diff
@@ -0,0 +1,73 @@
...
+		if (diagnosticReport == null) {
+			throw new ResourceNotFoundException("Could not find diagnosticReport with Id " + id.getIdPart());
+		}
+		
+		return VersionConvertor_30_40.convertDiagnosticReport(diagnosticReport);
+	}
+	
+	// TODO: Required or not?
+	@Create
+	public MethodOutcome createDiagnosticReport(@ResourceParam org.hl7.fhir.r4.model.DiagnosticReport diagnosticReport) {
```
ibacher, line 57: Required, but we will obviously be given the R3 version of the DiagnosticReport here.

omod-dstu3/src/main/java/org/openmrs/module/fhir2/providers/DiagnosticReportDstu3FhirResourceProvider.java
```diff
@@ -0,0 +1,73 @@
...
+		return FhirServerUtils.buildCreate(diagnosticReportService.saveDiagnosticReport(diagnosticReport));
+	}
+	
+	// TODO: Required or not?
```
ibacher, line 61: Required, but we will obviously be given the R3 version of the DiagnosticReport here.

omod-dstu3/src/main/java/org/openmrs/module/fhir2/providers/EncounterDstu3FhirResourceProvider.java
```diff
@@ -0,0 +1,82 @@
...
+	                Patient.SP_FAMILY,
+	                Patient.SP_NAME }, targetTypes = Patient.class) ReferenceAndListParam subjectReference) {
+		return VersionConvertor_30_40.convertBundle(FhirServerUtils.convertSearchResultsToBundle(
+		    encounterService.searchForEncounters(date, location, participantReference, subjectReference)));
+		
+	}
+	
+	@History
+	@SuppressWarnings("unused")
+	public List<Resource> getEncounterHistoryById(@IdParam @NotNull IdType id) {
```
ibacher, line 74: I'd prefer if these files were structured similarly, i.e. with the `@History` method following the `@Read` method.

omod-dstu3/src/main/java/org/openmrs/module/fhir2/providers/AllergyIntoleranceDstu3FhirResourceProvider.java
```diff
@@ -0,0 +1,81 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.providers;
```
ibacher, line 10: Thinking about this some more, I think that instead of naming every `ResourceProvider` `...DStu3ResourceProvider` we should move them into a `r3` package so, for instance, this class would become `org.openmrs.module.fhir2.providers.r3.AllergyIntoleranceFhirResourceProvider`.

omod-dstu3/src/main/java/org/openmrs/module/fhir2/providers/AllergyIntoleranceDstu3FhirResourceProvider.java
```diff
@@ -0,0 +1,81 @@
...
+import org.hl7.fhir.dstu3.model.*;
+import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.openmrs.module.fhir2.api.FhirAllergyIntoleranceService;
+import org.openmrs.module.fhir2.util.FhirServerUtils;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.stereotype.Component;
+
+@Component
+@Qualifier("fhirDstu3Resources")
```
ibacher, line 34: Let's shorten this to `fhirR3Resources` (R3 is the release, but was called STU3 — standard trial use; R2 was released and called DSTU2 — draft standard trial use; I'm just favouring the shorter one!)

omod-dstu3/src/main/java/org/openmrs/module/fhir2/providers/ObservationDstu3FhirResourceProvider.java
```diff
@@ -0,0 +1,84 @@
...
+	        @OptionalParam(name = Observation.SP_VALUE_STRING) StringAndListParam valueStringParam,
+	        @OptionalParam(name = Observation.SP_DATE) DateRangeParam date,
+	        @OptionalParam(name = Observation.SP_CODE) TokenAndListParam code, @Sort SortSpec sort) {
+		return VersionConvertor_30_40.convertBundle(FhirServerUtils.convertSearchResultsToBundle(
+		    observationService.searchForObservations(encounterReference, patientReference, hasMemberReference, valueConcept,
+		        valueDateParam, valueQuantityParam, valueStringParam, date, code, sort)));
+	}
+	
+	@History
+	public List<Resource> getObservationHistoryById(@IdParam @NotNull IdType id) {
```
ibacher, line 77: See above about the order of methods

omod-dstu3/src/main/java/org/openmrs/module/fhir2/providers/PatientDstu3FhirResourceProvider.java
```diff
@@ -0,0 +1,87 @@
...
+	        @OptionalParam(name = Patient.SP_ADDRESS_POSTALCODE) StringOrListParam postalCode,
+	        @OptionalParam(name = Patient.SP_ADDRESS_COUNTRY) StringOrListParam country, @Sort SortSpec sort) {
+		return VersionConvertor_30_40
+		        .convertBundle(FhirServerUtils.convertSearchResultsToBundle(patientService.searchForPatients(name, given,
+		            family, identifier, gender, birthDate, deathDate, deceased, city, state, postalCode, country, sort)));
+	}
+	
+	@History
+	@SuppressWarnings("unused")
+	public List<Resource> getPatientResourceHistory(@IdParam @NotNull IdType id) {
```
ibacher, line 79: `getPatientHistoryById`, also ordering.

omod-dstu3/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirDstu3RestServlet.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Collection;
+
+import ca.uhn.fhir.context.FhirContext;
+import ca.uhn.fhir.rest.server.IResourceProvider;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.stereotype.Component;
+
+@Component
+public class FhirDstu3RestServlet extends FhirRestServlet {
```
ibacher, line 21: To be consistent, let's call this FhirR3RestServlet

omod/src/main/java/org/openmrs/module/fhir2/web/filter/ForwardingFilter.java
```diff
@@ -33,7 +38,29 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
...
-			String newURI = requestURI.replace(prefix, "/ms/fhir2Servlet/");
+			String[] requestPath = requestURI.split(prefix, 1);
+			String maybeVersion = requestPath[1].split("/", 1)[0];
+			
+			String replacement;
+			switch (maybeVersion) {
+				case "R2":
+					prefix += "/" + maybeVersion + "/";
+					replacement = "/ms/fhir2Dstu2Servlet";
+					break;
```
ibacher, line 49: Let's remove this from this switch since we aren't going to use it right now.

omod/src/main/resources/config.xml
```diff
@@ -50,6 +50,16 @@
...
 
+	<servlet>
+		<servlet-name>fhir2Dstu3Servlet</servlet-name>
+		<servlet-class>org.openmrs.module.fhir2.web.servlet.FhirDstu3RestServlet</servlet-class>
+	</servlet>
+
+	<servlet>
+		<servlet-name>fhir2Dstu2Servlet</servlet-name>
+		<servlet-class>org.openmrs.module.fhir2.web.servlet.FhirDstu2RestServlet</servlet-class>
+	</servlet>
```
ibacher, line 61: We should remove this servlet mapping

omod/src/main/resources/config.xml
```diff
@@ -50,6 +50,16 @@
 		<servlet-class>org.openmrs.module.fhir2.web.servlet.FhirRestServlet</servlet-class>
 	</servlet>
 
+	<servlet>
+		<servlet-name>fhir2Dstu3Servlet</servlet-name>
```
ibacher, line 54: We should change this `servlet-name` to fhir2R3Servlet

omod/src/main/java/org/openmrs/module/fhir2/web/filter/ForwardingFilter.java
```diff
@@ -33,7 +38,29 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
...
+					prefix += "/" + maybeVersion + "/";
+					replacement = "/ms/fhir2Dstu3Servlet";
+					break;
+				case "R4":
+					prefix += "/" + maybeVersion + "/";
+					replacement = "/ms/fhir2Servlet";
+					break;
+				default:
+					replacement = "/ms/fhir2Servlet";
+					break;
```
ibacher, line 60: We need to rethink this. We should return an error if the value isn't a known version so that we're using a consistent access URL.

omod-dstu3/src/test/java/org/openmrs/module/fhir2/web/servlet/BaseFhirResourceProviderTest.java
```diff
@@ -0,0 +1,255 @@
...
+import org.openmrs.module.fhir2.api.impl.FhirGlobalPropertyServiceImpl;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import org.springframework.mock.web.MockHttpServletRequest;
+import org.springframework.mock.web.MockHttpServletResponse;
+import org.springframework.mock.web.MockServletConfig;
+import org.springframework.mock.web.MockServletContext;
+import org.springframework.transaction.annotation.Transactional;
+
+public abstract class BaseFhirResourceProviderTest<T extends IResourceProvider, U extends IBaseResource> {
```
ibacher, line 51: Instead of making a fork of this class, maybe it makes more sense to call the version here `BaseFhirR3ResourceProviderTest` and have it extend `BaseFhirResourceProviderTest` and then make the changes that we need here. This might involve changing `BaseFhirResourceProviderTest`, but that's ok.

omod-dstu3/src/test/java/org/openmrs/module/fhir2/web/servlet/BaseFhirProvenanceResourceTest.java
```diff
@@ -0,0 +1,112 @@
...
+import org.hl7.fhir.r4.model.Coding;
+import org.hl7.fhir.r4.model.Practitioner;
+import org.hl7.fhir.r4.model.Provenance;
+import org.hl7.fhir.r4.model.Reference;
+import org.mockito.Mock;
+import org.openmrs.User;
+import org.openmrs.module.fhir2.FhirConstants;
+import org.openmrs.module.fhir2.api.translators.PractitionerReferenceTranslator;
+
+public abstract class BaseFhirProvenanceResourceTest<T extends DomainResource> {
```
ibacher, line 29: Instead of making a fork of this class, maybe it makes more sense to call the version here `BaseFhirR3ProvenanceResourceTest` and have it extend `BaseFhirProvenanceResourceTest` and then make the changes that we need here. This might involve changing `BaseFhirProvenanceResourceTest`, but that's ok.

omod-dstu3/src/main/java/org/openmrs/module/fhir2/providers/EncounterDstu3FhirResourceProvider.java
VaishSiddharth: @ibacher This might need to be changed in master as well. I am not changing anything in omod  module because it might be irrelevant for this PR

omod-dstu3/src/main/java/org/openmrs/module/fhir2/providers/PatientDstu3FhirResourceProvider.java
VaishSiddharth: @ibacher This might need to be changed in master as well

ibacher: I'm aware. I've got to go through that code soon.

ibacher:
@VaishSiddharth One other thing this would benefit from is rebasing on `master`. We had a change go in to build support for paging, but it requires some pretty sizeable changes.

ibacher:
That's also, incidentally, why the Travis build is failing.

omod-dstu3/src/main/java/org/openmrs/module/fhir2/providers/r3/EncounterFhirResourceProvider.java
```diff
@@ -53,6 +53,16 @@ public Encounter getEncounterById(@IdParam @NotNull IdType id) {
...
 	
+	@History
+	@SuppressWarnings("unused")
+	public List<Resource> getEncounterHistoryById(@IdParam @NotNull IdType id) {
+		org.hl7.fhir.r4.model.Encounter encounter = encounterService.getEncounterByUuid(id.getIdPart());
+		if (encounter == null) {
+			throw new ResourceNotFoundException("Could not find encounter with Id " + id.getIdPart());
+		}
+		return VersionConvertor_30_40.convertEncounter(encounter).getContained();
+	}
```
VaishSiddharth, line 64: @ibacher Test for all these functions fail. Can you give an idea, what's wrong here

ibacher: Likely because the converter doesn't handled the contained resources properly, so when `getContained()` gets called here, it returns an empty list. We may need to look at implementing re-implementing the code to get the provenance resources as part of the R3 submodule.

omod/src/test/java/org/openmrs/module/fhir2/providers/FhirResourceProviderTest.java
```diff
@@ -0,0 +1,10 @@
+package org.openmrs.module.fhir2.providers;
+
+import org.hl7.fhir.instance.model.api.IBaseBundle;
+import org.springframework.mock.web.MockHttpServletResponse;
+
+import java.io.UnsupportedEncodingException;
+
+public interface FhirResourceProviderTest {
+    IBaseBundle readBundleResponse(MockHttpServletResponse response) throws UnsupportedEncodingException;
```
VaishSiddharth, line 9: @ibacher I have added an interface like this.

ibacher: @VaishSiddharth Sorry... I was apparently leading you astray. [This commit](https://github.com/openmrs/openmrs-module-fhir2/pull/158/commits/88232ddf1a43567bd4b11e0e1cbe88894bbafeee) has the refactoring necessary for this. I left some errors in the DSTU3 test classes though. See if you can work them out.

VaishSiddharth:
@ibacher Not Really :D. I had to manipulate a condition test because 

SP_RECORDED_DATE was not present so changed to SP_ASSERTED_DATE in provider so corresponding test needed some tweaks

omod-dstu3/src/main/java/org/openmrs/module/fhir2/providers/r3/ConditionFhirResourceProvider.java
```diff
@@ -0,0 +1,94 @@
...
+	public Bundle searchConditions(
+	        @OptionalParam(name = Condition.SP_PATIENT, chainWhitelist = { "", Patient.SP_IDENTIFIER, Patient.SP_NAME,
+	                Patient.SP_GIVEN, Patient.SP_FAMILY }) ReferenceAndListParam patientParam,
+	        @OptionalParam(name = Condition.SP_SUBJECT, chainWhitelist = { "", Patient.SP_IDENTIFIER, Patient.SP_NAME,
+	                Patient.SP_GIVEN, Patient.SP_FAMILY }) ReferenceAndListParam subjectParam,
+	        @OptionalParam(name = Condition.SP_CODE) TokenAndListParam code,
+	        @OptionalParam(name = Condition.SP_CLINICAL_STATUS) TokenAndListParam clinicalStatus,
+	        @OptionalParam(name = Condition.SP_ONSET_DATE) DateRangeParam onsetDate,
+	        @OptionalParam(name = Condition.SP_ONSET_AGE) QuantityAndListParam onsetAge,
+	        //SP_RECORDED_DATE not present so changed to SP_ASSERTED_DATE
```
ibacher, line 88: This is correct.

omod-dstu3/src/main/java/org/openmrs/module/fhir2/providers/r3/ObservationFhirResourceProvider.java
```diff
@@ -0,0 +1,84 @@
...
+		return VersionConvertor_30_40.convertObservation(observation).getContained();
+	}
+	
+	@Search
+	public IBundleProvider searchObservations(
+	        @OptionalParam(name = Observation.SP_ENCOUNTER) ReferenceAndListParam encounterReference,
+	        @OptionalParam(name = Observation.SP_SUBJECT, chainWhitelist = { "", Patient.SP_IDENTIFIER, Patient.SP_GIVEN,
+	                Patient.SP_FAMILY,
+	                Patient.SP_NAME }, targetTypes = Patient.class) ReferenceAndListParam patientReference,
+	        //TODO: SP_HAS_MEMBER not present so converted to SP_BASED_ON just to prevent errors
```
ibacher, line 71: This should be `SP_RELATED`, but note that `SP_RELATED` has a somewhat different structure.

VaishSiddharth: @ibacher Shouldn't this be `SP_RELATED_TYPE` [here](http://hl7.org/fhir/STU3/observation.html#search) is why I think so.
Also only changing the data type will not solve the problems, I think that we'll have to map it, as we did for `Task` also the `observation` resource returns `IBundleProvider` in search function which complicates things or if not all the tests are failing with this error message
```
ca.uhn.fhir.context.ConfigurationException: This context is for FHIR version "DSTU3" but the class "org.hl7.fhir.r4.model.OperationOutcome" is for version "R4"

```
I guess we'll need to override this method also in `BaseFhirR3ResourceProviderTest`

ibacher: No. `SP_RELATED`. `SP_RELATED_TYPE` refers to the type structure that is part of the related element. Technically, `SP_RELATED_TARGET` is the more direct replacement for `SP_HAS_MEMBER`, but only when the `related` structure looks like this:

```json
"related": {
    "type": "has-member",
    "target": "Observation/3"
}
```

Or, in terms of search parameters:

```
/Observation?related.type=has-member&related.target=Observation/3
```

This also isn't exactly like Task, because what we're converting here isn't a full resource, but a search parameter.

The message you highlight:

```
ca.uhn.fhir.context.ConfigurationException: This context is for FHIR version "DSTU3" but the class "org.hl7.fhir.r4.model.OperationOutcome" is for version "R4"
```

Does indicate a problem, but part of the problem is that we're not properly handling things when the result is an `OperationOutcome` (which itself needs to be translated). Please step through one of these failing tests with a debugger to see what is actually happening.

> I guess we'll need to override this method also in BaseFhirR3ResourceProviderTest

Which method are you referring to?

VaishSiddharth: > Which method are you referring to?


protected void describeMismatchSafely(MockHttpServletResponse item, Description mismatchDescription) in this function we have `OperationOutcome` which is of `import org.hl7.fhir.r4.model.OperationOutcome;` type so I think this needs to be generic so that all versions and use it

VaishSiddharth:
@ibacher Thanks for your commit. I was trying to do something similar to solve the error :+1:

VaishSiddharth:
@ibacher I have tested the latest commit with Postman and both endpoints work fine. :+1:

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirVersionUtils.java
```diff
@@ -0,0 +1,57 @@
...
+		R4,
+		R3,
+		ERROR
+	}
+	
+	public static Enum<FhirVersion> getFhirResourceVersion(HttpServletRequest request) {
+		String requestURI = request.getRequestURI();
+		String contextPath = request.getContextPath();
+		String prefix = contextPath + "/ws/fhir2";
+		if (requestURI.startsWith(prefix)) {
```
VaishSiddharth, line 26: @ibacher I get a NPE here when I do something like this

```java
return request.getScheme() + "://" + request.getServerName() + port + "/" + contextPath + "/ws/fhir2/"
		        + FhirVersionUtils.getFhirResourceVersion(request);
```
in `OpenmrsFhirAddressStrategy`

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirVersionUtils.java
```diff
@@ -0,0 +1,57 @@
...
+package org.openmrs.module.fhir2.web.servlet;
+
+import javax.servlet.http.HttpServletRequest;
+
+public class FhirVersionUtils {
+	
+	public enum FhirVersion {
+		R4,
+		R3,
+		ERROR
```
ibacher, line 19: I would make this `UNKNOWN`. Also I would order the values in the enum this way: `UNKNOWN`, `R3`, `R4`. That way we're adding values to the end and (hopefully) not messing up the ordinal order of the enum.

omod/src/main/java/org/openmrs/module/fhir2/web/filter/ForwardingFilter.java
```diff
@@ -39,45 +40,26 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
...
-						
-						String newURI = requestURI.replace(prefix, replacement);
-						req.getRequestDispatcher(newURI).forward(req, res);
-						return;
-					}
-				}
+			String fhirVersion = String.valueOf(FhirVersionUtils.getFhirResourceVersion(request));
+			
+			String replacement;
+			switch (fhirVersion) {
```
ibacher, line 46: It would be better to switch on the enum value here, even if we do need to calculate the string value. That way we have a compile-time check of the various cases.

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirVersionUtils.java
ibacher: Can you give me some more context for this? What does the new version of `OpenmrsFhirAddressStrategy` look like? What does the test case that you are using look like?

ibacher:
@VaishSiddharth I've rebased this on the latest changes and squashed things down.

PR #161, opened 2020-05-06
FM2-156:Update Patient DAO and Service

HerbertYiga:
https://issues.openmrs.org/browse/FM2-156

Description

Update Patient DAO and Service

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirPatientDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirPatientDaoImpl.java
@@ -41,7 +41,7 @@ public class FhirPatientDaoImpl extends BasePersonDao<Patient> implements FhirPa
 	}
 	
 	@Override
-	public Patient getPatientByUuid(String uuid) {
+	public Patient get(String uuid) {
 		return (Patient) getSessionFactory().getCurrentSession().createCriteria(Patient.class).add(eq("uuid", uuid))
 		        .uniqueResult();
 	}
--- a/api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirPatientServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirPatientServiceImpl.java
@@ -22,7 +22,9 @@ import org.hl7.fhir.r4.model.Identifier;
 import org.hl7.fhir.r4.model.Patient;
 import org.openmrs.PatientIdentifierType;
 import org.openmrs.module.fhir2.api.FhirPatientService;
+import org.openmrs.module.fhir2.api.dao.FhirDao;
 import org.openmrs.module.fhir2.api.dao.FhirPatientDao;
+import org.openmrs.module.fhir2.api.translators.OpenmrsFhirTranslator;
 import org.openmrs.module.fhir2.api.translators.PatientTranslator;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
@@ -31,7 +33,8 @@ import org.springframework.transaction.annotation.Transactional;
 @Component
 @Transactional
 @Setter(AccessLevel.PACKAGE)
-public class FhirPatientServiceImpl implements FhirPatientService {
+
+public class FhirPatientServiceImpl extends BaseFhirService<Patient, org.openmrs.Patient> implements FhirPatientService {
 	
 	@Autowired
 	private PatientTranslator translator;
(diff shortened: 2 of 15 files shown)


sherrif10:
Thanks herbert for taking this up, You can also change in tests classes like how you did here but on only person classes

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirPatientDaoImpl.java
```diff
@@ -41,7 +43,7 @@ public Patient getPatientById(Integer id) {
 	}
 	
 	@Override
-	public Patient getPatientByUuid(String uuid) {
+	public Patient get(String uuid) {
```
ibacher, line 46: Technically, you should be able to remove this entire method, as its implemented in `BaseFhirDao` now.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirPatientServiceImpl.java
```diff
@@ -41,8 +43,8 @@
 	
 	@Override
 	@Transactional(readOnly = true)
-	public Patient getPatientByUuid(String uuid) {
-		return translator.toFhirResource(dao.getPatientByUuid(uuid));
+	public Patient get(String uuid) {
+		return translator.toFhirResource(dao.get(uuid));
```
ibacher, line 47: Technically, you should be able to remove this entire method as it's implemented in `BaseFhirService` now.

HerbertYiga: hi @ian  did you mean the get() for this(String uuid)?

HerbertYiga: @ibacher hei ian,what do you think of classess like FhirPatientDaoImpl,it extends BasePersonDao and makes use of its methods,are we dropping it,if so what about some of its methods that we are using??thanks

VaishSiddharth: I don't think that you need to drop `BasePersonDao` obviously we need methods like `searchForPatients` etc. I guess @ibacher meant to say that you should remove `getPatientByUuid` method from Dao as well as from the Service class as the `get` method is already implemented in the `BaseFhirService`

ibacher: Yes, what @VaishSiddharth said. The simple get, create, update, delete methods should all be implemented by BaseFhirDao. The addition stuff shared among person objects (Person, Provider, Patient, User) should still live in BasePersonDao.

HerbertYiga: sicne FhirPatientDaoImpl extends BasePersonDao arleady,i am wondering on the work round to have the two extended ie plus BaseFhirDao. cc  @ibacher @VaishSiddharth

VaishSiddharth: @HerbertYiga So if you see closely in `BasePersonDao` it already extends `BaseFhirDao` so no need to extend that class again. And I don't think we can actually extend two classes.

ibacher: No, you can only extend one class at a time in Java... the alternative is multiple inheritance, which [can get messy](https://en.wikipedia.org/wiki/Multiple_inheritance#The_diamond_problem). That said, the class hierarchy can be as deep as you want so `FhirPatientDaoImpl` -> `BasePersonDao` -> `BaseFhirDao` -> `BaseDao` is fine. That way, `FhirPatientDaoImpl` has all the functionality of `BaseDao` and the functionality of `BaseFhirDao` and the functionality of `BasePersonDao`.

HerbertYiga:
@ibacher  @VaishSiddharth   got this passing localy but it looks like the CI breaks with this error (java.lang.NoClassDefFoundError: Could not initialize class org.openmrs.Obs) any thoughts?

VaishSiddharth:
@HerbertYiga It works on my machine as well. I would suggest you to rebase this branch with master.

HerbertYiga:
Ok thanks i really appreciate your efforts @VaishSiddharth

HerbertYiga:
@VaishSiddharth @ibacher   rebased this as suggested,though its still breaking

HerbertYiga:
CI fails with the same logs as previous @VaishSiddharth cc @ibacher r

ibacher:
@HerbertYiga Thanks for this! It has been committed [here](https://github.com/openmrs/openmrs-module-fhir2/commit/f619d930c5a90f8c36b7c015b1ad76b30888cd2d).

PR #162, opened 2020-05-06
Update Concept DAO and Service

HerbertYiga:
https://issues.openmrs.org/browse/FM2-150

Description
Update Concept DAO and Service

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/FhirConceptService.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/FhirConceptService.java
@@ -15,7 +15,7 @@ import org.openmrs.Concept;
 
 public interface FhirConceptService {
 	
-	Optional<Concept> getConceptByUuid(String uuid);
+	Concept getConceptByUuid(String uuid);
 	
 	Optional<Concept> getConceptBySourceNameAndCode(String sourceName, String code);
 }
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/FhirConceptDao.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/FhirConceptDao.java
@@ -11,11 +11,13 @@ package org.openmrs.module.fhir2.api.dao;
 
 import java.util.Optional;
 
+import javax.validation.constraints.NotNull;
+
 import org.openmrs.Concept;
 
-public interface FhirConceptDao {
+public interface FhirConceptDao extends FhirDao<Concept> {
 	
-	Optional<Concept> getConceptByUuid(String uuid);
+	Concept get(@NotNull String uuid);
 	
 	Optional<Concept> getConceptBySourceNameAndCode(String sourceName, String code);
 }
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConceptDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConceptDaoImpl.java
@@ -9,6 +9,8 @@
  */
 package org.openmrs.module.fhir2.api.dao.impl;
 
+import javax.validation.constraints.NotNull;
+
 import java.util.Optional;
 
 import lombok.AccessLevel;
@@ -21,14 +23,17 @@ import org.springframework.stereotype.Component;
 
 @Component
 @Setter(AccessLevel.PACKAGE)
-public class FhirConceptDaoImpl implements FhirConceptDao {
+public class FhirConceptDaoImpl extends BaseFhirDao<Concept> implements FhirConceptDao {
 	
 	@Autowired
 	private ConceptService conceptService;
 	
 	@Override
-	public Optional<Concept> getConceptByUuid(String uuid) {
-		return Optional.ofNullable(conceptService.getConceptByUuid(uuid));
+	public Concept get(@NotNull String uuid) {
+		if (conceptService.getConceptByUuid(uuid) != null) {
+			return conceptService.getConceptByUuid(uuid);
+		}
+		return null;
 	}
 	
 	@Override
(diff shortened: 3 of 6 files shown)


sherrif10:
Herbert there is some work that is needed here, this should follow the same convention of how you did, the personDAO and service ticket

api/src/main/java/org/openmrs/module/fhir2/api/dao/FhirConceptDao.java
```diff
@@ -13,7 +13,7 @@
 
 import org.openmrs.Concept;
 
-public interface FhirConceptDao {
+public interface FhirConceptDao extends FhirDao<Concept> {
 	
 	Optional<Concept> getConceptByUuid(String uuid);
```
sherrif10, line 18: Change this from  getConceptByUuid(String uuid);  to get(String uuid);, You could also Check FhirConceptService.java

api/src/main/java/org/openmrs/module/fhir2/api/dao/FhirConceptDao.java
```diff
@@ -13,7 +13,7 @@
 
 import org.openmrs.Concept;
 
-public interface FhirConceptDao {
+public interface FhirConceptDao extends FhirDao<Concept> {
```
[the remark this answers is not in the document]
sherrif10: Did you have a look at how we did it
  `https://github.com/openmrs/openmrs-module-fhir2/commit/256c7806ecb8152b6d07d2043da878c14204eff6`, Some classes are inherited thats why the travis is complaining of whether some class is abstract or not

tendomart:
@HerbertYiga  did you look at this 
api: Compilation failure
[ERROR] /home/travis/build/openmrs/openmrs-module-fhir2/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConceptDaoImpl.java:[24,8] org.openmrs.module.fhir2.api.dao.impl.FhirConceptDaoImpl is not abstract and does not override abstract method search(org.openmrs.module.fhir2.api.search.param.SearchParameterMap,int,int) in org.openmrs.module.fhir2.api.dao.FhirDao

api/src/main/java/org/openmrs/module/fhir2/api/dao/FhirConceptDao.java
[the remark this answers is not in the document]
HerbertYiga: @sharif, ohh yaa i did but stil it goes back to my comments above hi @ibacher   what could be your comments on this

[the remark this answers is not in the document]
ibacher: Yeah, I'd drop the optional... at least until we can apply it consistently across the API. Thanks for picking up on this.

HerbertYiga:
@ibacher   hei ian,what do you think of classess like FhirPatientDaoImpl,it extends BasePersonDao and makes use of its methods,are we dropping it,if so what about some of its methods that we are using??thanks

api/src/main/java/org/openmrs/module/fhir2/api/dao/FhirConceptDao.java
```diff
@@ -11,11 +11,13 @@
 
 import java.util.Optional;
 
+import javax.validation.constraints.NotNull;
+
 import org.openmrs.Concept;
 
-public interface FhirConceptDao {
+public interface FhirConceptDao extends FhirDao<Concept> {
```
HerbertYiga, line 18: @rnjn bacher hi ian,i got realise that we don't have a model for concept under org.hl7.fhir.r4.model,any thoughts for that since this is needed around here public interface FhirConceptDao extends FhirDao<Concept>

VaishSiddharth: @HerbertYiga I think OpenMRS `Concept` maps to `CodeableConcept` in Fhir. Hope that helps. 

cc @ibacher

ibacher: @HerbertYiga: It's exactly what @VaishSiddharth said. We aren't actually publishing any FHIR-related concept resources. We're just using `CodeableConcept` in a few places where it's required.

api/src/main/java/org/openmrs/module/fhir2/api/FhirConceptService.java
```diff
@@ -13,7 +13,7 @@
 
 import org.openmrs.Concept;
 
-public interface FhirConceptService {
+public interface FhirConceptService  extends FhirService<Concept> {
 	
 	Optional<Concept> getConceptByUuid(String uuid);
```
ibacher, line 18: We should also remove the Optional part here.

HerbertYiga:
heilo ,on using CodeableConcept, i get this (Bound mismatch: The type CodeableConcept is not a valid substitute for the bounded parameter <T extends IAnyResource> of the type FhirService<T>)  @ibacher  @VaishSiddharth

VaishSiddharth:
@HerbertYiga I think that you don't need to `extends FhirService<?>` here.
But I am not very sure
cc @ibacher

ibacher:
@HerbertYiga Oh! That's a good point. I forgot that we have service classes that return FHIR datatypes instead of Resource. The other one that's a bit weird is FhirGlobalPropertyService. In this case, we should change the DAO to extend `BaseFhirDao`, but not try to implement the `FhirService` or `BaseFhirService` methods. For API consistency, I would change the `getConceptByUuid()` method to be called `get()`, but otherwise the service layer can be left alone.

I'll update the ticket accordingly.

HerbertYiga:
heilo,@VaishSiddharth   made some changes on this too as suggested by @ibacher

ibacher:
@HerbertYiga Thanks for this! It has been committed [here](https://github.com/openmrs/openmrs-module-fhir2/commit/dc2619748acf6f49891422185a166fde6d1f6609).

PR #171, opened 2020-05-13
FM2-164: Update Patient resource to support paging

corneliouzbett:

Update patient resource search to return paged results

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-164

  No? -> execute above command


api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirPatientDaoImpl.java
```diff
@@ -104,4 +115,52 @@ protected String paramToProp(String param) {
 		
 		return super.paramToProp(param);
 	}
+	
+	private void handleNames(List<PropParam<?>> params, Criteria criteria) {
+		AtomicReference<StringAndListParam> name = new AtomicReference<>();
```
ibacher, line 120: So we need `AtomicReference`s here?

corneliouzbett: Not necessarily, I used it to avoid duplicate `alias`. That was the first thing I thought of but I think there's a better of handling this. 

The goal here is to call `handleNames(criteria, name, given, family)` once in a query to avoid duplicate alias. Alternative is to call it multiple times with different alias each the method is called. Didn't like the query generated. I chose the first option of storing passed in values temporarily.

ibacher: Doesn't it just make sense to avoid the using the lambdas altogether:

```java
private void handleNames(List<PropParam<?>> params, Criteria criteria) {
	StringAndListParam name = null;
	StringAndListParam given = null;
	StringAndListParam family = null;

	for (PropParam<?> param : params) {
		switch (param.getPropertyName()) {
			case FhirConstants.NAME_PROPERTY:
				name = (StringAndListParam) param.getParam();
				break;
			case FhirConstants.GIVEN_PROPERTY:
				given = (StringAndListParam) param.getParam();
				break;
			case FhirConstants.FAMILY_PROPERTY:
				family = (StringAndListParam) param.getParam();
				break;
		}
	}

	handleNames(criteria, name, given, family);
}
```

corneliouzbett: That make sense ☺️

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirPatientDaoImpl.java
```diff
@@ -104,4 +114,53 @@ protected String paramToProp(String param) {
...
+					break;
+				case FhirConstants.COUNTRY_PROPERTY:
+					country = ((StringAndListParam) param.getParam());
+					break;
+				case FhirConstants.POSTAL_CODE_PROPERTY:
+					postalCode = ((StringAndListParam) param.getParam());
+					break;
+				case FhirConstants.STATE_PROPERTY:
+					state = ((StringAndListParam) param.getParam());
+					break;
```
ibacher, line 157: Really minor point, but could we order these city, state, postal code, property (that's roughly ascending order of size (except that postal code probably belong at the front of the list) just to keep things consistent (my main hang-up is making sure "state" is before "country" since it isn't always obvious to people that "state" means state in the sense of "US state" or "Australian state" rather than "country").

omod/src/test/java/org/openmrs/module/fhir2/providers/PatientFhirResourceProviderTest.java
```diff
@@ -303,14 +316,15 @@ public void searchForPatients_shouldReturnMatchingBundleOfPatientsByCountry() {
...
-		    countryParam, null);
+		IBundleProvider results = resourceProvider.searchPatients(null, null, null, null, null, null, null, null, null, null,
+		    null, countryParam, null);
 		
 		assertThat(results, notNullValue());
-		assertThat(results.isResource(), is(true));
-		assertThat(results.getEntry().size(), greaterThanOrEqualTo(1));
+		assertThat(results.getResources(0, 1), hasSize(equalTo(1)));
+		assertThat(results.getResources(0, 1).get(0).fhirType(), is("Patient"));
+		assertThat(results.getResources(0, 1).get(0).getIdElement().getIdPart(), is(PATIENT_UUID));
```
ibacher, line 327: I would probably prefer caching the call of `results.getResources(0, 1)` and reusing that as each time we invoke that, it makes a query to the database and, while it's not a huge deal, I'd like to keep the unit tests as quick as possible so people continue to *want* to run them.

PR #185, opened 2020-05-27
FM2-178: Add patient search parameter where subject is allowed

reagan-meant:
https://issues.openmrs.org/browse/FM2-178
Added an option parameter "patient" that is evaluated before subject.

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/providers/r3/AllergyIntoleranceFhirResourceProvider.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/providers/r3/AllergyIntoleranceFhirResourceProvider.java
@@ -34,6 +34,7 @@ import org.hl7.fhir.dstu3.model.IdType;
 import org.hl7.fhir.dstu3.model.Patient;
 import org.hl7.fhir.dstu3.model.Resource;
 import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.hl7.fhir.r4.model.Observation;
 import org.hl7.fhir.r4.model.Provenance;
 import org.openmrs.module.fhir2.api.FhirAllergyIntoleranceService;
 import org.openmrs.module.fhir2.providers.util.FhirProviderUtils;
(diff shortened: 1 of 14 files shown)


omod/src/main/java/org/openmrs/module/fhir2/providers/EncounterFhirResourceProvider.java
```diff
@@ -69,7 +69,12 @@ public Bundle searchEncounter(@OptionalParam(name = Encounter.SP_DATE) DateRange
 	                Practitioner.SP_GIVEN, Practitioner.SP_FAMILY,
 	                Practitioner.SP_NAME }, targetTypes = Practitioner.class) ReferenceParam participantReference,
 	        @OptionalParam(name = Encounter.SP_SUBJECT, chainWhitelist = { "", Patient.SP_IDENTIFIER, Patient.SP_GIVEN,
-	                Patient.SP_FAMILY, Patient.SP_NAME }, targetTypes = Patient.class) ReferenceParam subjectReference) {
+					Patient.SP_FAMILY, Patient.SP_NAME }, targetTypes = Patient.class) ReferenceParam subjectReference,
+			@OptionalParam(name = Encounter.SP_PATIENT, chainWhitelist = { "", Patient.SP_IDENTIFIER, Patient.SP_GIVEN,
+	                Patient.SP_FAMILY, Patient.SP_NAME }, targetTypes = Patient.class) ReferenceParam patientReference) {
+						if(patientReference != null){
+							subjectReference = patientReference;
+						}
```
VaishSiddharth, line 77: @reagan-meant I don't think that we should do this here. 

cc @ibacher

VaishSiddharth:
@reagan-meant I would suggest that you should definitely rebase to master because lot of changes are made recently.

Support for DSTU3 is added because of which many files are refactored.

reagan-meant:
Oooh yes...I didn't rebase... Will look into it tomorrow

omod/src/main/java/org/openmrs/module/fhir2/providers/EncounterFhirResourceProvider.java
ibacher: Actually, this is explicitly what we do want to have happen. The `patient` search parameter is a sub-set of the `subject` search parameter where the `subject` is a Patient. So the `patient` parameter is more exact than the subject parameter. I suppose the other thing we could do is throw an exception here.

ibacher:
@reagan-meant Thanks for this! The change you're making looks correct. It should apply to a bunch more resources (so where we support `patient` we should also support `subject` and vice-versa). I'd also like to see unit tests for this. We have lots of examples available in the source, but let me know if you have any questions. Again, thanks for your work on this!

omod/src/main/java/org/openmrs/module/fhir2/providers/EncounterFhirResourceProvider.java
reagan-meant: @ibacher what would be the potential cause the exception?

PR #190, opened 2020-06-01
FM2-193: Create method for generating default narratives provided by HAPI FHIR

ShivanshRakesh:
Implemented a method for generating Default Narratives provided by HAPI FHIR.

## Issue I worked on
see https://issues.openmrs.org/browse/FM2-193

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/narrative/impl/DefaultNarrativeGeneratorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/narrative/impl/DefaultNarrativeGeneratorImpl.java
@@ -1,0 +1,46 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.narrative.impl;
+
+import ca.uhn.fhir.context.FhirContext;
+import ca.uhn.fhir.narrative.DefaultThymeleafNarrativeGenerator;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.hl7.fhir.r4.model.Narrative;
+import org.openmrs.module.fhir2.api.narrative.DefaultNarrativeGenerator;
+
+public class DefaultNarrativeGeneratorImpl implements DefaultNarrativeGenerator {
+	
+	@Override
+	public Narrative generateDefaultNarrative(IBaseResource iBaseResource) {
+		if (resourcesWithHAPIDefaultNarratives.contains(iBaseResource.getClass().getSimpleName())) {
+			return generateHAPIDefaultNarrative(iBaseResource);
+		}
+		return null;
+	}
+	
+	@Override
+	public Narrative generateHAPIDefaultNarrative(IBaseResource iBaseResource) {
+		FhirContext fhirContext = FhirContext.forR4();
+		fhirContext.setNarrativeGenerator(new DefaultThymeleafNarrativeGenerator());
+		String narrativeStr = fhirContext.newJsonParser().setPrettyPrint(true).encodeResourceToString(iBaseResource);
+		
+		JsonParser jsonParser = new JsonParser();
+		JsonObject jsonObject = (JsonObject) jsonParser.parse(narrativeStr);
+		JsonObject narrativeText = jsonObject.getAsJsonObject("text");
+		
+		Narrative narrative = new Narrative();
+		narrative.setStatusAsString(narrativeText.get("status").getAsString());
+		narrative.setDivAsString(narrativeText.get("div").getAsString());
+		
+		return narrative;
+	}
+}
--- a/pom.xml
+++ b/pom.xml
@@ -200,6 +200,11 @@
 			</dependency>
 			<dependency>
 				<groupId>ca.uhn.hapi.fhir</groupId>
+				<artifactId>hapi-fhir-narrativegenerator</artifactId>
+				<version>${hapifhirVersion}</version>
+			</dependency>
+			<dependency>
+				<groupId>ca.uhn.hapi.fhir</groupId>
 				<artifactId>hapi-fhir-converter</artifactId>
 				<version>${hapifhirVersion}</version>
 			</dependency>
@@ -272,6 +277,26 @@
 	</dependencyManagement>
 
 	<dependencies>
+		<dependency>
+			<groupId>org.thymeleaf</groupId>
+			<artifactId>thymeleaf</artifactId>
+			<version>3.0.11.RELEASE</version>
+		</dependency>
+		<dependency>
+			<groupId>commons-io</groupId>
+			<artifactId>commons-io</artifactId>
+			<version>2.6</version>
+		</dependency>
+		<dependency>
+			<groupId>com.github.ben-manes.caffeine</groupId>
+			<artifactId>caffeine</artifactId>
+			<version>2.8.2</version>
+		</dependency>
+		<dependency>
+			<groupId>com.google.code.gson</groupId>
+			<artifactId>gson</artifactId>
+			<version>2.8.5</version>
+		</dependency>
 		<dependency>
 			<groupId>org.openmrs.api</groupId>
 			<artifactId>openmrs-api</artifactId>
--- a/api/src/main/java/org/openmrs/module/fhir2/api/narrative/DefaultNarrativeGenerator.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/narrative/DefaultNarrativeGenerator.java
@@ -1,0 +1,26 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.narrative;
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.hl7.fhir.instance.model.api.IBaseResource;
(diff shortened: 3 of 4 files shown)


tendomart:
I think from my side , the only thing missing is notifying the Reviewer's at  https://issues.openmrs.org/browse/FM2-193 by tagging them.

ShivanshRakesh:
Thanks for reviewing @tendomart! :)
Sure, will notify the reviewers from now on.

api/src/main/java/org/openmrs/module/fhir2/api/narrative/impl/DefaultNarrativeGenerator.java
```diff
@@ -0,0 +1,51 @@
...
+		narrative.setStatusAsString(narrativeText.get("status").getAsString());
+		narrative.setDivAsString(narrativeText.get("div").getAsString());
+		
+		return narrative;
+	}
+	
+	@Override
+	public Narrative generateCustomDefaultNarrative(IBaseResource iBaseResource) {
+		return null;
+	}
```
corneliouzbett, line 50: Instead of returning null, just remove the method completely.

api/src/main/java/org/openmrs/module/fhir2/api/narrative/IDefaultNarrativeGenerator.java
```diff
@@ -0,0 +1,28 @@
...
+package org.openmrs.module.fhir2.api.narrative;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.hl7.fhir.r4.model.Narrative;
+
+public interface IDefaultNarrativeGenerator {
+
```
corneliouzbett, line 19: `DefaultNarrativeGenerator` over `IDefaultNarrativeGenerator`. Check module class naming convections [here](https://github.com/openmrs/openmrs-module-fhir2#class-naming-conventions)

api/src/main/java/org/openmrs/module/fhir2/api/narrative/impl/DefaultNarrativeGenerator.java
```diff
@@ -0,0 +1,51 @@
...
+	
+	@Override
+	public Narrative generateHAPIDefaultNarrative(IBaseResource iBaseResource) {
+		FhirContext fhirContext = FhirContext.forR4();
+		fhirContext.setNarrativeGenerator(new DefaultThymeleafNarrativeGenerator());
+		String narrativeStr = fhirContext.newJsonParser().setPrettyPrint(true).encodeResourceToString(iBaseResource);
+		
+		JsonParser jsonParser = new JsonParser();
+		JsonObject jsonObject1 = (JsonObject) jsonParser.parse(narrativeStr);
```
varung-31, line 37: Why can't we just name this as jsonObject ?

api/src/main/java/org/openmrs/module/fhir2/api/narrative/IDefaultNarrativeGenerator.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.hl7.fhir.r4.model.Narrative;
+
+public interface IDefaultNarrativeGenerator {
+	
+	ArrayList<String> resourcesWithHAPIDefaultNarratives = new ArrayList<>(
```
varung-31, line 20: `List<String> ... = Arrays.asList(...);` should be a better way to write this I guess.

api/src/main/java/org/openmrs/module/fhir2/api/narrative/impl/DefaultNarrativeGenerator.java
ShivanshRakesh: Okay. Will remove it for now. Actually it is another task in the scope of the parent issue ([FM2-192](https://issues.openmrs.org/browse/FM2-192)), so I just added the method for now.

api/src/main/java/org/openmrs/module/fhir2/api/narrative/IDefaultNarrativeGenerator.java
ShivanshRakesh: Okay, will change the name.

api/src/main/java/org/openmrs/module/fhir2/api/narrative/impl/DefaultNarrativeGenerator.java
ShivanshRakesh: Ah yes. Actually I previously declared an object with name `jsonObject` but then removed it. Forgot to rename this one. Thanks for pointing it out!

api/src/main/java/org/openmrs/module/fhir2/api/narrative/IDefaultNarrativeGenerator.java
ShivanshRakesh: Okay, will look into it. Thanks!

ShivanshRakesh:
@corneliouzbett @varung-31 PTAL! Have made the changes.

pom.xml
```diff
@@ -272,6 +277,26 @@
...
+		<dependency>
+			<groupId>com.github.ben-manes.caffeine</groupId>
+			<artifactId>caffeine</artifactId>
+			<version>2.8.2</version>
+		</dependency>
+		<dependency>
+			<groupId>com.google.code.gson</groupId>
+			<artifactId>gson</artifactId>
+			<version>2.8.5</version>
+		</dependency>
```
ibacher, line 299: Do we need this dependency on GSON?

api/src/main/java/org/openmrs/module/fhir2/api/narrative/impl/DefaultNarrativeGeneratorImpl.java
```diff
@@ -0,0 +1,46 @@
...
+		
+		JsonParser jsonParser = new JsonParser();
+		JsonObject jsonObject = (JsonObject) jsonParser.parse(narrativeStr);
+		JsonObject narrativeText = jsonObject.getAsJsonObject("text");
+		
+		Narrative narrative = new Narrative();
+		narrative.setStatusAsString(narrativeText.get("status").getAsString());
+		narrative.setDivAsString(narrativeText.get("div").getAsString());
+		
+		return narrative;
```
ibacher, line 44: This seems a little... roundabout. We first encode the resources as a JSON document in order to apply the `DefaultThymeleafNarrativeGenerator` and then parse the JSON object to extract the narrative to be returned from this method. Why do we need this conversion to JSON and back? Why do we need a `Narrative` object at all? Couldn't this just have been solved by calling the `setNarrativeGenerator()` function inside the FhirRestServlet initialisation method?

ShivanshRakesh: @ibacher Actually the conversion from JSON to back is done to fetch the `text` element from the narrative string which is generated. For the `Narrative` object, I was just following the approach that was decided upon, after discussing with @HerbertYiga  and posting on the thread [here](https://talk.openmrs.org/t/gsoc20-add-support-for-fhir-narratives/28408/31?u=iamsr).

pom.xml
ShivanshRakesh: The dependency is required for the JSON conversions which are done.

api/src/main/java/org/openmrs/module/fhir2/api/narrative/impl/DefaultNarrativeGeneratorImpl.java
ibacher: Well, I think what I proposed ends up being much simpler to implement.

ShivanshRakesh: Okay, so how do you plan to link the resources to their narratives? The reason I was creating these `Narrative` objects was to call the generator method from the translator and assign the narratives before returning the translated resource.

ibacher: Well, if HAPI will take a Narrative Generator and simply apply to the resource once it's generated for us, we don't need to touch the translators at all. Ultimately, that makes things simpler than I thought it was going to be. Unless there's some advantage to capturing and manipulating the narrative I'm missing? If there is, the correct thing to do would seem to be to call `DefaultThymeleafNarrativeGenerator#populateResourceNarrative()` where we could at least pull the narrative out of the resource without jumping through the hoops of dumping to JSON and parsing back to Java objects.

ShivanshRakesh: Ah I see. So in that case, we won't be needing the `generateHAPIDefaultNarrative()` method at all and instead, we shall be setting a `FhirContext` in the `initialize` method of [FhirRestServlet](https://github.com/openmrs/openmrs-module-fhir2/blob/e3114b0aef80ae7f5c5d6bc43becfda8164e66e6/omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java) class; after which we will set the `DefaultThymeleafNarrativeGenerator()` for the `FhirContext` using `setNarrativeGenerator`. 

Though, there would be a check for _provision of HAPI default narrative_ involved similar to what is being done in the `generateDefaultNarrative()` method, which can be done using a method in the `FhirUtils` class I suppose?

Did I get it right?

ShivanshRakesh:
@ibacher I guess I shall close this PR for now since there's an entire change in approach. I'll create a new one following [the approach you suggested](https://github.com/openmrs/openmrs-module-fhir2/pull/190/files/eaaee3d1ef68f436a10cf49331eb577dcab22e0f#r433992912).

api/src/main/java/org/openmrs/module/fhir2/api/narrative/impl/DefaultNarrativeGeneratorImpl.java
ShivanshRakesh: @ibacher One thing though. How are we going to get the resource name at runtime if we put this into the `initialize` method? I mean, the resource name is needed to check for presence of HAPI Default Narrative right?

ibacher: We may end up needing to use the `CustomThymeleafNarrativeGenerator` instead of the `DefaultThymeleafNarrativeGenerator` so we can support custom templates, but as a first pass, this looks good. I would suggest we avoid hard-coding a list of narratives supported by the HAPI FHIR library, because this is bound to be an evolving target. I think instead, I would populate the `propertyFile` property of the `CustomThymeleafNarrativeGenerator` with both the default resource properties and the custom resource properties file.

ShivanshRakesh: Yeah, even I was thinking about using just the `CustomThymeleafNarrativeGenerator` and include the HAPI defaults in the property file too, since otherwise there could be complications in getting the resource name and checking through the list during runtime.

PR #194, opened 2020-06-04
FM2-193: Generate default narratives provided by HAPI FHIR

ShivanshRakesh:
Set Thymeleaf Narrative Generator (with `propFile` as the properties file of HAPI FHIR) while initialising `FhirRestServlet`.

## Issue I worked on
see https://issues.openmrs.org/browse/FM2-193

### DIFF
--- a/omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
+++ b/omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
@@ -12,6 +12,7 @@ package org.openmrs.module.fhir2.web.servlet;
 import java.util.Collection;
 
 import ca.uhn.fhir.context.FhirContext;
+import ca.uhn.fhir.narrative.CustomThymeleafNarrativeGenerator;
 import ca.uhn.fhir.rest.api.EncodingEnum;
 import ca.uhn.fhir.rest.server.FifoMemoryPagingProvider;
 import ca.uhn.fhir.rest.server.IResourceProvider;
@@ -60,6 +61,9 @@ public class FhirRestServlet extends RestfulServer {
 		setPagingProvider(pp);
 		setDefaultResponseEncoding(EncodingEnum.JSON);
 		registerInterceptor(loggingInterceptor);
+		
+		getFhirContext().setNarrativeGenerator(
+		    new CustomThymeleafNarrativeGenerator(FhirConstants.HAPI_DEFAULT_NARRATIVES_PROPERTY_FILE));
 	}
 	
 	@Override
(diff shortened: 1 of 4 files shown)


ShivanshRakesh:

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
```diff
@@ -92,4 +97,9 @@ public void setResourceProviders(Collection<IResourceProvider> theProviders) {
 	public void setServerAddressStrategy(IServerAddressStrategy theServerAddressStrategy) {
 		super.setServerAddressStrategy(theServerAddressStrategy);
 	}
+	
+	protected void setDefaultNarrativeGenerator() {
+		this.setFhirContext(FhirContext.forR4());
+		getFhirContext().setNarrativeGenerator(new CustomThymeleafNarrativeGenerator(hapiDefaultNarrativesPropFile));
+	}
```
ibacher, line 104: Please don't do this this way. The problem here is that you're hard-coding the `FhirContext` which is actually being set by Spring (see above). The correct thing to do here would be to replace where you have the call to `this.setDefaultNarrativeGenerator()` with the last line of code:

```java
getFhirContext().setNarrativeGenerator(new CustomThymeleafNarrativeGenerator(hapiDefaultNarrativesPropFile));
```

That way, it only depends on the FhirContext that is set by Spring, so that we don't break support for DSTU3 while doing this.

It might also be worth thinking about if we could load the `ThymeleafNarrativeGenerator` from Spring, as this might give us more flexibility in defining the property files to load, instead of needing to know about them ahead of time.

ShivanshRakesh: Okay, I'll do the changes.

ShivanshRakesh:
> @ShivanshRakesh Thanks for this! This is looking much cleaner. I've left some comments on some things I'd like to see changed.

@ibacher Thanks! Okay, will do the changes.

ShivanshRakesh:
@ibacher PTAL!

pom.xml
```diff
@@ -319,6 +319,21 @@
...
+		<dependency>
+			<groupId>commons-io</groupId>
+			<artifactId>commons-io</artifactId>
+			<version>2.6</version>
+		</dependency>
+		<dependency>
+			<groupId>com.github.ben-manes.caffeine</groupId>
+			<artifactId>caffeine</artifactId>
+			<version>2.8.2</version>
+		</dependency>
```
ibacher, line 336: So the set of dependencies here should be kept as minimal as possible. These should be moved to the `dependencyManagement` section and then included in the POM where they are necessary (possibly only in the OMOD). I'd also drop the dependency on commons-io or change it's scope to "provided", since it's included in the platform.

ShivanshRakesh:
Thanks @ibacher :)

PR #198, opened 2020-06-08
FM2-194: Create default narratives for AllergyIntolerance, MedicationRequest, and Condition

ShivanshRakesh:
Added OpenMRS Narratives for the following resources:
* AllergyIntolerance
* Condition
* MedicationRequest

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-194


ShivanshRakesh:
I've opened this PR so that it'll be easier to track and check the Narrative Templates I'll be adding.
cc @ibacher @HerbertYiga @suthagar23

ShivanshRakesh:
I've added narrative templates for the following:
* AllergyIntolerance
* MedicationRequest
* Generic datatypes which have been used in the above (eg. Reference, etc)

Along with this, I've tried to create an object (following one of the HAPI examples) for each resource so that the narrative template could be tested. I've created a `DefaultNarrativeGeneratorForTest` class which contains a narrative generator method just for the testing purpose.

ShivanshRakesh:
I see the build is failing because I haven't added support for DSTU3. Will do the changes required.

omod/src/test/java/org/openmrs/module/fhir2/narrative/DefaultNarrativeTemplateTest.java
```diff
@@ -0,0 +1,230 @@
...
+		medicationRequest.addDetectedIssue().setReference("DetectedIssue/allergy");
+		medicationRequest.addEventHistory().setReference("#signature").setDisplay("Author's Signature");
+		
+		assertThat(testHelper(new MedicationRequest()).getDivAsString(), notNullValue());
+		assertThat(testHelper(medicationRequest).getDivAsString(), notNullValue());
+	}
+	
+	public Narrative testHelper(IBaseResource iBaseResource) {
+		Narrative narrative = DefaultNarrativeGeneratorForTest.generateDefaultNarrative(iBaseResource);
+		System.out.println(iBaseResource.getClass().getSimpleName() + " Narrative Status:" + narrative.getStatusAsString());
```
suthagar23, line 225: Better to use logger if you really need or you can remove the `System.out.println` from the code

ShivanshRakesh: Thanks for reviewing @suthagar23 :) 
Actually I've put this just for convenience in checking the output for now. It isn't intended to be a part of the final commit.

omod/src/test/java/org/openmrs/module/fhir2/narrative/DefaultNarrativeTemplateTest.java
```diff
@@ -0,0 +1,230 @@
...
+		signature.setDataElement(new Base64BinaryType("dGhpcyBibG9iIGlzIHNuaXBwZWQ="));
+		provenance.addSignature(signature);
+		medicationRequest.addContained(provenance);
+		
+		medicationRequest.addIdentifier().setUse(Identifier.IdentifierUse.OFFICIAL)
+		        .setSystem("http://www.bmc.nl/portal/prescriptions").setValue("12345689");
+		medicationRequest.setStatus(MedicationRequest.MedicationRequestStatus.COMPLETED);
+		
+		CodeableConcept statusReason = new CodeableConcept();
+		statusReason.addCoding().setDisplay("Try another treatment first").setCode("altchoice")
```
HerbertYiga, line 142: @ibacher what do you think of @ShivanshRakesh   arranging his data into data sets ?

omod/src/test/java/org/openmrs/module/fhir2/narrative/DefaultNarrativeTemplateTest.java
```diff
@@ -0,0 +1,230 @@
...
+import org.hl7.fhir.r4.model.StringType;
+import org.hl7.fhir.r4.model.Timing;
+import org.hl7.fhir.r4.model.codesystems.AllergyintoleranceClinical;
+import org.hl7.fhir.r4.model.codesystems.AllergyintoleranceVerification;
+import org.junit.Test;
+
+public class DefaultNarrativeTemplateTest {
+	
+	@Test
+	public void testForAllergyIntolerance() {
```
ibacher, line 43: Always prefer the `should` style of naming tests over `test` to make it clear what the test is testing. For example, I would name this `shouldGenerateAnAllergyIntoleranceNarrative`.

ibacher: Rather than populating this resource completely in code like this, it seems like it would be simpler to download the file from [here](https://www.hl7.org/fhir/allergyintolerance-example.json), remove the "text" section and save that in the test resources, then load it.

For example:

```java
// ensure ctx is a FhirContext with the correct NarrativeGenerator
IParser parser = ctx.newJsonParser();

// load initial data (this is our "given" for the test)
AllergyIntolerance given = parser.parseResource(AllergyIntolerance.class, getClass().getClassLoader().getResourceAsStream("allergyintolerance-example.json"));

AllergyIntolerance result = parser.parseResource(AllergyIntolerance.class, parser.encodeResourceToString(allergyIntolerance));

// now our assertions
assertThat(result, notNullValue());
assertThat(result.getText(), notNullValue());
assertThat(result.getText().getStatusAsString(), equalTo("generated"));
// note that we should check the generated result against a reference of what we expect it to be
assertThat(result.getText().getDivAsString(), equalTo(getClass().getClassLoader().getResourceAsStream("expected/allergyintolerance-example.html"));
```

omod/src/test/java/org/openmrs/module/fhir2/narrative/DefaultNarrativeTemplateTest.java
```diff
@@ -0,0 +1,230 @@
...
+import org.hl7.fhir.r4.model.Range;
+import org.hl7.fhir.r4.model.Reference;
+import org.hl7.fhir.r4.model.Signature;
+import org.hl7.fhir.r4.model.StringType;
+import org.hl7.fhir.r4.model.Timing;
+import org.hl7.fhir.r4.model.codesystems.AllergyintoleranceClinical;
+import org.hl7.fhir.r4.model.codesystems.AllergyintoleranceVerification;
+import org.junit.Test;
+
+public class DefaultNarrativeTemplateTest {
```
ibacher, line 40: We should get away from referring to "DefaultNarrative". I understand that there's a difference  whether the narrative is generated from the HAPI defaults or ours, but by and large, we don't care about that difference. In my mind, it would almost make sense to break out a new series of classes to test narrative generation alone, so, e.g. having an `AllergyIntoleranceNarrativeTest` class, etc.

omod/src/test/java/org/openmrs/module/fhir2/narrative/DefaultNarrativeGeneratorForTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+import ca.uhn.fhir.context.FhirContext;
+import ca.uhn.fhir.narrative.CustomThymeleafNarrativeGenerator;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.hl7.fhir.r4.model.Narrative;
+import org.openmrs.module.fhir2.FhirConstants;
+
+public class DefaultNarrativeGeneratorForTest {
```
ibacher, line 20: Why is this a separate class?

omod/src/test/java/org/openmrs/module/fhir2/narrative/DefaultNarrativeTemplateTest.java
ibacher: Rather than using a logger (which is a good idea if we were going to output anything) or... manually reading the output, use assertions to check that these are the expected values. That's why they exist.

omod/pom.xml
```diff
@@ -79,6 +79,25 @@
...
+			<groupId>com.github.ben-manes.caffeine</groupId>
+			<artifactId>caffeine</artifactId>
+		</dependency>
+<!--	///	Just for testing ///    -->
+		<dependency>
+			<groupId>com.google.code.gson</groupId>
+			<artifactId>gson</artifactId>
+			<version>2.8.5</version>
+		</dependency>
+<!--	////////////////////////    -->
```
ibacher, line 100: Several things:

* Comments here are not very helpful. If we need a resource just for testing, it **must** to be included in the base POM with the `<scope>test</scope>`. Otherwise, this jar ends up bundled in the OMOD (by default, Maven artifacts are loaded with the scope `compile` unless you tell Maven otherwise).
* We should move away from using GSON at all. HAPI has moved from GSON to using Jackson, which should already be available on the class path.

omod/src/main/resources/defaultNarratives.properties
```diff
@@ -0,0 +1,35 @@
+#
```
ibacher, line 1: Why not just call this `narratives.properties`?

omod/src/main/resources/defaultNarratives/AllergyIntolerance.html
```diff
@@ -0,0 +1,153 @@
...
+            </tbody>
+        </table>
+    </th:block>
+</div>
+
+<!--/*-->
+</body>
+
+</html>
+<!--*/-->
```
ibacher, line 153: Always have a new line at the end of the file. This is primarily to prevent some Git weirdness issues.

omod/src/test/java/org/openmrs/module/fhir2/narrative/DefaultNarrativeTemplateTest.java
ibacher: @HerbertYiga If by this you mean "should this data be stored in a file and loaded from there" my answer would be yes. This should be a different mechanism from loading data into the database, if only because there are a number of layers between the raw data in the database and the narrative that we don't need to worry about here.

omod/src/test/java/org/openmrs/module/fhir2/narrative/DefaultNarrativeGeneratorForTest.java
ShivanshRakesh: Actually I thought to keep the generator separate. But I'll move it to the same class with the tests.

omod/pom.xml
ShivanshRakesh: @ibacher Actually I did not mean it that way. I meant that it won't actually be a part of the final commit. 

Also, for the `print` statements, I do not intend to keep it as a part of the final commit. I just used it to see the output for now. I've used `assert` statements in the tests which would be a part of the final commit.

But yes, I'll remove such code which I don't plan to keep in the final commit, i.e. the `google-gson` dependency and the `print` statements. Or maybe if I need it at all, I'll use Jackson as you suggested.

ShivanshRakesh:
@ibacher Thanks a lot! :) Sure, I'll do the required changes.

omod/src/test/java/org/openmrs/module/fhir2/narrative/DefaultNarrativeTemplateTest.java
ShivanshRakesh: @ibacher Thanks a lot for this! I myself was looking for alternative ways.

omod/pom.xml
ibacher: Ah! Ok... I see now!

ShivanshRakesh:
Have made the changes. PTAL!
cc @ibacher @HerbertYiga @suthagar23

omod/src/test/java/org/openmrs/module/fhir2/narrative/MedicationRequestNarrativeTest.java
```diff
@@ -0,0 +1,52 @@
...
+		FhirContext ctx = FhirContext.forR4();
+		ctx.setNarrativeGenerator(new CustomThymeleafNarrativeGenerator(FhirConstants.HAPI_NARRATIVES_PROPERTY_FILE,
+		        FhirConstants.OPENMRS_NARRATIVES_PROPERTY_FILE));
+		IParser parser = ctx.newJsonParser();
+		
+		MedicationRequest given = parser.parseResource(MedicationRequest.class,
+		    getClass().getClassLoader().getResourceAsStream(EXAMPLE_RESOURCE_PATH));
+		MedicationRequest result = parser.parseResource(MedicationRequest.class, parser.encodeResourceToString(given));
+		String expected = IOUtils.readLines(getClass().getClassLoader().getResourceAsStream(EXPECTED_NARRATIVE_PATH),
+		    Charsets.toCharset(Charset.defaultCharset())).get(0);
```
ShivanshRakesh, line 45: @ibacher PTAL. Doing it this way is fine right? 
Using just `equalTo(getClass().getClassLoader().getResourceAsStream(<path>)` raised errors since it returns a stream.

ibacher: Why not just use `IOUtils.resourceToString(EXPECTED_NARRATIVE_PATH, StandardCharsets.UTF_8)`? 

Also, some general points about unit tests. We want to structure unit tests so that it is clear what is being tested in each test. We also want to factor out the common setup parts so that things aren't repeating themselves too much.

For unit tests, I use this structure:

```java
// given - setup the preconditions

// when - do the thing being tested

// then - assertions to ensure state is right
```

Given this structure, I think this test should look something like this:

```java
// do this once per class
private FhirContext ctx = FhirContext.forR4();

private IParser parser;

@Before
private void setup() {
    ctx.setNarrativeGenerator(new CustomThymeleafNarrativeGenerator(FhirConstants.HAPI_NARRATIVES_PROPERTY_FILE,
        FhirConstants.OPENMRS_NARRATIVES_PROPERTY_FILE));
    parser = ctx.newJsonParser();
}

@Test
public void shouldGenerateMedicationRequestNarrative() throws IOException {
    // given
    MedicationRequest given = parser.parseResource(MedicationRequest.class,
         getClass().getClassLoader().getResourceAsStream(EXAMPLE_RESOURCE_PATH));

    // when
    MedicationRequest result = parser.parseResource(MedicationRequest.class, parser.encodeResourceToString(given));

    // then
    assertThat(result, notNullValue());
    assertThat(result.getText(), notNullValue());
    assertThat(result.getText().getStatusAsString(), equalTo("generated"));
    // if you don't do it in one line like this, I would load the expected narrative here
    // that way, if some error comes from just loading the expected result, we can at least be
    // assured that the other assertions passed
    assertThat(result.getText().getDivAsString(),
        equalTo(IOUtils.resourceToString(EXPECTED_NARRATIVE_PATH, StandardCharsets.UTF_8)));
}
```

The comments there don't need to be preserved. They're just to show you why I've put the line breaks where I did.

ShivanshRakesh: Okay! Will restructure the tests this way.

ShivanshRakesh: Actually I used `IOUtils.readlines()` because I just needed the first line, since I had to keep a newline in the end. Though, I guess I can use `IOUtils.resourceToString()` too and then split it at a newline.

Okay, will restructure the tests.

ShivanshRakesh: @ibacher I've done the changes. PTAL.

ShivanshRakesh: Done the changes.

ibacher:
@ShivanshRakesh Smaller PRs are preferable to larger PRs.

omod/src/test/java/org/openmrs/module/fhir2/narrative/AllergyIntoleranceNarrativeTest.java
```diff
@@ -0,0 +1,59 @@
...
+		    getClass().getClassLoader().getResourceAsStream(EXAMPLE_RESOURCE_PATH));
+		
+		AllergyIntolerance result = parser.parseResource(AllergyIntolerance.class, parser.encodeResourceToString(given));
+		
+		assertThat(result, notNullValue());
+		assertThat(result.getText(), notNullValue());
+		assertThat(result.getText().getStatusAsString(), equalTo("generated"));
+		assertThat(result.getText().getDivAsString(),
+		    equalTo(IOUtils.resourceToString(EXPECTED_NARRATIVE_PATH, StandardCharsets.UTF_8, getClass().getClassLoader())
+		            .split("\n")[0]));
```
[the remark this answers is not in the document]
ibacher: That makes sense

api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
```diff
@@ -202,5 +202,7 @@
 	
 	public static final String TAG_SEARCH_HANDLER = "tag.search.handler";
 	
-	public static final String HAPI_DEFAULT_NARRATIVES_PROPERTY_FILE = "classpath:ca/uhn/fhir/narrative/narratives.properties";
+	public static final String HAPI_NARRATIVES_PROPERTY_FILE = "classpath:ca/uhn/fhir/narrative/narratives.properties";
+	
+	public static final String OPENMRS_NARRATIVES_PROPERTY_FILE = "file:src/main/resources/narratives.properties";
```
ibacher, line 207: This should be "classpath:narrative.properties"

ShivanshRakesh: @ibacher If I do that, it uses the `narratives.properties` of HAPI, instead of the local one. So I initially kept it as `defaultNarratives.properties`, but later I changed it to `narratives.properties`. So I used the file path instead. So should I rename it back?

ShivanshRakesh:
I've limited this PR to the addition of narratives for the resources: `AllergyIntolerance`, `Condition` and `MedicationRequest`. Once it gets reviewed, approved and merged, I'll create another PR for the next set of 2 or 3 resources.

api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
ShivanshRakesh: @ibacher If I do that, it uses the narratives.properties of HAPI, instead of the local one. So I've moved the resources to a sub-directory `org/openmrs/module/fhir2/` inside the `resources` directory, and changed `OPENMRS_NARRATIVES_PROPERTY_FILE` to `classpath:org/openmrs/module/fhir2/narratives.properties`.

ibacher: Ok. My point was primarily that `file:` with `src/main/resources/...` is only going to work in a development environment. Instead of differentiating these two, it might make sense to just use something like `classpath*:narratives.properties`, allowing us to use other `narratives.properties` files that exist on the classpath. Basically, we want to be forced to know as little as possible about HAPI.

ShivanshRakesh: Ah, I see.

ShivanshRakesh:
Are there any further changes required in this PR?

cc @ibacher @HerbertYiga @suthagar23

HerbertYiga:
This looks good to me,@ibacher helo ian,do you think we can have this merged,so as we get focus on  FM2-208

ShivanshRakesh:
Thanks @HerbertYiga! :)

ShivanshRakesh:
Thanks for reviewing @HerbertYiga :)

ShivanshRakesh:
Thanks @ibacher! :)

PR #203, opened 2020-06-11
FM2-139: Add initial changes for the related person search.

Akayeshmantha:
Add search support for related person resources. 

## Issue I worked on
https://issues.openmrs.org/browse/FM2-139

- [ X] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

- [ X] I have **added tests** to cover my changes. (If you refactored

- [X ] I ran `mvn clean package` right before creating this pull request and

  No? -> execute above command

- [ X] All new and existing **tests passed**.

- [ X] My pull request is **based on the latest changes** of the master branch.

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDao.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDao.java
@@ -701,11 +701,20 @@ public abstract class BaseDao {
 	
 	protected void handleNames(Criteria criteria, StringAndListParam name, StringAndListParam given,
 	        StringAndListParam family) {
+		handleNames(criteria, name, given, family, null);
+	}
+	
+	protected void handleNames(Criteria criteria, StringAndListParam name, StringAndListParam given,
+	        StringAndListParam family, String personAlias) {
 		if (name == null && given == null && family == null) {
 			return;
 		}
 		
-		criteria.createAlias("names", "pn");
+		if (StringUtils.isNotBlank(personAlias)) {
+			criteria.createAlias(String.format("%s.names", personAlias), "pn");
+		} else {
+			criteria.createAlias("names", "pn");
+		}
 		
 		if (name != null) {
 			handleAndListParamAsStream(name,
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirRelatedPersonDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirRelatedPersonDaoImpl.java
@@ -9,8 +9,37 @@
  */
 package org.openmrs.module.fhir2.api.dao.impl;
 
+import static org.hibernate.criterion.Restrictions.and;
+import static org.hibernate.criterion.Restrictions.eq;
+import static org.hibernate.criterion.Restrictions.isNull;
+import static org.hibernate.criterion.Restrictions.or;
+import static org.hibernate.criterion.Restrictions.sqlRestriction;
+import static org.hl7.fhir.r4.model.Patient.SP_FAMILY;
+import static org.hl7.fhir.r4.model.Patient.SP_GIVEN;
+import static org.hl7.fhir.r4.model.Person.SP_ADDRESS_CITY;
+import static org.hl7.fhir.r4.model.Person.SP_ADDRESS_COUNTRY;
+import static org.hl7.fhir.r4.model.Person.SP_ADDRESS_POSTALCODE;
+import static org.hl7.fhir.r4.model.Person.SP_ADDRESS_STATE;
+import static org.hl7.fhir.r4.model.Person.SP_BIRTHDATE;
+import static org.hl7.fhir.r4.model.Person.SP_NAME;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+import ca.uhn.fhir.rest.api.SortSpec;
+import ca.uhn.fhir.rest.param.DateRangeParam;
+import ca.uhn.fhir.rest.param.StringAndListParam;
+import ca.uhn.fhir.rest.param.TokenAndListParam;
 import lombok.AccessLevel;
 import lombok.Setter;
+import org.hibernate.Criteria;
+import org.hibernate.criterion.DetachedCriteria;
+import org.hibernate.criterion.Order;
+import org.hibernate.criterion.Projections;
+import org.hibernate.criterion.Subqueries;
+import org.hibernate.sql.JoinType;
+import org.openmrs.PersonName;
 import org.openmrs.Relationship;
 import org.openmrs.module.fhir2.api.dao.FhirRelatedPersonDao;
 import org.springframework.stereotype.Component;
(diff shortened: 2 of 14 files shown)


varung-31:
@Akayeshmantha Thanks for this. Could you also incorporate paging support for this ?

Akayeshmantha:
@varung-31 sure that will be the next step 🙂

varung-31:
> @varung-31 sure that will be the next step slightly_smiling_face
> First lets get this task closed

Agreed. But going forward, every resource will have paging implemented by the end of this week. It's better if you can incorporate paging within this ticket only.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseRelatedPersonDao.java
```diff
@@ -0,0 +1,127 @@
...
+import org.hibernate.criterion.DetachedCriteria;
+import org.hibernate.criterion.Order;
+import org.hibernate.criterion.Projections;
+import org.hibernate.criterion.Subqueries;
+import org.hibernate.sql.JoinType;
+import org.openmrs.Auditable;
+import org.openmrs.OpenmrsObject;
+import org.openmrs.PersonName;
+
+public abstract class BaseRelatedPersonDao<T extends OpenmrsObject & Auditable> extends BaseFhirDao<T> {
```
ibacher, line 40: Does this need to be in a separate class? The reason `BasePersonDao` is separated out is because there are a number of resources that all use `Person` and need to have similar logic, but the logic here seems pretty specific to `RelatedPerson`. Or is there something I'm missing?

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDao.java
```diff
@@ -652,12 +652,16 @@ protected void handleIdentifier(Criteria criteria, TokenAndListParam identifier)
...
 		if (name == null && given == null && family == null) {
 			return;
 		}
 		
-		criteria.createAlias("names", "pn");
+		if (isInnerQuery) {
+			criteria.createAlias("m.names", "pn");
+		} else {
+			criteria.createAlias("names", "pn");
+		}
```
ibacher, line 664: Instead of doing this, I think it would be better to do something like this:

```java
protected void handleNames(Criteria criteria, StringAndListParam name, StringAndListParam given, StringAndListParam family) {
    return handleNames(criteria, name, given, family, null);
}

protected void handleNames(Criteria criteria, StringAndListParam name, StringAndListParam given, StringAndListParam family, String personAlias) {
    if (name == null & given == null && family == null) {
        return;
    }

    if (StringUtils.isNotBlank(personAlias) {
        criteria.createAlias(String.format("%.names", personAlias), "pn");
    } else {
        criteria.createAlias("names", "pn");
    }
    
    ...
}
```

This doesn't break existing code and, to me, at any rate `personAlias` tells me what that parameter is and why it's important more than `isInnerQuery` does.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirRelatedPersonDaoImpl.java
```diff
@@ -9,14 +9,48 @@
...
+		
+		handleSort(criteria, sort);
+		
+		return criteria.list();
+	}
 	
+	@Override
+	protected String getSqlAlias() {
+		return "this_";
+	}
```
ibacher, line 55: I don't think we actually need this unless we're actually reusing the `BasePersonDao` class.

api/src/main/java/org/openmrs/module/fhir2/providers/r3/RelatedPersonFhirResourceProvider.java
```diff
@@ -11,17 +11,26 @@
 
 import javax.validation.constraints.NotNull;
 
+import ca.uhn.fhir.rest.annotation.*;
```
varung-31, line 14: We should remove star-imports

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseRelatedPersonDao.java
Akayeshmantha: Hi @ibacher  I think yes since the relationship is not directly person. So inside relation yes we have personA which is the related person in fhir. The queries are a bit different as you can see.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDao.java
Akayeshmantha: of course, it's a great way. Sorry for this.

api/src/main/java/org/openmrs/module/fhir2/providers/r3/RelatedPersonFhirResourceProvider.java
Akayeshmantha: that's a good catch yes this is bad

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseRelatedPersonDao.java
Akayeshmantha: Hi @ibacher  I think atleast we have to get the use of BasePersonDao class. Since we need to have all the personrelated sort stuff in the relatedperson as well.

but anyway if we are not maintaining a sperate class I think its better to alter the 
handleSort in the base dao to include a person alisa parameter as we did for the name search/ 
wdyt?

Akayeshmantha: @ibacher  waiting for youre response on this

Akayeshmantha:
@ibacher  @varung-31 I will start working on the paging till @ibacher  gets back

varung-31:
> @ibacher @varung-31 I will start working on the paging till @ibacher gets back.
> I hope we need to have a separate issue for that if am right.

@Akayeshmantha  It's fine. I have taken up the issue to implement paging for this.

Akayeshmantha:
Oh ok will for another issue then

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseRelatedPersonDao.java
ibacher: I just meant "do we need a `BaseRelatedPersonDao` class or can we move all of this into `RelatedPersonDao`?" We aren't really reusing `BasePersonDao` here at all.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirRelatedPersonDaoImpl.java
ibacher: Basically, this can be hard-coded where needed

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseRelatedPersonDao.java
Akayeshmantha: oh yes sorry about this @ibacher

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirRelatedPersonDaoImpl.java
Akayeshmantha: true thanks I fixed the mistake

Akayeshmantha:
@ibacher  pushed changes thanks

varung-31:
@Akayeshmantha  I think a better way to do this is to extend the BasePersonDao class and override the paramToProps method. This way we'll be able to take advantage of the paramToProp and handleAdresses methods in BasePersonDao as well.

ibacher:
@varung-31 That's actually probably a bad idea because of how relationships are stored in OpenMRS. In particular, we may need to be able to search a relationship on either of the two related people, which means it will need special handling above and beyond what we can reasonably support in `BasePersonDao`.

Akayeshmantha:
@varung-31  BasePersonDao also overrides the BaseDao paramstoprop and does implements for a specific use case. And the relationship is not a person in openmrs we have to dig another level in relationship to achieve the functionality of a person. So it won't be valid in the way you proposed. If you look at the queries you may understand what is the diff is.

ibacher:
@Akayeshmantha Can you remind me: does this search the properties of PersonA or PersonB?

Akayeshmantha:
It's personA as from what we have in translator

PR #207, opened 2020-06-13
FM2-207: Create default narratives for Encounter, Observation and Location

ShivanshRakesh:
Added OpenMRS Narratives for the following resources:
* Encounter
* Observation
* Location

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-207


ShivanshRakesh:
I've added narrative templates for the following:
* Encounter
* Location
* Observation
* Generic datatypes which have been used in the above (eg. SampleData, etc.)

[PR#198](https://github.com/openmrs/openmrs-module-fhir2/pull/198) should be merged before this PR. Please refer to [this commit](https://github.com/openmrs/openmrs-module-fhir2/pull/207/commits/1cb2b4b0240ebab0a043c2640566a1765904b8cf) for the changes specifically intended for this PR.

HerbertYiga:
This looks good to me,@ibacher  hei hi ian,how does this look to you,we could get it merged so as @ShivanshRakesh continues with FM2-208

ShivanshRakesh:
Thanks for reviewing @HerbertYiga :)

ibacher:
@ShivanshRakesh Since this doesn't build off your previous commit, there's now a conflict. Could you fix the conflict please?

ShivanshRakesh:
@ibacher I've made the changes. PTAL!

ShivanshRakesh:
@ibacher I've made the changes in accordance with [PR#225](https://github.com/openmrs/openmrs-module-fhir2/pull/225) (for [FM2-243](https://issues.openmrs.org/browse/FM2-243)). PTAL.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirRelatedPersonDaoImpl.java
```diff
@@ -90,18 +90,15 @@
...
-							    .add(eq("pn2.preferred", true))
+			    and(Subqueries.notExists(DetachedCriteria.forClass(PersonName.class, "pn2").add(eq("pn2.preferred", true))
 			            // WARNING this is fragile
 			            .add(sqlRestriction(String.format("pn2_.person_id = %s.person_a", "this_")))
 			            .setProjection(Projections.id())),
 			        Subqueries.propertyEq("pn.personNameId",
-			            DetachedCriteria.forClass(PersonName.class, "pn3")
-					            .add(eq("pn3.preferred", false))
+			            DetachedCriteria.forClass(PersonName.class, "pn3").add(eq("pn3.preferred", false))
 			                    // WARNING this is fragile
```
ShivanshRakesh, line 102: I guess this file got changed due to autoformat.

ibacher:
@ShivanshRakesh Why are we adding so many subtables to this? Is there a way in which we can get away with, say, a single table for all the properties? We seem to have a table-per-row, which seems to me to be counter-productive. Here for example:

Screenshot 2020-06-26 at 09 34 45

We can see that the properties look mis-aligned because of the table-per-row effect. The header for the Participants table is also a little less clear than it could be.

I'd also think we'd want to display the coding details where possible. For example, the default narrative for that same resource looks like this:

Screenshot 2020-06-26 at 09 42 36

I'm not saying we need to duplicate that view directly, but, for example, from our narrative, I have no idea which of those values are coded responses and which are free text whereas with the default rendering, I can clearly see which are coded responses and the corresponding codes for them.

ShivanshRakesh:
> @ShivanshRakesh Why are we adding so many subtables to this? Is there a way in which we can get away with, say, a single table for all the properties? We seem to have a table-per-row, which seems to me to be counter-productive. Here for example:

@ibacher Actually there's just one table for most of the properties. But, for the properties which can have multiple values and have multiple sub-properties, I've used a subPropertyTable, trying to keep it the way it is on the FHIR site. So for this, I had to split the main table in between to maintain the order. Maybe we could keep all such subPropertyTables in the end, so that the main property table does not get disturbed. Will that be fine?

> I'd also think we'd want to display the coding details where possible. For example, the default narrative for that same resource looks like this:

Okay. Will add the coding details for `Coding` type. Since that involves changing the narrative for `Coding` type alone, should I rather create a separate PR for that?

ibacher:
> Okay. Will add the coding details for Coding type. Since that involves changing the narrative for Coding type alone, should I rather create a separate PR for that?

Yes that's fine.

> But, for the properties which can have multiple values and have multiple sub-properties, I've used a subPropertyTable, trying to keep it the way it is on the FHIR site. So for this, I had to split the main table in between to maintain the order.

Why not just make these subproperty tables nested tables of a single main table? E.g., something like this:

```html
<table class="hapiPropertyTable">
    <tbody>
        <tr th:if="${resource.id}">
            <td>Id:</td>
            <td th:text="${resource.id}"></td>
        </tr>
        <tr th:if="${not resource.identifier.empty}">
            <td>Identifier:</td>
            <td>
                <th:block th:each="identifier: ${resource.identifier}">
                    <div th:narrative="${identifier}"></div>
                </th:block>
            </td>
        </tr>
        <tr th:if="${resource.status}">
            <td>Status:</td>
            <td th:text="${resource.status}"></td>
        </tr>

        <th:block th:if="${not resource.statusHistory.empty}">
            <tr>
                <td>Status History:</td>
                <td>
                    <table class="subPropertyTable">
                        <tbody>
                            <tr>
                                <th>#</th>
                                <th>Status</th>
                                <th>Period</th>
                            </tr>
                            <tr th:each="statusHistory, statusHistoryItr: ${resource.statusHistory}">
                                <td th:text="${statusHistoryItr.count"></td>
                                <td th:narrative="${statusHistory.status"></td>
                                <td th:narrative="${statusHistory.period"></td>
                            </tr>
                        </tbody>
                    </table>
                </td>
            </tr>
        </th:block>
    </tbody>
</table>
```

The alternative is to get away from tables and move to something that more closely resembles what's in the FHIR spec examples, where the `<h2>` headers make a bit more sense.

ShivanshRakesh:
@ibacher Yeah, this would be better.
Also, since it is something which has to be done for all the resources, I guess it would be better to create a separate PR for this, working on top of commits in [PR#198](https://github.com/openmrs/openmrs-module-fhir2/pull/198), [PR#207](https://github.com/openmrs/openmrs-module-fhir2/pull/207) (this), [PR#219](https://github.com/openmrs/openmrs-module-fhir2/pull/219) and [PR#232](https://github.com/openmrs/openmrs-module-fhir2/pull/232), since if I make changes in this one, the other two in line will have to be rebased again. Will that be fine?

ShivanshRakesh:
@ibacher @HerbertYiga @suthagar23 Are there any changes, other than those suggested by @ibacher, required in the PRs PR#207 (this PR), [PR#219](https://github.com/openmrs/openmrs-module-fhir2/pull/219) and [PR#232](https://github.com/openmrs/openmrs-module-fhir2/pull/232)?

If these could get merged, I'd be able to work on the changes suggested by @ibacher, since making changes in narrative of `Coding` and the changes in table layout, will require all the expected narratives to be changed too, for which the new commits need to be on top of all these commits.

ShivanshRakesh:
Thanks @ibacher :)

PR #214, opened 2020-06-16
FM2-225:Add buildDelete static method for FhirProviderUtils

sherrif10:
https://issues.openmrs.org/browse/FM2-225

cc @ibacher  @corneliouzbett .kindly review.

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/providers/util/FhirProviderUtils.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/providers/util/FhirProviderUtils.java
@@ -15,7 +15,9 @@ import ca.uhn.fhir.rest.api.MethodOutcome;
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
 import org.hl7.fhir.r4.model.Bundle;
+import org.hl7.fhir.r4.model.CodeableConcept;
 import org.hl7.fhir.r4.model.DomainResource;
+import org.hl7.fhir.r4.model.OperationOutcome;
 import org.hl7.fhir.r4.model.Resource;
 
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
@@ -31,6 +33,22 @@ public class FhirProviderUtils {
 		MethodOutcome methodOutcome = new MethodOutcome();
 		methodOutcome.setCreated(true);
 		return buildWithResource(methodOutcome, resource);
+	}
+	
+	public static OperationOutcome buildDelete(DomainResource resource) {
+		
+		final CodeableConcept MSG_DELETED = new CodeableConcept();
+		{
+			MSG_DELETED.addCoding().setSystem("http://terminology.hl7.org/CodeSystem/operation-outcome")
+			        .setCode("MSG_DELETED").setDisplay("This resource has been deleted");
+			MSG_DELETED.setText("This resource has been deleted");
+		}
+		
+		OperationOutcome outcome = new OperationOutcome();
+		outcome.addIssue().setSeverity(OperationOutcome.IssueSeverity.INFORMATION)
+		        .setCode(OperationOutcome.IssueType.INFORMATIONAL).setDetails(MSG_DELETED);
+		return outcome;
+		
 	}
 	
 	private static MethodOutcome buildWithResource(MethodOutcome methodOutcome, DomainResource resource) {


sherrif10:
Hello @ibacher  i wish to test this type of code an see its effects

api/src/main/java/org/openmrs/module/fhir2/providers/util/FhirProviderUtils.java
```diff
@@ -33,6 +36,22 @@ public static MethodOutcome buildCreate(DomainResource resource) {
...
+	public static OperationOutcome buildDelete(OperationOutcome.IssueSeverity code, OperationOutcome.IssueType issueType,
+	        BaseServerResponseException exception) {
+		//created codeable concepts to be called by operationOutcome, codeableconcept only limited to two properties text and coding
+		CodeableConcept codeableConcept = new CodeableConcept().setText("MSG_DELETED").setText("informational")
+		        .setText("This resource has been deleted").setText("informational")
+		        .setText("http://terminology.hl7.org/CodeSystem/operation-outcome").setText("This resource has been deleted")
+		        .setText(exception.getMessage());
+		
+		OperationOutcome operationOutcome = new OperationOutcome();
+		operationOutcome.addIssue().setSeverity(OperationOutcome.IssueSeverity.ERROR).setCode(issueType)
```
corneliouzbett, line 48: Depending on the server outcome, IssueSeverity can be either `ERROR, FATAL, WARNING `or `INFORMATION`. So this should be dynamic.
More information about OperationOutcome [here](https://www.hl7.org/fhir/operationoutcome.html)

sherrif10: Thanks for clarifying about this, Anything to respond about my code above

sherrif10, line 48: i set issueSeverity.ERROR  as it should be supposed to be

ibacher: @sherrif10 Actually, this should be `IssueSeverity.INFORMATION, because we're sending a result that says the object was successfully deleted. No error occurred. I've made that change.

api/src/main/java/org/openmrs/module/fhir2/providers/util/FhirProviderUtils.java
```diff
@@ -33,6 +36,21 @@ public static MethodOutcome buildCreate(DomainResource resource) {
...
+		        .setText("This resource has been deleted").setText("informational")
+		        .setText("http://terminology.hl7.org/CodeSystem/operation-outcome").setText("This resource has been deleted")
+		        .setText(exception.getMessage());
+		
+		OperationOutcome operationOutcome = new OperationOutcome();
+		operationOutcome.addIssue().setSeverity(OperationOutcome.IssueSeverity.INFORMATION).setCode(issueType)
+		        .setDetails(codeableConcept);// am assumming setting details will trigger only called codeableconcepts above             	
+		exception.setOperationOutcome(operationOutcome);
+		return operationOutcome;
+	}
```
ibacher, line 52:
```suggestion
	public static OperationOutcome buildDelete(DomainResource resource) {
		OperationOutcome outcome = new OperationOutcome();
		outcome.addIssue().setSeverity(OperationOutcome.IssueSeverity.INFORMATION)
				.setCode(OperationOutcome.IssueType.INFORMATIONAL).setDetails(MSG_DELETED);
		return outcome;
	}
```

With `MSG_DELETED` defined like this:

```java
private static final CodeableConcept MSG_DELETED = new CodeableConcept();

static {
	MSG_DELETED.addCoding().setSystem("http://terminology.hl7.org/CodeSystem/operation-outcome")
			.setCode("MSG_DELETED").setDisplay("This resource has been deleted");
	MSG_DELETED.setText("This resource has been deleted");
}
```

api/src/main/java/org/openmrs/module/fhir2/providers/util/FhirProviderUtils.java
sherrif10: Thanks for replying @ibacher , Wont we need to call these errors as variables, i mean something like publi static string INFORMATION="A message that the server will send";. this is my idea however, or we just define the extension of issueSeverity and the server knows the what to send!

api/src/main/java/org/openmrs/module/fhir2/providers/util/FhirProviderUtils.java
sherrif10: I  didnot add private access modifier since it triggered me to make MSG_DELETED  defined as a constant

PR #228, opened 2020-06-23
FM2-224:Add delete method for Task

sherrif10:
https://issues.openmrs.org/browse/FM2-224

@ibacher  @corneliouzbett , am not getting this method whether this is the right way of how its going to be implmented. Using a method suggested, encounters that we need to have a service class variable. hence am not seeing how fhirProviderUtils is used as service except if its abstracted. More help here will be gladly welcome


corneliouzbett:
@sherrif10 Actually what you trying to do here has been done #214.  For this ticket, you are required to add `DELETE` to Task resource providers for both [R3](https://github.com/openmrs/openmrs-module-fhir2/blob/master/api/src/main/java/org/openmrs/module/fhir2/providers/r3/TaskFhirResourceProvider.java) and[R4](https://github.com/openmrs/openmrs-module-fhir2/blob/master/api/src/main/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceProvider.java)

```java
@Delete
public OperationOutcome deleteTask(@IdParam IdType id) {
    Task task = taskService.delete(id.getIdPart());
	if (task == null){
	    throw new ResourceNotFoundException("Could not find task resource with id " + id.getIdPart() + "to delete");	
	}
	return FhirProviderUtils.buildDelete(task);
}
```

PR #252, opened 2020-07-09
FM2-251: Add support for overriding default narratives

ShivanshRakesh:

Added support for overriding default narratives, with the narrative override property file location to be stored as the global property `narrativesOverridePropertyFile`.

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-251

### DIFF
--- a/omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
+++ b/omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
@@ -62,8 +62,15 @@ public class FhirRestServlet extends RestfulServer {
 		setDefaultResponseEncoding(EncodingEnum.JSON);
 		registerInterceptor(loggingInterceptor);
 		
-		getFhirContext().setNarrativeGenerator(new CustomThymeleafNarrativeGenerator(
-		        FhirConstants.HAPI_NARRATIVES_PROPERTY_FILE, FhirConstants.OPENMRS_NARRATIVES_PROPERTY_FILE));
+		String narrativeOverridePropertyFile = globalPropertyService
+		        .getGlobalProperty(FhirConstants.NARRATIVES_OVERRIDE_PROPERTY_FILE, "");
+		if (narrativeOverridePropertyFile.isEmpty()) {
+			getFhirContext().setNarrativeGenerator(new CustomThymeleafNarrativeGenerator(
+			        FhirConstants.HAPI_NARRATIVES_PROPERTY_FILE, FhirConstants.OPENMRS_NARRATIVES_PROPERTY_FILE));
+		} else {
+			getFhirContext().setNarrativeGenerator(new CustomThymeleafNarrativeGenerator(narrativeOverridePropertyFile,
+			        FhirConstants.HAPI_NARRATIVES_PROPERTY_FILE, FhirConstants.OPENMRS_NARRATIVES_PROPERTY_FILE));
+		}
 	}
 	
 	@Override
--- a/omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
+++ b/omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
@@ -9,7 +9,10 @@
  */
 package org.openmrs.module.fhir2.web.servlet;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
 
 import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.narrative.CustomThymeleafNarrativeGenerator;
@@ -62,15 +65,12 @@ public class FhirRestServlet extends RestfulServer {
 		setDefaultResponseEncoding(EncodingEnum.JSON);
 		registerInterceptor(loggingInterceptor);
 		
-		String narrativeOverridePropertyFile = globalPropertyService
-		        .getGlobalProperty(FhirConstants.NARRATIVES_OVERRIDE_PROPERTY_FILE, "");
-		if (narrativeOverridePropertyFile.isEmpty()) {
-			getFhirContext().setNarrativeGenerator(new CustomThymeleafNarrativeGenerator(
-			        FhirConstants.HAPI_NARRATIVES_PROPERTY_FILE, FhirConstants.OPENMRS_NARRATIVES_PROPERTY_FILE));
-		} else {
-			getFhirContext().setNarrativeGenerator(new CustomThymeleafNarrativeGenerator(narrativeOverridePropertyFile,
-			        FhirConstants.HAPI_NARRATIVES_PROPERTY_FILE, FhirConstants.OPENMRS_NARRATIVES_PROPERTY_FILE));
-		}
+		List<String> propertyFiles = new ArrayList<>(
+		        Arrays.asList(globalPropertyService.getGlobalProperty(FhirConstants.NARRATIVES_OVERRIDE_PROPERTY_FILE, ""),
+		            FhirConstants.HAPI_NARRATIVES_PROPERTY_FILE, FhirConstants.OPENMRS_NARRATIVES_PROPERTY_FILE));
+		propertyFiles.removeAll(Arrays.asList("", null));
+		getFhirContext().setNarrativeGenerator(
+		    new CustomThymeleafNarrativeGenerator(propertyFiles.toArray(new String[propertyFiles.size()])));
 	}
 	
 	@Override
--- a/omod/src/main/resources/config.xml
+++ b/omod/src/main/resources/config.xml
@@ -128,5 +128,10 @@
 		<description>Set provider attribute type uuid</description>
 	</globalProperty>
 
-</module>
+	<globalProperty>
+		<property>${project.parent.artifactId}.narrativesOverridePropertyFile</property>
+		<defaultValue></defaultValue>
+		<description>Absolute path of narrative override properties file (prefixed with "file:")</description>
+	</globalProperty>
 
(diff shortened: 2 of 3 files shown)


ShivanshRakesh:

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
```diff
@@ -62,8 +62,15 @@ protected void initialize() {
 		setDefaultResponseEncoding(EncodingEnum.JSON);
 		registerInterceptor(loggingInterceptor);
 		
-		getFhirContext().setNarrativeGenerator(new CustomThymeleafNarrativeGenerator(
-		        FhirConstants.HAPI_NARRATIVES_PROPERTY_FILE, FhirConstants.OPENMRS_NARRATIVES_PROPERTY_FILE));
+		String narrativeOverridePropertyFile = globalPropertyService
+		        .getGlobalProperty(FhirConstants.NARRATIVES_OVERRIDE_PROPERTY_FILE, "");
+		if (narrativeOverridePropertyFile.isEmpty()) {
```
HerbertYiga, line 67: This looks good to  me cc @ibacher

omod/src/main/resources/config.xml
```diff
@@ -23,10 +23,10 @@
 	</description>
 
 	<activator>${project.parent.groupId}.${project.parent.artifactId}.FhirActivator</activator>
-	
+
 	<!-- <updateURL>https://modules.openmrs.org/modules/download/@MODULE_ID@/update.rdf</updateURL> -->
 	<!-- /Base Module Properties -->
-	
+
```
HerbertYiga, line 29: @ShivanshRakesh  we came out with a few unwanted code formats, mvn clean package could always help you clean up stuff before pushing

ShivanshRakesh: @HerbertYiga I actually did `mvn clean package`, don't know why this happened. I'll push again to remove these unwanted changes.

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
ShivanshRakesh: Thanks @HerbertYiga :)

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
```diff
@@ -62,8 +65,12 @@ protected void initialize() {
...
 		registerInterceptor(loggingInterceptor);
 		
-		getFhirContext().setNarrativeGenerator(new CustomThymeleafNarrativeGenerator(
-		        FhirConstants.HAPI_NARRATIVES_PROPERTY_FILE, FhirConstants.OPENMRS_NARRATIVES_PROPERTY_FILE));
+		List<String> propertyFiles = new ArrayList<>(
+		        Arrays.asList(globalPropertyService.getGlobalProperty(FhirConstants.NARRATIVES_OVERRIDE_PROPERTY_FILE, ""),
+		            FhirConstants.HAPI_NARRATIVES_PROPERTY_FILE, FhirConstants.OPENMRS_NARRATIVES_PROPERTY_FILE));
+		propertyFiles.removeAll(Arrays.asList("", null));
+		getFhirContext().setNarrativeGenerator(
+		    new CustomThymeleafNarrativeGenerator(propertyFiles.toArray(new String[propertyFiles.size()])));
```
ShivanshRakesh, line 73: Refactored it this way since the [previous commit](https://github.com/openmrs/openmrs-module-fhir2/pull/252/commits/2204e655ba3a9aeae7f05855281c982923b49969) didn't had full coverage.

omod/src/main/resources/config.xml
```diff
@@ -128,5 +128,10 @@
 		<description>Set provider attribute type uuid</description>
 	</globalProperty>
 
-</module>
+	<globalProperty>
+		<property>${project.parent.artifactId}.narrativesOverridePropertyFile</property>
+		<defaultValue></defaultValue>
+		<description>Absolute path of narrative override properties file (prefixed with "file:")</description>
```
ibacher, line 134: Is this requiring that it be prefixed with `file:`? I would prefer to do something like:

Does it start with `classpath:`? Use the `classpath`. Otherwise if it doesn't start with `file:` add file. We should also be careful when loading a file from a user-specified directory. We should check that it's in the OpenMRS Application Data directory, ends in `.properties`, etc.

ShivanshRakesh: Yeah, actually it does support the `classpath:` and the `file:` prefix, since even this uses the `NarrativeTemplateManifest.loadResource()` method. I'll add it in the description too.

Okay, I'll add some validation for `file:` prefixed paths while loading the file.

ShivanshRakesh: @ibacher Does the properties file have to necessarily be in the OpenMRS Application Data directory?

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
```diff
@@ -62,8 +65,12 @@ protected void initialize() {
 		setDefaultResponseEncoding(EncodingEnum.JSON);
 		registerInterceptor(loggingInterceptor);
 		
-		getFhirContext().setNarrativeGenerator(new CustomThymeleafNarrativeGenerator(
-		        FhirConstants.HAPI_NARRATIVES_PROPERTY_FILE, FhirConstants.OPENMRS_NARRATIVES_PROPERTY_FILE));
+		List<String> propertyFiles = new ArrayList<>(
+		        Arrays.asList(globalPropertyService.getGlobalProperty(FhirConstants.NARRATIVES_OVERRIDE_PROPERTY_FILE, ""),
+		            FhirConstants.HAPI_NARRATIVES_PROPERTY_FILE, FhirConstants.OPENMRS_NARRATIVES_PROPERTY_FILE));
+		propertyFiles.removeAll(Arrays.asList("", null));
```
ibacher, line 71: Sometimes, it's better to do things the more straight-forward way.

```suggestion
		String narrativeOverridePropertyFile = globalPropertyService.getGlobalProperty(FhirConstants.NARRATIVES_OVERRIDE_PROPERTY_FILE);
		
		String[] narrativePropertiesFiles;
		if (narrativeOverridePropertyFile != null) {
		    narrativePropertiesFiles = new String[] { FhirConstants.HAPI_NARRATIVE_PROPERTY_FILE, FhirConstants.OPENMRS_NARRATIVES_PROPERTY_FILE, narrativeOverridePropertyFile };
		} else {
			narrativePropertiesFiles = new String[] { FhirConstants.HAPI_NARRATIVE_PROPERTY_FILE, FhirConstants.OPENMRS_NARRATIVES_PROPERTY_FILE }
		}
```

This results in fewer throw-away objects.

omod/src/main/resources/config.xml
ibacher: I suppose that's less of a concern really. We do want make sure that users can supply something like `/opt/OpenMRS/overrides/mynarratives.properties`, however, without needing to prefix it with `file:`.

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
ibacher: Also, I'm unclear on this, but is the order in which properties files are loaded significant (I would guess so) and if so, which order is correct? Naïvely, I would think that you want the base file specified before the override file(s), but I haven't checked the implementation.

ShivanshRakesh: I actually did it this way initially, but then changed it to maintain coverage.

ShivanshRakesh: > Also, I'm unclear on this, but is the order in which properties files are loaded significant (I would guess so) and if so, which order is correct?

The override file has to be specified first to give it more priority. Yeah, it is kind of counter-intuitive, but this is the order it follows.

omod/src/main/resources/config.xml
ShivanshRakesh: Oh okay. But then we'll be needing to append the `file:` prefix since it also uses `NarrativeTemplateManifest.loadResource()` method to load the properties file, which further needs the `file:` or `classpath:` prefix. (Later when we add support for `openmrs:` we would be able to use it for loading properties file too)

ShivanshRakesh: @ibacher Also, while validating the narrative override properties filepath, in case of some error, should we raise a ValidationError or just log the error and continue without overriding the properties?

ShivanshRakesh: > We should also be careful when loading a file from a user-specified directory. We should check that it's in the OpenMRS Application Data directory, ends in .properties, etc.

@ibacher I was thinking, do we actually need to validate it on our end? Because, validation for existence of the file is already being done in the [`NarrativeTemplateManifest.loadResource()`](https://github.com/jamesagnew/hapi-fhir/blob/29bf6b8171f48f14ec2c44c997122c3917444942/hapi-fhir-base/src/main/java/ca/uhn/fhir/narrative2/NarrativeTemplateManifest.java#L184) method. We could add a validation for `.properties` extension if is needed though.

ShivanshRakesh:
I have pushed a commit to `FM2-251` branch of my fork but it is not getting reflected here. I guess it is happening because of squashing the commits and force pushing. I'll create a new PR for this then.

cc @ibacher @HerbertYiga @suthagar23

PR #255, opened 2020-07-13
FM2-252: Add support for relative path with prefix 'openmrs:' for narrative templates

ShivanshRakesh:

Added support for relative paths prefixed with `openmrs:` for location of narratives and properties file. 
For example, the relative path `openmrs:configuration/Patient.html` would point to the location `<OPENMRS_APPLICATION_DATA_DIRECTORY>/configuration/Patient.html`.

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-252

### DIFF
--- a/omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenMRSNarrativeTemplate.java
+++ b/omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenMRSNarrativeTemplate.java
@@ -1,0 +1,124 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+
+/*
+ * This class is derived and modified from the HAPI FHIR Core Library under the
+ * terms of the Apache License, Version 2.0. You may obtain a copy of the
+ * License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * The portions of this class that have not been modified by OpenMRS are
+ * Copyright (C) 2014 - 2020 University Health Network.
+ */
+
+package org.openmrs.module.fhir2.narrative;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+import ca.uhn.fhir.narrative2.INarrativeTemplate;
+import ca.uhn.fhir.narrative2.TemplateTypeEnum;
+import ca.uhn.fhir.rest.server.exceptions.InternalErrorException;
+import org.hl7.fhir.instance.model.api.IBase;
+
+public class OpenMRSNarrativeTemplate implements INarrativeTemplate {
+	
+	private String contextPath;
+	
+	private Set<String> appliesToProfiles = new HashSet<>();
+	
+	private Set<String> appliesToResourceTypes = new HashSet<>();
+	
+	private Set<Class<? extends IBase>> appliesToResourceClasses = new HashSet<>();
+	
+	private TemplateTypeEnum templateType = TemplateTypeEnum.THYMELEAF;
+	
+	private String templateName;
+	
+	private String templateFilename;
+	
+	private Set<String> appliesToDatatypes = new HashSet<>();
+	
+	@Override
+	public String getContextPath() {
+		return this.contextPath;
+	}
+	
+	protected void setContextPath(String contextPath) {
+		this.contextPath = contextPath;
+	}
+	
+	@Override
+	public Set<String> getAppliesToProfiles() {
+		return Collections.unmodifiableSet(this.appliesToProfiles);
+	}
+	
+	protected void addAppliesToProfile(String profile) {
+		this.appliesToProfiles.add(profile);
+	}
+	
+	@Override
+	public Set<String> getAppliesToResourceTypes() {
+		return Collections.unmodifiableSet(this.appliesToResourceTypes);
+	}
+	
+	protected void addAppliesToResourceType(String resourceType) {
+		this.appliesToResourceTypes.add(resourceType);
+	}
+	
+	@Override
+	public Set<Class<? extends IBase>> getAppliesToResourceClasses() {
+		return Collections.unmodifiableSet(this.appliesToResourceClasses);
+	}
+	
+	@Override
+	public TemplateTypeEnum getTemplateType() {
+		return this.templateType;
+	}
+	
+	protected void setTemplateType(TemplateTypeEnum templateType) {
+		this.templateType = templateType;
+	}
+	
+	@Override
+	public String getTemplateName() {
+		return this.templateName;
+	}
+	
+	protected OpenMRSNarrativeTemplate setTemplateName(String templateName) {
+		this.templateName = templateName;
+		return this;
+	}
+	
+	@Override
+	public String getTemplateText() {
+		try {
+			return OpenMRSNarrativeTemplateManifest.loadResource(this.templateFilename);
+		}
+		catch (IOException e) {
+			throw new InternalErrorException(e);
+		}
+	}
+	
+	protected void setTemplateFileName(String templateFileName) {
+		this.templateFilename = templateFileName;
+	}
+	
+	public Set<String> getAppliesToDataTypes() {
+		return Collections.unmodifiableSet(this.appliesToDatatypes);
+	}
+	
+	protected void addAppliesToDatatype(String dataType) {
+		this.appliesToDatatypes.add(dataType);
+	}
+}
--- a/omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenMRSNarrativeTemplateManifest.java
+++ b/omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenMRSNarrativeTemplateManifest.java
@@ -1,0 +1,231 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+
+/*
+ * This class is derived and modified from the HAPI FHIR Core Library under the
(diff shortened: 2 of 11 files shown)


ShivanshRakesh:

omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenMRSNarrativeTemplateManifest.java
```diff
@@ -0,0 +1,243 @@
...
+			}
+		} else if (name.startsWith("file:")) {
+			File file = new File(name.substring("file:".length()));
+			if (file.exists() == false) {
+				throw new IOException("File not found: " + file.getAbsolutePath());
+			}
+			try (FileInputStream inputStream = new FileInputStream(file)) {
+				return IOUtils.toString(inputStream, Charsets.UTF_8);
+			}
+		} else if (name.startsWith("openmrs:")) {
```
ShivanshRakesh, line 219: Here is the main modification which has been done to support `openmrs:` prefixed relative paths.

omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenMRSNarrativeTemplate.java
```diff
@@ -0,0 +1,120 @@
...
+	
+	OpenMRSNarrativeTemplate setTemplateName(String theTemplateName) {
+		myTemplateName = theTemplateName;
+		return this;
+	}
+	
+	@Override
+	public String getTemplateText() {
+		try {
+			return OpenMRSNarrativeTemplateManifest.loadResource(getTemplateFileName());
```
ShivanshRakesh, line 110: The modified `loadResource()` method has been called here.

omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenMRSNarrativeTemplate.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Set;
+
+import ca.uhn.fhir.narrative2.INarrativeTemplate;
+import ca.uhn.fhir.narrative2.TemplateTypeEnum;
+import ca.uhn.fhir.rest.server.exceptions.InternalErrorException;
+import org.hl7.fhir.instance.model.api.IBase;
+
+public class OpenMRSNarrativeTemplate implements INarrativeTemplate {
+	
+	private String CONTEXT_PATH;
```
ibacher, line 24: Definitely don't use ALL_CAPS for instance names. They are used for constants, which are usually `static final`.

omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenMRSNarrativeTemplate.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+import ca.uhn.fhir.narrative2.INarrativeTemplate;
+import ca.uhn.fhir.narrative2.TemplateTypeEnum;
+import ca.uhn.fhir.rest.server.exceptions.InternalErrorException;
+import org.hl7.fhir.instance.model.api.IBase;
+
+public class OpenMRSNarrativeTemplate implements INarrativeTemplate {
```
ibacher, line 22: So my main question here is why we have this whole implementation when it seems like we could get away with this:

```java
public class OpenMRSNarrativeTemplate extends ca.uhn.fhir.narrative2.NarrativeTemplate {
    @Override
	public String getTemplateText() {
		try {
			return OpenMRSNarrativeTemplateManifest.loadResource(getTemplateFileName());
		} catch (IOException e) {
			throw new InternalErrorException(e);
		}
	}
}
```

omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenMRSNarrativeTemplateManifest.java
```diff
@@ -0,0 +1,243 @@
...
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.Validate;
+import org.hl7.fhir.instance.model.api.IBase;
+import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.openmrs.util.OpenmrsUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class OpenMRSNarrativeTemplateManifest implements INarrativeTemplateManifest {
```
ibacher, line 46: Couldn't this just be something like:

```java
public class OpenMRSNarrativeTemplateManifest extends ca.uhn.fhir.narrative2.NarrativeTemplateManifest {
	public static NarrativeTemplateManifest forManifestFileLocation(String... propertyFiles) throws IOException {
		return forManifestFileLocation(Arrays.asList(propertyFiles));
	}

	public static NarrativeTemplateManifest forManifestFileLocation(Collection<String> propertyFiles) throws IOException {
		log.debug("Loading narrative properties file(s): {}", propertyFiles);
		return new NarrativeTemplateManifest(propertyFiles.stream().map(r -> loadResource(r)).map(p -> NarrativeTemplateManifest.loadProperties(p))
			.flatMap(p -> p.stream()).map(t -> new OpenmrsNarrativeTemplate(t)).collect(Collectors.toList()));
	}

	static String loadResource(String name) throws IOException {
		if (name.startsWith("openmrs:")) {
			File file = new File(OpenmrsUtil.getApplicationDataDirectory(), name.substring("openmrs:".length()));

			if (file.exists() == false) {
				throw new IOException("File not found: " + file.getAbsolutePath());
			}

			try (FileInputStream inputStream = new FileInputStream(file)) {
				return IOUtils.toString(inputStream, Charsets.UTF_8);
			}
		} else {
			return NarrativeTemplateManifest.loadResource(name);
		}
	}
}
```

(The above adds a small dependency that we create a constructor for `OpenmrsNarrativeTemplate` that takes a `NarrativeTemplate`, but that isn't too hard. That way we cut the code down to what we minimally need to support this change.

ShivanshRakesh: Yeah actually we could do this. I didn't think of it this way initially because of the constructor. But yeah, this looks much better.

omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenMRSNarrativeTemplate.java
ShivanshRakesh: Actually doing this resulted in errors in `loadProperties()` method of `NarrativeTemplateManifest`, as it calls non-public methods of `NarrativeTemplate` like `setTemplateName()`, etc.

omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenMRSNarrativeTemplateManifest.java
ShivanshRakesh: @ibacher I just forgot about it, but `NarrativeTemplateManifest` cannot be extended because of a private constructor right?

omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenMRSNarrativeTemplate.java
ibacher: I think that's what I'd like to see here. The idea is if we can make our implementation as light as possible, we're able to evolve with HAPI as easily as possible.

omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenMRSNarrativeTemplateManifest.java
ibacher: Yes... that's a good point. And I guess there isn't an easy way to remove a bunch of duplication here.

ShivanshRakesh: @ibacher Hmm.. and since some methods of `NarrativeTemplateManifest` call some private methods of `NarrativeTemplate` class, we can't reduce that either. So what should be done instead?

ibacher: Are you sure there are methods in `NarrativeTemplateManifest` that call *private* methods of `NarrativeTemplate`? Because, by definition, that's not possible... If it's just a matter of package-private methods, we can always stick our version of `OpenMRSNarrativeTemplateManifest` in the `ca.uhn.fhir.narrative2` package.

The alternative is to dress these classes up so that they more closely resemble OpenMRS coding conventions: Setters should be `public` or at least `protected`. We should drop `the` and `my` as prefixes to variables, etc. I'd also like to minimize duplicating code as much as possible.

ShivanshRakesh: > Are you sure there are methods in `NarrativeTemplateManifest` that call _private_ methods of `NarrativeTemplate`?

Ah, my bad. I meant non-public methods.

ShivanshRakesh: > The alternative is to dress these classes up so that they more closely resemble OpenMRS coding conventions: Setters should be `public` or at least `protected`. We should drop `the` and `my` as prefixes to variables, etc. I'd also like to minimize duplicating code as much as possible.

Okay.

omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenMRSNarrativeTemplateManifest.java
```diff
@@ -117,10 +118,6 @@ public static OpenMRSNarrativeTemplateManifest forManifestFileLocation(Collectio
 		return forManifestFileContents(manifestFileContents);
 	}
 	
-	public static OpenMRSNarrativeTemplateManifest forManifestFileContents(String... resources) throws IOException {
-		return forManifestFileContents(Arrays.asList(resources));
-	}
```
ShivanshRakesh: Removed the `forManifestFileContents(String... resources)`, `forManifestFileLocation(String... propertyFilePaths)` methods since they were not used anywhere.

omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenMRSNarrativeTemplateManifest.java
```diff
@@ -51,8 +63,6 @@
 	
 	private final Map<String, List<OpenMRSNarrativeTemplate>> styleToNameToTemplate;
 	
-	private final int templateCount;
```
ShivanshRakesh: Removed `templateCount` since it was not being used anywhere.

ibacher:
@ShivanshRakesh Any reason to leave this PR in draft status rather than making it Ready for review?

ShivanshRakesh:
I actually planned to make it ready for review with the previous commit, but I forgot :sweat_smile:.

ShivanshRakesh:
I actually planned to make it ready for review with the previous commit, but it just slipped off my mind :sweat_smile:.

ShivanshRakesh:
Added tests for coverage.

ShivanshRakesh:
Thanks @ibacher! :)

PR #259, opened 2020-07-17
FM2-246: Add search for _include using Location:partof

varung-31:

Add search for _include using Location:partof
## Issue I worked on

see https://issues.openmrs.org/browse/FM2-246

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/FhirService.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/FhirService.java
@@ -11,11 +11,16 @@ package org.openmrs.module.fhir2.api;
 
 import javax.validation.constraints.NotNull;
 
+import java.util.Collection;
+import java.util.List;
+
 import org.hl7.fhir.instance.model.api.IAnyResource;
 
 public interface FhirService<T extends IAnyResource> {
 	
 	T get(@NotNull String uuid);
+	
+	List<T> get(@NotNull Collection<String> distinctUUIDs);
 	
 	T create(T newResource);
 	
--- a/api/src/main/java/org/openmrs/module/fhir2/api/search/SearchQueryIncludeImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/search/SearchQueryIncludeImpl.java
@@ -11,23 +11,62 @@ package org.openmrs.module.fhir2.api.search;
 
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 
+import ca.uhn.fhir.model.api.Include;
 import lombok.NoArgsConstructor;
 import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.hl7.fhir.r4.model.Location;
+import org.hl7.fhir.r4.model.Reference;
+import org.openmrs.module.fhir2.FhirConstants;
+import org.openmrs.module.fhir2.api.FhirLocationService;
+import org.openmrs.module.fhir2.api.search.param.PropParam;
 import org.openmrs.module.fhir2.api.search.param.SearchParameterMap;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
+import org.springframework.util.CollectionUtils;
 
 @Component
 @NoArgsConstructor
 public class SearchQueryIncludeImpl<U extends IBaseResource> implements SearchQueryInclude<U> {
 	
+	@Autowired
+	private FhirLocationService locationService;
+	
 	@Override
+	@SuppressWarnings("unchecked")
 	public Set<IBaseResource> getIncludedResources(List<U> resourceList, SearchParameterMap theParams) {
 		Set<IBaseResource> includedResourcesSet = new HashSet<>();
 		
-		// TODO -> Add implementation to compute the included resources
+		List<PropParam<?>> includeParamList = theParams.getParameters(FhirConstants.INCLUDE_SEARCH_HANDLER);
+		
+		if (CollectionUtils.isEmpty(includeParamList)) {
+			return includedResourcesSet;
+		}
+		
+		Set<Include> includeSet = (HashSet<Include>) includeParamList.get(0).getParam();
+		includeSet.forEach(includeParam -> {
+			switch (includeParam.getParamName()) {
+				case FhirConstants.INCLUDE_PART_OF_PARAM:
+					includedResourcesSet.addAll(handleParentLocationInclude((List<Location>) resourceList));
+					break;
+			}
+		});
 		
 		return includedResourcesSet;
 	}
+	
+	private List<Location> handleParentLocationInclude(List<Location> resourceList) {
+		Set<String> uniqueParentLocationUUIDs = resourceList.stream().map(Location::getPartOf)
+		        .map(SearchQueryIncludeImpl::getIdFromReference).filter(Objects::nonNull).collect(Collectors.toSet());
+		
+		return locationService.get(uniqueParentLocationUUIDs);
+	}
+	
+	private static String getIdFromReference(Reference reference) {
+		return reference != null ? reference.getReferenceElement().getIdPart() : null;
+	}
+	
 }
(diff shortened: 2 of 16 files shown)


api/src/main/java/org/openmrs/module/fhir2/api/search/SearchQueryIncludeImpl.java
```diff
@@ -11,23 +11,71 @@
...
+				resourceList.forEach(
+				    resource -> uniqueParentLocationUUIDs.add(getIdFromReference(((Location) resource).getPartOf())));
+				
+				uniqueParentLocationUUIDs.removeIf(Objects::isNull);
+				includedResources.addAll(locationService.get(uniqueParentLocationUUIDs));
+				break;
+		}
+		
+		return includedResources;
+	}
```
ibacher, line 75: This seems unnecessarily verbose. Couldn't this whole thing be simplified to something like:

```
private static Set<Location> handlePartofInclude(List<Location> locations) {
	resourceList.stream().map(Location::getPartOf).filter(Objects::nonNull).map(ref -> getIdFromReference(ref)).collect(Collectors.toSet());
}
```

I.e., do we actually need a setup similar to what we do for searching (where the point is to be able to combine a series of operations to produce a single Criteria object for running the query) or can we get away with something much simpler?

api/src/main/java/org/openmrs/module/fhir2/api/FhirService.java
```diff
@@ -11,12 +11,17 @@
...
+import java.util.List;
+import java.util.Set;
+
 import org.hl7.fhir.instance.model.api.IAnyResource;
 
 public interface FhirService<T extends IAnyResource> {
 	
 	T get(@NotNull String uuid);
 	
+	List<T> get(@NotNull Set<String> distinctUUIDs);
```
ibacher, line 23: Does this *need* to take a set? Could we instead support a list, array, or even generic collection?

api/src/main/java/org/openmrs/module/fhir2/api/search/SearchQueryIncludeImpl.java
varung-31: No, so the thing here is that we could have partof for multiple resources in the future. Its use is not clearly evident here but let us say we support _include=Location:partof as well as _include=Observation:partof, instead of creating a separate function for each of these, we can just club them together based on the resource type.

varung-31: This would look something like - 
```
	private Set<IBaseResource> handlePatientInclude(List<U> resourceList, String paramType) {
		Set<IBaseResource> includedResources = new HashSet<>();
		Set<String> uniquePatientUUIDs = new HashSet<>();
		
		switch (paramType) {
			case FhirConstants.OBSERVATION:
				resourceList.forEach(
				    resource -> uniquePatientUUIDs.add(getIdFromReference(((Observation) resource).getSubject())));
				break;
			case FhirConstants.ALLERGY_INTOLERANCE:
				resourceList.forEach(
				    resource -> uniquePatientUUIDs.add(getIdFromReference(((AllergyIntolerance) resource).getPatient())));
				break;
			case FhirConstants.DIAGNOSTIC_REPORT:
				resourceList.forEach(
				    resource -> uniquePatientUUIDs.add(getIdFromReference(((DiagnosticReport) resource).getSubject())));
				break;
			case FhirConstants.ENCOUNTER:
				resourceList.forEach(
				    resource -> uniquePatientUUIDs.add(getIdFromReference(((Encounter) resource).getSubject())));
				break;
			case FhirConstants.PERSON:
				resourceList.forEach(resource -> {
					List<Reference> patientReferenceList = new ArrayList<>();
					((Person) resource).getLink().forEach(patient -> patientReferenceList.add(patient.getTarget()));
					uniquePatientUUIDs.addAll(getIdsFromReferenceList(patientReferenceList));
				});
				break;
			case FhirConstants.RELATED_PERSON:
				resourceList.forEach(resource -> uniquePatientUUIDs.add(getIdFromReference(((RelatedPerson) resource).getPatient())));
				break;
		}
		
		uniquePatientUUIDs.removeIf(Objects::isNull);
		includedResources.addAll(patientService.get(uniquePatientUUIDs));
		
		return includedResources;
	}
```

api/src/main/java/org/openmrs/module/fhir2/api/FhirService.java
varung-31: It can have other types as well, but I chose a set so that we only supply distinct uuids. To make the list as unique, we will anyway need to put it in a set, why not just pass that set directly. What do you think?

api/src/main/java/org/openmrs/module/fhir2/providers/r3/LocationFhirResourceProvider.java
```diff
@@ -84,8 +88,13 @@ public IBundleProvider searchLocations(@OptionalParam(name = Location.SP_NAME) S
 	                Location.SP_ADDRESS_STATE, Location.SP_ADDRESS_COUNTRY,
 	                Location.SP_ADDRESS_POSTALCODE }, targetTypes = Location.class) ReferenceAndListParam parent,
 	        @OptionalParam(name = Location.SP_RES_ID) TokenAndListParam id,
-	        @OptionalParam(name = "_lastUpdated") DateRangeParam lastUpdated, @Sort SortSpec sort) {
+	        @OptionalParam(name = "_lastUpdated") DateRangeParam lastUpdated,
+	        @IncludeParam(allow = "Location:partof") HashSet<Include> theIncludes, @Sort SortSpec sort) {
```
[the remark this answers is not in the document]
varung-31: I actually used `theIncludes` because we have been using `theParams` for SearchParameterMap.

api/src/main/java/org/openmrs/module/fhir2/api/search/SearchQueryIncludeImpl.java
ibacher: So, I can get that there are cases where this pattern can be useful, but in the specific case of `Location:partof` I don't think it is. I think, like with your patient example, it makes sense to group the includes into functions *based on the result type rather than the search query*.

For instance, if we had a `handlePartOf()` method that handles both Location and Observation, there's essentially no logic shared between those paths. But it would make sense to have a `handleLocation()` function that takes various parameters to extract Location UUIDs and then runs a single `locationService.get()` call.

api/src/main/java/org/openmrs/module/fhir2/api/FhirService.java
ibacher: So, while we want only the unique list of results, we can also defer the uniqueness part to the database (`select * from Concept where uuid in ('e110793d-939b-4661-82c1-b613883ea08d')` produces exactly the same output as `select * from Concept where uuid in ('e110793d-939b-4661-82c1-b613883ea08d', 'e110793d-939b-4661-82c1-b613883ea08d')`).

But my real thinking here was that if we build a method that takes a `Collection` instead of a `Set`, we can still continue to pass in a set, but we also gain some flexibility in usage (so that if someone has constructed a List, etc., they don't wind up having to convert it to a Set type).

api/src/main/java/org/openmrs/module/fhir2/api/search/SearchQueryIncludeImpl.java
ibacher: Incidentally, functions based on the result type are essentially what your `handlePatientInclude()` does.

api/src/main/java/org/openmrs/module/fhir2/providers/r3/LocationFhirResourceProvider.java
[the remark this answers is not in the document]
ibacher: Ugh... we should really fix that too. I suppose we can defer it.

api/src/main/java/org/openmrs/module/fhir2/api/FhirService.java
varung-31: Yes, that makes sense. I'll change it to collection.

varung-31:
@ibacher  Pushed the changes. Please take a look!

PR #264, opened 2020-07-27
FM2-185: Immunization Resource as CIEL:1421 obs group.

mks-d:
## Issue I worked on

see https://issues.openmrs.org/browse/FM2-185


api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ImmunizationTranslatorImpl.java
```diff
@@ -0,0 +1,215 @@
...
+		immunization.setPatient(
+		    new Reference().setType(PATIENT).setReference(PATIENT + "/" + openMrsImmunization.getPerson().getUuid()));
+		immunization.setEncounter(new Reference().setType(ENCOUNTER)
+		        .setReference(ENCOUNTER + "/" + openMrsImmunization.getEncounter().getVisit().getUuid()));
+		immunization.setPerformer(Arrays.asList(new ImmunizationPerformerComponent(new Reference().setType(PRACTITIONER)
+		        .setReference(PRACTITIONER + "/" + getAdministeringProvider(openMrsImmunization).getUuid()))));
+		
+		Map<String, Obs> members = getObsMembersMap(openMrsImmunization);
+		
+		CodeableConcept codeableConcept = new CodeableConcept();
```
ibacher, line 194: Here, I would use our default `ConceptTranslator` to translate the OpenMRS concept into a codeable concept.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ImmunizationTranslatorImpl.java
```diff
@@ -0,0 +1,215 @@
...
+	}
+	
+	@Override
+	public Immunization toFhirResource(Obs openMrsImmunization) {
+		validateImmunizationObsGroup(openMrsImmunization);
+		
+		Immunization immunization = new Immunization();
+		immunization.setId(openMrsImmunization.getUuid());
+		immunization.setStatus(ImmunizationStatus.COMPLETED);
+		immunization.setPatient(
```
ibacher, line 185: Here, I would use our methods from `BaseReferenceHandlingTranslator` or use the corresponding  `PatientReferenceTranslator` and `PractitionerReferenceTranslator` classes as well as the `VisitReferenceTranslator` class you added. While there's nothing wrong with this, it does help ensure some consistency in what we're populating (e.g., our default implementations of patient references and practitioner references include the `display` field, which in turn helps populate the narrative, etc).

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ImmunizationTranslatorImpl.java
```diff
@@ -0,0 +1,215 @@
...
+		        .setReference(PRACTITIONER + "/" + getAdministeringProvider(openMrsImmunization).getUuid()))));
+		
+		Map<String, Obs> members = getObsMembersMap(openMrsImmunization);
+		
+		CodeableConcept codeableConcept = new CodeableConcept();
+		Coding coding = new Coding();
+		coding.setCode(members.get(ciel984).getValueCoded().getUuid());
+		coding.setDisplay(members.get(ciel984).getValueCoded().getFullySpecifiedName(Context.getLocale()).getName());
+		codeableConcept.addCoding(coding);
+		immunization.setVaccineCode(codeableConcept);
```
ibacher, line 199: For at least a couple of these, you could use the `ObservationValueTranslator`, though it won't work for the Immunization.protocolApplied

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/LocationReferenceTranslatorImpl.java
```diff
@@ -0,0 +1,55 @@
...
+import org.openmrs.module.fhir2.api.translators.LocationReferenceTranslator;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+@Setter(AccessLevel.PACKAGE)
+public class LocationReferenceTranslatorImpl extends BaseReferenceHandlingTranslator implements LocationReferenceTranslator {
+	
+	@Autowired
+	private LocationService locationService;
```
ibacher, line 26: It would be preferable, I think, to use the module-specific `FhirLocationDao`.

api/src/main/java/org/openmrs/module/fhir2/providers/r4/ImmunizationFhirResourceProvider.java
```diff
@@ -0,0 +1,101 @@
...
+		
+		existingImmunization.setId(id.getIdPart());
+		
+		return FhirProviderUtils.buildUpdate(immunizationService.updateImmunization(id.getIdPart(), existingImmunization));
+	}
+	
+	@Search
+	@SuppressWarnings("unused")
+	public Collection<Immunization> searchImmunizations(@OptionalParam(name = Immunization.SP_PATIENT, chainWhitelist = { "",
+	        Patient.SP_IDENTIFIER }) ReferenceAndListParam patienParam, @Sort SortSpec sort) {
```
ibacher, line 97: `patientParam`

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/BaseImmunizationTranslator.java
```diff
@@ -0,0 +1,127 @@
...
+import org.openmrs.Concept;
+import org.openmrs.EncounterRole;
+import org.openmrs.Obs;
+import org.openmrs.Provider;
+import org.openmrs.api.APIException;
+import org.openmrs.api.ConceptService;
+import org.openmrs.module.fhir2.FhirActivator;
+
+@Setter(AccessLevel.PACKAGE)
+public class BaseImmunizationTranslator {
```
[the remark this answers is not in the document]
mks-d: >I'm kind of just wondering why this class exists?

@ibacher good point, and probably sketchy design on my end to use inheritance here. In fact this should all be wrapped up in a immunizations helper/utils class that would help keep the actual translator thinner. Would that be ok with you? And if yes, where should this class live?

api/src/main/java/org/openmrs/module/fhir2/api/FhirImmunizationService.java
```diff
@@ -0,0 +1,30 @@
...
+import javax.validation.constraints.NotNull;
+
+import java.util.Collection;
+
+import ca.uhn.fhir.rest.annotation.Sort;
+import ca.uhn.fhir.rest.api.SortSpec;
+import ca.uhn.fhir.rest.param.ReferenceAndListParam;
+import org.hl7.fhir.r4.model.Immunization;
+
+public interface FhirImmunizationService {
```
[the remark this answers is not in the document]
mks-d: Addressed [here](https://github.com/openmrs/openmrs-module-fhir2/pull/264/files#diff-c452ec7fab646a93c68818b16f9b2340R20).

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirImmunizationServiceImpl.java
```diff
@@ -0,0 +1,133 @@
...
+	
+	@Override
+	public Immunization updateImmunization(String uuid, Immunization updatedImmunization) {
+		Obs existingObs = obsService.getObsByUuid(uuid);
+		Obs obs = translator.toOpenmrsType(existingObs, updatedImmunization);
+		obs = obsService.saveObs(obs, "Updated when translating a FHIR Immunization resource.");
+		return translator.toFhirResource(obs);
+	}
+	
+	public Patient getPatient(ReferenceAndListParam patientParam) {
```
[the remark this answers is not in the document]
mks-d: I changed all references to `ReferenceAndListParam` to become `ReferenceParam`, I adapted the tests  accordingly... etc. My understanding is that right now either of those two can be done:
```
/Immunization?patient.identifier=10001X
```
```
/Immunization?patient=26678f02-7da5-428e-9497-a0ee5d44fbe4
```
I hope that [this](https://github.com/openmrs/openmrs-module-fhir2/pull/264/files#diff-0c788cc62b2dba56bcfc83734a5aa64fR97) is ok though in `ImmunizationFhirResourceProvider `?

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/BaseImmunizationTranslator.java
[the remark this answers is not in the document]
mks-d: I did introduce a new non-bean [`ImmunizationObsGroupHelper`](https://github.com/openmrs/openmrs-module-fhir2/pull/264/files#diff-1994b49f7c32134b749bdc067819ae56R38).

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ImmunizationTranslatorImpl.java
mks-d: Addressed [here](https://github.com/openmrs/openmrs-module-fhir2/pull/264/files#diff-56eb3743524091606fd9093205f0bc07R196).

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ImmunizationTranslatorImpl.java
mks-d: Addressed [here](https://github.com/openmrs/openmrs-module-fhir2/pull/264/files#diff-56eb3743524091606fd9093205f0bc07R189-R192).

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/LocationReferenceTranslatorImpl.java
mks-d: It turned out that `LocationReferenceTranslator` wasn't needed (yet), I removed it for now.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ImmunizationTranslatorImpl.java
mks-d: I actually could only do it for the vaccine _occurrence_ [here](https://github.com/openmrs/openmrs-module-fhir2/pull/264/files#diff-56eb3743524091606fd9093205f0bc07R197) because it takes a `Type` as argument. For all those that do not take a `Type` I'm not sure if there is a recommended way to do it?

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirImmunizationServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+		// TODO Simply void the underlying obs
+		throw new UnsupportedOperationException("Deleting a FHIR Immunization resource is currently not supported.");
+	}
+	
+	@Override
+	public Concept getOpenmrsImmunizationConcept() {
+		return helper.concept(immunizationGroupingConcept);
+	}
+	
+	public Patient getPatient(ReferenceParam patientParam) {
```
ibacher, line 100: I get the motivation here with the `IllegalArgumentException`s, but they're going to provide a sub-optimal experience for the FHIR client trying to consume these. If we are going to throw exceptions at this point, it's probably best to use HAPI's `UnprocessableEntityException` for this, which results in a FHIR-readable error with an appropriate structure including the message.

api/src/main/java/org/openmrs/module/fhir2/api/translators/VisitReferenceTranslator.java
```diff
@@ -0,0 +1,35 @@
...
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.translators;
+
+import org.hl7.fhir.r4.model.Reference;
+import org.openmrs.Encounter;
+import org.openmrs.Visit;
+
+public interface VisitReferenceTranslator extends OpenmrsFhirTranslator<Visit, Reference> {
```
ibacher, line 16: So, in the interim between when you started working on this and now, we added a `VisitReferenceTranslatorImpl` class, which can be gotten by requesting an instance of `EncounterReferenceTranslator<Visit>`. It would probably be a good idea to use that.

ibacher (commented):
@mks-d I haven't fully gone through this yet, but here are a few comments so far.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirImmunizationServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+		Obs obs = translator.toOpenmrsType(existingObs, updatedImmunization);
+		obs = obsService.saveObs(obs, "Updated when translating a FHIR Immunization resource.");
+		return translator.toFhirResource(obs);
+	}
+	
+	@Override
+	public Immunization delete(String uuid) {
+		// TODO Simply void the underlying obs
+		throw new UnsupportedOperationException("Deleting a FHIR Immunization resource is currently not supported.");
+	}
```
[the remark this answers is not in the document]
mks-d: @ibacher the idea here is that we would in fact support it "shortly", this was a temporary measure in the meantime. Can we keep it for now?

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirImmunizationServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+		this.patientService = patientService;
+	}
+	
+	@Autowired
+	public void setImmunizationHelper(ConceptService conceptService) {
+		this.helper = new ImmunizationObsGroupHelper(conceptService);
+	}
+	
+	@Override
+	public Immunization get(String uuid) {
```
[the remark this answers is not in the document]
mks-d: Yes, I'm about to push those changes. Actually I had to revert to using `ReferenceAndListParam` rather than just `ReferenceParam` in order to being able to leverage `FhirObservationDao` out of the box:
```java
public Collection<Immunization> searchImmunizations(ReferenceAndListParam patientParam, SortSpec sort) { ...
```

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirImmunizationServiceImpl.java
mks-d: This method will be gone :-)
I may have used `IllegalArgumentExceptions` in other places though, we'll see.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirImmunizationServiceImpl.java
```diff
@@ -0,0 +1,106 @@
...
+		}
+		obs = obsDao.createOrUpdate(obs);
+		return translator.toFhirResource(obs);
+	}
+	
+	@Override
+	public Immunization update(String uuid, Immunization updatedImmunization) {
+		Obs obs = translator.toOpenmrsType(obsDao.get(uuid), updatedImmunization);
+		obs = obsService.saveObs(obs, "Updated when translating a FHIR Immunization resource.");
+		//		obs = obsDao.createOrUpdate(obs);
```
mks-d, line 74: ↑ above is the last reference to `ObsService`, however uncommenting the line that switches to using `FhirObservationDao` instead does throw the infamous "cannot change immutable obs" exception. Any idea?

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirImmunizationServiceImpl.java
```diff
@@ -0,0 +1,106 @@
...
+		SearchParameterMap searchParams = new SearchParameterMap();
+		searchParams.addParameter(PATIENT_REFERENCE_SEARCH_HANDLER, patientParam);
+		TokenAndListParam conceptParam = new TokenAndListParam();
+		TokenParam token = new TokenParam();
+		token.setValue(getOpenmrsImmunizationConcept().getId().toString());
+		conceptParam.addAnd(token);
+		searchParams.addParameter(CODED_SEARCH_HANDLER, conceptParam);
+		
+		List<String> matchingResourceUuids = obsDao.getSearchResultUuids(searchParams);
+		Collection<Obs> obs = obsDao.getSearchResults(searchParams, matchingResourceUuids);
```
mks-d, line 101: I found this in a Spring test, and, I've got an issue with it.
Mostly that `getSearchResults` breaks down with an empty second param.

For now I'm doing some trials to work around that with an ugly if, but surely there's a better way that I don't know of to do something equivalent.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirImmunizationServiceImpl.java
ibacher: Hmmmm... yeah. This is actually why I'm not planning on implementing the "update" interaction for observations, mostly because the immutability requirement means that the UUID will change and so you have a PUT request operating to create a new resource and no good way to communicate that to the client. So, actually, my first recommendation would just be to not implement updates at all.

If we really do need to support updates to immunization resources, the only other alternative I could think of would be to essentially leave the main Obs itself alone, but void and replace the existing group members.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirImmunizationServiceImpl.java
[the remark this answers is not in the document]
ibacher: Sure. Btw, `obsDao.delete(uuid)` should at least void the main obs.

api/src/main/java/org/openmrs/module/fhir2/FhirActivator.java
```diff
@@ -19,8 +25,58 @@
 @Slf4j
 public class FhirActivator extends BaseModuleActivator {
 	
+	/**
+	 * @return The immunizations encounter type
+	 */
+	public static EncounterType getImmunizationsEncounterTypeOrCreateIfMissing() {
```
ibacher, line 31: So here's the last thing I'm a little hesitant about with this: should we be automatically creating metadata like this? I think I'd be happier if we just bailed with an error in the case where the necessary metadata didn't exist. This is because while we're creating default mapping for FHIR resources, part of the goal is to have the implementations be reasonably easy to override locally to meet whatever requirements, so I don't love the idea of creating unnecessary encounters for people who decide to create their own implementation of the Immunization resource.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/BaseReferenceHandlingTranslator.java
```diff
@@ -43,6 +44,10 @@ protected Reference createEncounterReference(@NotNull OpenmrsObject encounter) {
 		        .setType(FhirConstants.ENCOUNTER);
 	}
 	
+	protected Reference createVisitReference(@NotNull Visit visit) {
+		return new Reference().setReference(FhirConstants.VISIT + "/" + visit.getUuid()).setType(FhirConstants.ENCOUNTER);
```
ibacher, line 48: This should still return a valid FHIR reference, so it needs to be `Encounter/<uuid>` rather than `Visit/<uuid>`.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ImmunizationTranslatorImpl.java
```diff
@@ -0,0 +1,206 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+@Setter(AccessLevel.PACKAGE)
+public class ImmunizationTranslatorImpl implements ImmunizationTranslator {
+	
+	public static final String immunizationGroupingConcept = "CIEL:1421";
+	
+	public static final String[] immunizationConcepts = { "CIEL:984", "CIEL:1410", "CIEL:1418", "CIEL:1419", "CIEL:1420",
```
ibacher, line 54: Aside from populating the constants below, this seems to always be cast to a `List` using `Arrays.asList()` when it's read. It seems like it would be better to just create a list here (even if it's just `Arrays.asList(new String[] { "CIEL:984", "CIEL:1410", "CIEL:1418", "CIEL:1419", "CIEL:1420", "CIEL:165907" })`), since that involves allocating and deallocating fewer objects.

api/src/test/java/org/openmrs/module/fhir2/api/FhirImmunizationServiceTest.java
```diff
@@ -0,0 +1,262 @@
...
+		    equalTo(new DateTimeType("2022-07-31T18:30:00.000Z").getValue()));
+	}
+	
+	@Test
+	public void updateImmunization_shouldUpdateImmunizationAccordingly() throws ParseException {
+		
+		// setup
+		FhirContext ctx = FhirContext.forR4();
+		IParser parser = ctx.newJsonParser();
+		Immunization updatedImmunization = parser.parseResource(Immunization.class,
```
ibacher, line 166: I'd tend to prefer this being externalised in a file, but there's no real harm.

ibacher (commented):
I think we added some support for Visits mapped as Encounters, so may want to take a look at that in addition to the few other comments I've made here.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirImmunizationServiceImpl.java
ibacher: You should really be returning a `SearchQueryBundleProvider` from this method to support paging, etc. The `SearchQueryBundleProvider` is also designed to do those DAO calls directly. See [how things our structured in our other service classes](https://github.com/openmrs/openmrs-module-fhir2/blob/4f3ce0524f82735927abf17fc6da615e5b51e1e3/api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirPatientServiceImpl.java#L38).

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/BaseReferenceHandlingTranslator.java
mks-d: Turns out that this wasn't needed anymore, removing it.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirImmunizationServiceImpl.java
mks-d: Thanks @ibacher that was a very helpful tip, I did use `SearchQueryBundleProvider` in my next commit 👍

api/src/test/java/org/openmrs/module/fhir2/api/FhirImmunizationServiceTest.java
mks-d: I hear you, however for the ease of debugging I'll currently leave those to not be in test resources, if that's ok?

api/src/main/java/org/openmrs/module/fhir2/FhirActivator.java
mks-d: I just changed that, the activator is now left alone ;-)

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirImmunizationServiceImpl.java
ibacher: @mks-d Apart from this unresolved issue, I think the code looks good.

ibacher (approved):

PR #302, opened 2020-10-19
fix liquibase error

mozzy11:
Ticket :https://issues.openmrs.org/browse/FM2-308


dkayiwa:
A ticket id in the commit message would not hurt. Would it? 😊

sherrif10:
i would also make title as FM2-308:Fix Liquibase Error perhaps looks better

PR #308, opened 2020-11-19
 FM2-307: Implementing Condition Resource for OpenMRS 2.1 And Below

mozzy11:
## Issue I worked on

Ticket: https://issues.openmrs.org/browse/FM2-307

  No? -> execute above command


api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConditionDaoImpl.java
```diff
@@ -0,0 +1,288 @@
...
+	
+	@Override
+	@Authorized(PrivilegeConstants.DELETE_OBS)
+	public Obs delete(@Nonnull String uuid) {
+		return super.delete(uuid);
+	}
+	
+	@Override
+	@Authorized(PrivilegeConstants.GET_OBS)
+	public List<String> getSearchResultUuids(@Nonnull SearchParameterMap theParams) {
```
ibacher, line 100: Generally, we don't want to override `getSearchResultsUuids()` and change the implementation. The right way to do this is to add the custom logic we need to the `setupSearchParams()` method (which is overridable for just this purpose). That way we only keep this logic in one place.

api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
```diff
@@ -277,4 +277,6 @@
 	public static final String INCLUDE_RESULT_PARAM = "result";
 	
 	public static final String REVERSE_INCLUDE_SEARCH_HANDLER = "_revinclude.search.handler";
+	
+	public static final Integer CONDITION_OBSERVATION_CONCEPT_ID = 1284;
```
ibacher, line 281: It would be safer to use the UUID for the concept ("1284AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA") since we don't want to be relying on a concept having a given concept id (i.e., we don't want to rely on a dictionary having IDs in a certain order). Even better might be to just be to leverage the system for CIEL and the concept ID, via something like:

```java
TokenAndListParam problemList = new TokenAndListParam();
		problemList.addAnd(new TokenParam().setSystem("https://openconceptlab.org/orgs/CIEL/sources/CIEL").setValue("1284"));
		handleCodeableConcept(criteria, problemList, "c", "cm", "crt");
```

Then this "1284" string would have a place in the `FhirConstants` file.

test-data/src/test/resources/org/openmrs/module/fhir2/api/dao/impl/FhirObsConditionDaoImplTest_initial_data.xml
```diff
@@ -0,0 +1,27 @@
...
+	<obs obs_id="31" person_id="7" concept_id="1284" encounter_id="3" obs_datetime="2008-07-01 00:00:00.0" location_id="1"  comments="" creator="1" date_created="2008-08-18 14:09:35.0" voided="false" value_coded="116128"   uuid="86sgf-1f7d-4394-a316-0a458edf28c7"/>
+    <obs obs_id="32" person_id="7" concept_id="116128" encounter_id="3" obs_datetime="2008-07-01 10:00:00.0" location_id="1"  comments="" creator="1" date_created="2008-08-18 14:09:35.0" voided="false"  value_coded="116128" uuid="942ec003-a55d-43c4-ac7a-bd6d1ba63382"/>
+    <obs obs_id="33" person_id="7" concept_id="1284" encounter_id="3" obs_datetime="2008-07-01 10:00:00.0" location_id="1"  comments="" creator="1" date_voided="2010-09-03 14:09:35.0" date_created="2008-08-18 14:09:35.0" voided="true"  value_coded="116128" uuid="94dhs003-a55d-43c4-ac7a-bd6d1ba63388"/>	
+    <fhir_concept_source fhir_concept_source_id="1" name="Some Made-up Terminology" url="http://made_up_concepts.info/sct" concept_source_id="1" creator="1" date_created="2005-01-01 00:00:00.0" retired="0" uuid="0d30bea3-4ba2-4ab4-ac7b-5525840bde20" />	
+</dataset>
```
ibacher, line 27:
```suggestion
</dataset>

```

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConditionDaoImpl.java
```diff
@@ -0,0 +1,288 @@
...
+	}
+	
+	@Override
+	@Authorized(PrivilegeConstants.GET_OBS)
+	public List<Obs> getSearchResults(@Nonnull SearchParameterMap theParams, @Nonnull List<String> matchingResourceUuids,
+	        int firstResult, int lastResult) {
+		return super.getSearchResults(theParams, matchingResourceUuids, firstResult, lastResult);
+	}
+	
+	private Optional<Criterion> handleAgeByDateProperty(@Nonnull String datePropertyName, @Nonnull QuantityParam age) {
```
ibacher, line 134: It might be better to refactor things so this method is in `BaseDao` and can be used from both this class and the 2.2 and later implementation.

ibacher (changes requested):
@mozzy11 Thanks for this. I've added a few comments. In addition, we need some comprehensive unit tests for the `FhirConditionDaoImpl` class.

corneliouzbett:
Many thanks, @mozzy11 for working on this!  🙏🏽. 

However, I think this code belongs to emrapi -> fhircondition submodule rather than here. According to me both the implementation 'conditions in obs' & 'conditions from emrapi'  should be combined. I.e When searching for conditions (that is for openmrs platform version 2.0.5 - 2.1.*), it should search in `obs` table and `emrapi.conditions` table then combined the results. In regards to creating & updating conditions, I would prefer storing in emrapi conditions table.

FYI, this piece (submodule jar) of functionality will be loaded conditionally;
```java
<conditionalResource>
    <path>/lib/emrapi-fhir-condition-${project.parent.version}.jar</path>
        <loadIfModulesPresent>
            <openmrsModule>
                 <moduleId>org.openmrs.module.fhir2</moduleId>
                 <version>1.*</version>
             </openmrsModule>
         </loadIfModulesPresent>
      <openmrsVersion>2.0.5 - 2.1.*</openmrsVersion>
</conditionalResource>
```

corneliouzbett:
> Many thanks, @mozzy11 for working on this! 🙏🏽.
> 
> However, I think this code belongs to emrapi -> fhircondition submodule rather than here. According to me both the implementation 'conditions in obs' & 'conditions from emrapi' should be combined. I.e When searching for conditions (that is for openmrs platform version 2.0.5 - 2.1.*), it should search in `obs` table and `emrapi.conditions` table then combined the results. In regards to creating & updating conditions, I would prefer storing in emrapi conditions table.
> 
> FYI, this piece (submodule jar) of functionality will be loaded conditionally;
> 
> ```java
> <conditionalResource>
>     <path>/lib/emrapi-fhir-condition-${project.parent.version}.jar</path>
>         <loadIfModulesPresent>
>             <openmrsModule>
>                  <moduleId>org.openmrs.module.fhir2</moduleId>
>                  <version>1.*</version>
>              </openmrsModule>
>          </loadIfModulesPresent>
>       <openmrsVersion>2.0.5 - 2.1.*</openmrsVersion>
> </conditionalResource>
> ```

Just imaging an implementation with conditions stored in both tables, as obs, and in conditions table. I don't know if such an implementation exists.

mozzy11:
Thanks @corneliouzbett  that suggestion. 
i would think we can separate the implementation of `conditions in obs` and `conditions from emrapi `.
That is we implement `conditions in obs` here in the fhir2 , 
and we implement `conditions from emrapi` in emrapi as your suggestion above

the idea is to avoid conditionally implementing `conditions in obs` only if the emrapi module loaded.

Just in case an implementation has stored conditons in emrapi.conditons table . it can be implemented as fhir from the emrapi -> fhircondition submodule

corneliouzbett:
> Thanks @corneliouzbett for that suggestion.
> i would think we can separate the implementation of **conditions in obs** from that of **conditions from emrapi** .
> That is we implement **conditions in obs** here in the fhir2 ,
> and we implement **conditions from emrapi** in emrapi as your suggestion above
> 
> the idea is to avoid conditionally implementing **conditions in obs** only if the emrapi module loaded ie not to depend on emrapi module for the implementation of **conditions in obs**
> 
> Just in case an implementation has stored conditons in emrapi.conditons table . it can be implemented as fhir from the emrapi -> fhircondition submodule

Got you 👍🏼 
I would think this is a read-only implementation. Or are you planning to implement create & update?

mozzy11:
ideally for an implementation that runs openmrs version below 2.2 , they would be need to create new Conditon-Obs resources (translate back the Fhir Condition resource into a given obs object.)via the FHIR API ,. see this transaltion [here ](https://github.com/openmrs/openmrs-module-fhir2/blob/dcf7a76de2eb7eb5f0e3d07318da1d157bbf0926/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConditionTranslatorImpl.java#L103-L126)

However ,since Obs are not Updatable , These conditions are not Updatable

mozzy11:
@ibacher   , updated Pr as per your review comments

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConditionClinicalStatusTranslatorImpl.java
```diff
@@ -0,0 +1,38 @@
...
+@Component
+@OpenmrsProfile(openmrsPlatformVersion = "2.0.5 - 2.1.*")
+public class ConditionClinicalStatusTranslatorImpl implements ConditionClinicalStatusTranslator<Obs> {
+	
+	@Override
+	public CodeableConcept toFhirResource(Obs clinicalStatus) {
+		CodeableConcept codeableConcept = new CodeableConcept();
+		codeableConcept.addCoding().setCode("active").setDisplay("active")
+		        .setSystem(FhirConstants.CONDITION_CLINICAL_STATUS_SYSTEM_URI);
+		return codeableConcept;
```
ibacher, line 30: In this case, because the `clinicalStatus` is not a required field and we do not really have a way of representing it solely from the Obs, I would prefer that this implementation just return `null`. That is, if we do not know the right value, we shouldn't lie about knowing the right value (unless we're forced to guess).

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConditionDaoImpl.java
```diff
@@ -0,0 +1,135 @@
...
+				case FhirConstants.QUANTITY_SEARCH_HANDLER:
+					entry.getValue().forEach(param -> handleOnsetAge(criteria, (QuantityAndListParam) param.getParam()));
+					break;
+				case FhirConstants.COMMON_SEARCH_HANDLER:
+					handleCommonSearchParameters(entry.getValue()).ifPresent(criteria::add);
+					break;
+			}
+		});
+		criteria.createAlias("concept", "conc");
+		criteria.add(eq("conc.uuid", FhirConstants.CONDITION_OBSERVATION_CONCEPT_UUID));
```
ibacher, line 104: I'd prefer this to be aliased to `"c"` for consistency. Also, this should go before the `forEach` block.

ibacher:
So, I think this poses an interesting challenge for the module: how do we set things up so that we use the base Condition implementation in 2.2, the EMRAPI in 2.1 and below if that module is loaded and this implementation in 2.1 and below if EMRAPI isn't loaded.

mozzy11:
what i would think , we implement **conditons  in emrapi** in the emrapi module its self ,that is the fhir2 module is loaded

```
<conditionalResource>
    <path>/lib/emrapi-fhir-condition-${project.parent.version}.jar</path>
        <loadIfModulesPresent>
            <openmrsModule>
                 <moduleId>org.openmrs.module.fhir2</moduleId>
                 <version>1.*</version>
             </openmrsModule>
         </loadIfModulesPresent>
      <openmrsVersion>2.0.5 - 2.1.*</openmrsVersion>
</conditionalResource>
```

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConditionClinicalStatusTranslatorImpl.java
mozzy11: >  we shouldn't lie about knowing the right value (unless we're forced to guess).
haha  :blush:

mozzy11: @ibacher , i think let me just entirely remove it out

corneliouzbett:
> So, I think this poses an interesting challenge for the module: how do we set things up so that we use the base Condition implementation in 2.2, the EMRAPI in 2.1 and below if that module is loaded and this implementation in 2.1 and below if EMRAPI isn't loaded.

we can use global properties to track if the necessary modules are loaded then draft the necessary functionality based on that. OR trying searching for the registered beans in the context

mozzy11:
@ibacher worked on your review comments

mozzy11:
@ibacher ,Any comments for me ??

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConditionDaoImpl.java
mozzy11: sure , i did that

mozzy11:
@ibacher any Update on this ??

test-data/src/test/resources/org/openmrs/module/fhir2/api/dao/impl/FhirObsConditionDaoImplTest_initial_data.xml
```diff
@@ -0,0 +1,27 @@
...
```
ibacher, line 27:
```suggestion
</dataset>

```

test-data/src/test/resources/org/openmrs/module/fhir2/providers/ConditionWebTest_create_r3.xml
```diff
@@ -27,5 +27,5 @@
         </identifier>
         <display value="Horatio Hornblower (OpenMRS ID:101-6)"/>
     </subject>
-    <recordedDate value="2019-06-18T06:37:26+03:00"/>
+    <onsetDateTime value="2019-06-18T06:37:26+03:00"/>
 </Condition>
```
ibacher, line 31:
```suggestion
</Condition>

```

ibacher (approved):

PR #310, opened 2020-12-07
FM-189: Add Timing.TimingRepeatComponent.DurationUnit property

theanandankit:
## Issue I worked on

see https://issues.openmrs.org/browse/FM2-189

  No? -> execute above command


theanandankit:
i'm currently using the `drugOrder.getDurationUnits().getDateRetired()` and `drugOrder.getDurationUnits().getDateCreated()` to get the expected unit for Duration by calculating the duration gap between the creation and retirement.

corneliouzbett:
Many thanks, @theanandankit for working on this!

The Ideal implementation here should be mapping `drugorder.getDurationUnit()` to this [valueset](https://www.hl7.org/fhir/valueset-units-of-time.html) - units-of-time. However, `drugorder.getDurationUnit()` is a concept that makes is somehow challenging. To implement this, use the CIEL dictionary as the default, extract concept uuids for duration units in days, hour, minute, seconds e.t.c

A static implementation would look like;
```java
private TimeOfUnits getTimeOfUnits(DrugOrder drugOrder) {
String unitUUID = drugOrder.getDurationUnits().getUuid();
    switch (unitUUID) {
	case "XXX-XXX-XXX-XXX-XXX":
	  // Duration units in days
	  return Timing.UnitsOfTime.D;
	case "...":
	  // Duration units in minutes
	  return Timing.UnitsOfTime.MIN;
	 ...
    }
}
```

theanandankit:
thanks @corneliouzbett  for the review. Suggested approach would be helpful.

ibacher:
For this, I would take a look at the `org.openmrs.module.fhir2.api.mappings` package, specifically to setup a mapping that allows us to map between a duration unit (as defined by the FHIR spec) and a concept. We could then pre-populate this table for, e.g., CIEL if those concepts exist.

theanandankit:
@ibacher thank you for your review.

theanandankit:
I found the CIEL dictionary and got the respective UUID of concepts for [HOUR](https://openconceptlab.org/orgs/CIEL/sources/CIEL/concepts/1822/) [MIN](https://openconceptlab.org/orgs/CIEL/sources/CIEL/concepts/1733/) [SECONDS](https://openconceptlab.org/orgs/CIEL/sources/CIEL/concepts/162583/) [WEEKS](https://openconceptlab.org/orgs/CIEL/sources/CIEL/concepts/1073/) . so that we can map the duration unit with their concepts. @ibacher @corneliouzbett PTAL.

ibacher:
@theanandankit Yep, those are the concepts we'd likely want to map. More adequately, we can probably map the concepts found in [this set](https://openconceptlab.org/orgs/CIEL/sources/CIEL/concepts/1732/) (so adding month and year, which are also part of the FHIR duration units).

theanandankit:
@ibacher thanks, commit the mapping ASAP

theanandankit:
@ibacher @corneliouzbett i've made the change PTAL

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/DurationUnitTranslatorImpl.java
```diff
@@ -0,0 +1,36 @@
...
+public class DurationUnitTranslatorImpl implements DurationUnitTranslator {
+
+	@Override
+	public Timing.UnitsOfTime toFhirResource(@Nonnull DrugOrder drugOrder) {
+
+		String unitUUID = drugOrder.getDurationUnits().getUuid();
+
+		switch (unitUUID) {
+
+			case "162583":
```
ibacher, line 18: So these aren't actually what the UUIDs for these concepts would look like in practice. If you look at the concepts as displayed on openconceptlab.org, you'll see an "External ID". That (at least for CIEL) corresponds to the UUID as it would exist in the OpenMRS concept dictionary. For example, [CIEL:1822 (Hour)](https://openconceptlab.org/orgs/CIEL/sources/CIEL/concepts/1822/) has an external id of "1822AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" (this is actually the pattern for CIEL concept UUIDs: the concept id padded with As to reach 36 characters).

That said, I don't want to have concept ids hard-coded in the translator like this. I'd prefer if we did something closer to what we have for `ObservationCategoryMap` which maps between the [FHIR Observation Category terminology set](https://www.hl7.org/fhir/codesystem-observation-category.html) and OpenMRS concept classes. This support requires a few different components:

1. We have a translator, `ObservationCategoryTranslator` and a corresponding [implementation](https://github.com/openmrs/openmrs-module-fhir2/blob/master/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationCategoryTranslatorImpl.java) which adds the observation category to the FHIR resource.

2. We have a simple table that stores mappings between OpenMRS concept classes and the FHIR Observation Category. This table is defined by [this class](https://github.com/openmrs/openmrs-module-fhir2/blob/master/api/src/main/java/org/openmrs/module/fhir2/model/FhirObservationCategoryMap.java).

3. Finally, we have [this simple class](https://github.com/openmrs/openmrs-module-fhir2/blob/master/api/src/main/java/org/openmrs/module/fhir2/api/mappings/ObservationCategoryMap.java) which defines the lookups to get the category given the concept class UUID and to get the concept class UUID using the category.

The flexibility afforded by this model is that we're able to map arbitrary OpenMRS concept dictionaries (not all of which use the CIEL concepts) to the corresponding FHIR terminology. Does that make some kind of sense?

theanandankit:
@ibacher thanks, i'm working on this as suggested by you and have couple of questions.

https://github.com/theanandankit/openmrs-module-fhir2/blob/d66a68eac9cf71e18fa7224f508760b90609f894/api/src/main/java/org/openmrs/module/fhir2/model/FhirDurationUnitMap.java#L28-L41

is this the required table that we looking for ?? i've gone through the openMRS concept dictionary and didn't found "1822AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" or any as UUID, i think i'm finding the UUID in the wrong table. can you help me out ?

theanandankit:
@ibacher thanks, i'm working on this and have some questions

https://github.com/theanandankit/openmrs-module-fhir2/blob/d66a68eac9cf71e18fa7224f508760b90609f894/api/src/main/java/org/openmrs/module/fhir2/model/FhirDurationUnitMap.java#L28-L41

Is this the required table that we are looking for ?? i've gone through the openMRS concept dictionary and didn't find "1822AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" or any as UUID, i think i'm finding the UUID in the wrong table. can you help me out ?

theanandankit:
https://github.com/theanandankit/openmrs-module-fhir2/blob/test/api/src/main/java/org/openmrs/module/fhir2/model/FhirDurationUnitMap.java

@ibacher can you review is this ok ? or need any kind of change

theanandankit:
Screenshot from 2020-12-11 19-05-10
@ibacher unable to find the concept with uuid="1822AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA". i'm using the MySQL Workbench. PTAL

theanandankit:
https://gist.github.com/theanandankit/a7f779b2498623ed50518514b89fcc07 
@ibacher  i've added the table definition in liquibase.xml. is this ok ?

ibacher:
> unable to find the concept with uuid="1822AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA". i'm using the MySQL Workbench. PTAL

So, being able to find that concept in your install really depends on how you went about setting up the install. If you created an OpenMRS instance using the RefApp, you would have it in your concept table. If you created an OpenMRS instance using only the *platform* you wouldn't. This is because the RefApp includes a range of OpenMRS-related metadata, but the platform generally does not (unless you choose to add the metadata). Does that make sense?

> @ibacher can you review this, is this ok or need any kind of change ?

You need to provide an strategy for handling the `Timing.UnitsOfTime` enum so that Hibernate knows what to store. Long story short, you need to add an annotation like this `@Enumerated(EnumType.STRING)` to that field. Otherwise, things won't work.

Additionally, the order of your imports is somewhat off from the default OpenMRS format. The imports should look like this:

```java
import javax.persistence.Column;
import javax.persistence.Entity;
import javax.persistence.GeneratedValue;
import javax.persistence.GenerationType;
import javax.persistence.Id;
import javax.persistence.JoinColumn;
import javax.persistence.ManyToOne;
import javax.persistence.Table;

import java.util.Date;
import java.util.UUID;

import lombok.Data;
import lombok.EqualsAndHashCode;
import org.hibernate.search.annotations.Field;
import org.hl7.fhir.r4.model.Timing;
import org.openmrs.Auditable;
import org.openmrs.Concept;
import org.openmrs.Retireable;
import org.openmrs.User;
```

Assuming you're using IntelliJ (the format we use is the default for Eclipse), see [our documentation on how to configure your IDE for this](https://wiki.openmrs.org/display/docs/Developer+How-To+Setup+And+Use+IntelliJ#DeveloperHow-ToSetupAndUseIntelliJ-JavaImports) or just add this to the FHIR2 module and run `mvn clean install`, as we have a Maven plugin that will re-order imports for you.

Otherwise, the file looks fine.

> @ibacher i've added the table definition in liquibase.xml. is this ok ?

So comparing this with the file for the Hibernate mappings you created, the column named `duration_unit` should be renamed to `unit_of_time` and can probably be shrunk to, e.g. `varchar(20)`. Otherwise things look fine.

In the future, it would be good if you can add files you'd like me to review to the PR (just commit them and push your local branch) so I can take advantage of GitHub's tools rather than having to click a bunch of links.

Thanks for all your effort on this!

theanandankit:
@ibacher thanks , i'll take care of this in future's PRs. suggested information would be helpful. commit the following changes ASAP.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/DurationUnitTranslatorImpl.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class DurationUnitTranslatorImpl implements DurationUnitTranslator {
+	
+	@Autowired
+	private DurationUnitMap durationUnitMap;
+	
+	@Override
+	public Timing.UnitsOfTime toFhirResource(@Nonnull DrugOrder drugOrder) {
+		
+		if (drugOrder.getDurationUnits().getUuid() == null) {
```
ibacher, line 29:
```suggestion
		if (drugOrder.getDurationUnits() == null || drugOrder.getDurationUnits().getUuid() == null) {
```

api/src/main/java/org/openmrs/module/fhir2/api/mappings/DurationUnitMap.java
```diff
@@ -0,0 +1,47 @@
...
+public class DurationUnitMap {
+	
+	@Autowired
+	@Qualifier("sessionFactory")
+	private SessionFactory sessionFactory;
+	
+	public Timing.UnitsOfTime getDurationUnit(@Nonnull String conceptUuid) {
+		
+		try {
+			return (Timing.UnitsOfTime) sessionFactory.getCurrentSession().createCriteria(FhirDurationUnitMap.class)
```
ibacher, line 37:
```suggestion
			return sessionFactory.getCurrentSession().createCriteria(FhirDurationUnitMap.class,  Timing.UnitsOfTime.class)
```

api/src/main/java/org/openmrs/module/fhir2/api/mappings/DurationUnitMap.java
```diff
@@ -0,0 +1,47 @@
...
+			        .createAlias("concept", "c").add(eq("c.uuid", conceptUuid))
+			        .setProjection(Projections.property("unit_of_time")).uniqueResult();
+		}
+		catch (HibernateException e) {
```
ibacher, line 41: I don't love swallowing random exceptions like this; I think I'd rather tend to fail the operation (i.e. just throw the exception) with the possible exception of `NonUniqueResultException`. That we can do something useful with if we catch it.

api/src/main/java/org/openmrs/module/fhir2/api/translators/DurationUnitTranslator.java
```diff
@@ -0,0 +1,21 @@
...
+
+import javax.annotation.Nonnull;
+
+import org.hl7.fhir.r4.model.Timing;
+import org.openmrs.DrugOrder;
+
+public interface DurationUnitTranslator extends ToFhirTranslator<DrugOrder, Timing.UnitsOfTime> {
+	
+	@Override
+	Timing.UnitsOfTime toFhirResource(@Nonnull DrugOrder drugOrder);
```
ibacher, line 20: Should this map between the `DrugOrder` or the `Concept`? If we define this to take a concept then it could be more generally useful, if we need these mappings in other places.

api/src/main/java/org/openmrs/module/fhir2/model/FhirDurationUnitMap.java
```diff
@@ -0,0 +1,90 @@
...
+public class FhirDurationUnitMap implements Auditable, Retireable {
+	
+	@EqualsAndHashCode.Include
+	@Id
+	@GeneratedValue(strategy = GenerationType.AUTO)
+	@Column(name = "duration_unit_map_id")
+	private Integer id;
+	
+	@ManyToOne(optional = false)
+	@JoinColumn(nullable = false, name = "concept_uuid")
```
ibacher, line 48:
```suggestion
	@JoinColumn(nullable = false, name = "concept_id")
```

api/src/main/resources/liquibase.xml
```diff
@@ -696,4 +696,237 @@
...
+				</sqlCheck>
+
+				<sqlCheck expectedResult="1">
+					select 1 from concept where uuid = '1074AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
+				</sqlCheck>
+
+				<sqlCheck expectedResult="1">
+					select 1 from concept where uuid = '1734AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
+				</sqlCheck>
+
```
ibacher, line 785:
```suggestion

				<sqlCheck expectedResult="1">
					select 1 from concept where uuid = '162583AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
				</sqlCheck>
				<sqlCheck expectedResult="1">
					select 1 from concept where uuid = '1733AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
				</sqlCheck>
				<sqlCheck expectedResult="1">
					select 1 from concept where uuid = '1822AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
				</sqlCheck>
				<sqlCheck expectedResult="1">
					select 1 from concept where uuid = '1072AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
				</sqlCheck>
				<sqlCheck expectedResult="1">
					select 1 from concept where uuid = '1073AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
				</sqlCheck>
				<sqlCheck expectedResult="1">
					select 1 from concept where uuid = '1074AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
				</sqlCheck>
				<sqlCheck expectedResult="1">
					select 1 from concept where uuid = '1734AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
				</sqlCheck>
```

api/src/main/resources/liquibase.xml
```diff
@@ -696,4 +696,237 @@
...
```
ibacher, line 874:
```suggestion

				<sqlCheck expectedResult="1">
					select 1 from concept where uuid = '162583AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
				</sqlCheck>
				<sqlCheck expectedResult="1">
					select 1 from concept where uuid = '1733AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
				</sqlCheck>
				<sqlCheck expectedResult="1">
					select 1 from concept where uuid = '1072AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
				</sqlCheck>
				<sqlCheck expectedResult="1">
					select 1 from concept where uuid = '1073AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
				</sqlCheck>
				<sqlCheck expectedResult="1">
					select 1 from concept where uuid = '1073AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
				</sqlCheck>
				<sqlCheck expectedResult="1">
					select 1 from concept where uuid = '1074AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
				</sqlCheck>
				<sqlCheck expectedResult="1">
					select 1 from concept where uuid = '1734AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
				</sqlCheck>
```

ibacher (commented):
@theanandankit Thanks again! This is looking very good! I think the one main piece that's missing here is some unit tests to show that this is working. Alternatively, instead of specific unit tests, I suppose we could just add tests to translating `DrugOrder` to show that this translator works.

theanandankit:
@ibacher thanks again for your time. recommended changes would be helpful.

ibacher:
@theanandankit Recommended changes for what, exactly? Happy to provide whatever examples I can...

theanandankit:
@ibacher I meant to say that the changes you suggested would be helpful.

theanandankit:
@ibacher Committed the tests for durationUnitTranslatorImpl PTAL. I have two questions in my mind:

1)  Preconditions is failing every time. I'm using the openMRS instance to test this code but when I was using the MySQL workbench with the same SQL query, it was working fine as expected. Am I missing anything?
```
WARNING 13/12/20 12:56 AM:liquibase: Continuing past: liquibase.precondition.core.PreconditionContainer@65a1a0d6 despite precondition error:
      1 preconditions failed
     liquibase.xml : Empty result set, expected one row
```
2) How can we test the durationUnitTranslationImpl as concept data isn’t available on the platform. It will always return the null ?

theanandankit:
@ibacher I come across that when our FHIR2 module instantiating after running the `mvn openmrs-sdk:run` by that time concept table doesn't have any value in it. I think that was the reason for failing preconditions.  How can we instantiate the FHIR-2 module after the concept table fully loaded?

theanandankit:
@ibacher I've made the changes. PTAL

api/src/main/resources/liquibase.xml
```diff
@@ -696,4 +696,223 @@
         ]]>
         </sql>
     </changeSet>
+
+	<changeSet id="add_fhir_duration_unit_map_20200930" author="ibacher">
```
ibacher, line 700:
```suggestion
	<changeSet id="add_fhir_duration_unit_map_20200930" author="theanandakit">
```

api/src/main/resources/liquibase.xml
```diff
@@ -696,4 +696,223 @@
...
+		<sql><![CDATA[
+            insert into fhir_duration_unit_map (concept_id, unit_of_time, creator, date_created, uuid)
+            select concept_id, 'A' ,1, now(), uuid()
+            from concept
+            where uuid = '1734AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
+        ]]>
+		</sql>
+	</changeSet>
+
+	<changeSet id="add_default_duration_unit_20200930" author="ibacher" dbms="postgresql">
```
ibacher, line 832:
```suggestion
	<changeSet id="add_default_duration_unit_20200930" author="theanandakit" dbms="postgresql">
```

ibacher (approved):

PR #319, opened 2021-01-27
FM2-337: Implement Group resource

corneliouzbett:

Implement Group resource

Supported properties 👇🏾
```json
{
  "resourceType": "Group",
  "id": "08f90de6-887b-40a7-8200-3e751cd4fde0",
  "extension": [
    {
      "url": "http://fhir.openmrs.org/ext/group/description",
      "valueString": "Patients with at least one encounter"
    }
  ],
  "active": true,
  "type": "person",
  "actual": true,
  "name": "Test group",
  "quantity": 1,
  "member": [
    {
      "id": "56c70a72-503b-4bcb-ba5b-e8bb49ebd901",
      "entity": {
        "reference": "Patient/b70c8cf2-379c-49ca-9f9d-b8aa64d59171",
        "type": "Patient",
        "display": "Michelle Pérez (OpenMRS ID: 100GCN)"
      },
      "period": {
        "start": "2019-10-16T00:00:28+03:00"
      },
      "inactive": false
    }
  ]
}
```
## Issue I worked on

see https://issues.openmrs.org/browse/FM2-337

  No? -> execute above command


api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/GroupMemberTranslatorImpl_2_1.java
```diff
@@ -0,0 +1,95 @@
...
+			        .setPatientId(patientReferenceTranslator.toOpenmrsType(groupMemberComponent.getEntity()).getPatientId());
+		}
+		
+		if (groupMemberComponent.hasPeriod()) {
+			existingCohort.setStartDate(groupMemberComponent.getPeriod().getStart());
+			existingCohort.setEndDate(groupMemberComponent.getPeriod().getEnd());
+		}
+		
+		if (groupMemberComponent.hasInactive()) {
+			existingCohort.setVoided(groupMemberComponent.getInactive());
```
ibacher, line 90: I think we also need to set the `voidReason` value when voiding a value.

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/GroupMemberTranslatorImpl_2_1.java
```diff
@@ -0,0 +1,95 @@
...
+	@Override
+	public Group.GroupMemberComponent toFhirResource(@Nonnull CohortMembership cohortMember) {
+		notNull(cohortMember, "CohortMember object should not be null");
+		Group.GroupMemberComponent groupMemberComponent = new Group.GroupMemberComponent();
+		groupMemberComponent.setId(cohortMember.getUuid());
+		groupMemberComponent.setInactive(!cohortMember.isActive());
+		
+		Patient patient = patientDao.getPatientById(cohortMember.getPatientId());
+		if (patient != null) {
+			groupMemberComponent.setEntityTarget(patientTranslator.toFhirResource(patient));
```
ibacher, line 55: I'm not sure I quite understand what this gives us that we don't get from the next line.

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/GroupTranslatorImpl_2_1.java
```diff
@@ -0,0 +1,74 @@
...
+	@Autowired
+	private GroupMemberTranslator<CohortMembership> groupMemberTranslator;
+	
+	@Override
+	public Group toFhirResource(@Nonnull Cohort cohort) {
+		notNull(cohort, "Cohort object should not be null");
+		Group group = super.toFhirResource(cohort);
+		
+		Collection<CohortMembership> memberships = cohort.getMemberships();
+		log.info("Number of members " + memberships.size());
```
ibacher, line 47:
```suggestion
		log.info("Number of members {}", memberships.size());
```

api/src/main/java/org/openmrs/module/fhir2/api/FhirPatientService.java
```diff
@@ -28,6 +30,10 @@
 	@Override
 	Patient get(@Nonnull String uuid);
 	
+	List<Patient> getByIds(Set<Integer> ids);
```
ibacher, line 33: Is there a reason to prefer `Set` over `Collection` here?

ibacher, line 33: We should probably have a `@Nonnull` here too.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirPatientServiceImpl.java
```diff
@@ -53,6 +58,18 @@
 	@Autowired
 	private SearchQuery<org.openmrs.Patient, Patient, FhirPatientDao, PatientTranslator, SearchQueryInclude<Patient>> searchQuery;
 	
+	@Override
+	public List<Patient> getByIds(Set<Integer> ids) {
+		List<Patient> patients = new ArrayList<>();
+		ids.forEach(id -> patients.add(translator.toFhirResource(dao.getPatientById(id))));
```
ibacher, line 64: I'd suggest that it's probably better to create a `dao.getPatientsByIds()` method that will directly take the collection of ids and return the matching result. That would generally be a lot more performant for the normal case.

test-data/src/test/resources/org/openmrs/module/fhir2/api/dao/impl/FhirCohortDaoImplTest_initial_data.xml
```diff
@@ -0,0 +1,14 @@
...
+    obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+    the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+
+    Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+    graphic logo is a trademark of OpenMRS Inc.
+-->
+<dataset>
+    <cohort cohort_id="1" name="John's patientList" description="cohort voided" creator="1" date_created="2005-01-01 00:00:00.0" voided="true" voided_by="1" date_voided="2005-01-01 00:00:00.0" void_reason="This is voided" uuid="985ff1a2-c2ef-49fd-836f-8a1d936d9ef9"/>
+    <cohort cohort_id="2" name="Covid19 patients" description="Covid19 patients" creator="1" date_created="2005-01-01 00:00:00.0" voided="false" uuid="1d64befb-3b2e-48e5-85f5-353d43e23e46"/>
+</dataset>
```
ibacher, line 14:
```suggestion
</dataset>

```

ibacher (commented):
Looks like a decent approach. I wish we didn't have to have the dependency on the DAO object in the translator, but I can't think of a good way around that issue.

api/src/main/java/org/openmrs/module/fhir2/api/FhirPatientService.java
corneliouzbett: Yes. To avoid duplicates. `Set` enforces the *set logic* - no duplicates & unordered

api-2.1/src/main/java/org/openmrs/module/fhir2/api/translators/impl/GroupMemberTranslatorImpl_2_1.java
corneliouzbett: Oh no, that shouldn't be there.

api/src/main/java/org/openmrs/module/fhir2/api/FhirPatientService.java
ibacher: Well, so where I'm going with this is:

* We've used `Collection` for other similar places
* The `Restrictions.in()` takes a `Collection` rather than a set.
* The de-duplication isn't exactly necessary here (since `SELECT * FROM patients WHERE patient_id IN (1, 2, 3, 1)` has the exact same result as `SELECT * FROM patients WHERE patient_id IN (1, 2, 3)`

Either way, it's not a big deal, just a question.

ibacher (approved):

PR #328, opened 2021-02-15
FM2-130:Adding  support for the FHIR Media resource

tendomart:
## Issue I worked on

  No? -> execute above command

@ibacher  @pmanko  

https://issues.openmrs.org/browse/FM2-130

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
@@ -279,4 +279,6 @@ public class FhirConstants {
 	public static final String REVERSE_INCLUDE_SEARCH_HANDLER = "_revinclude.search.handler";
 	
 	public static final String CONDITION_OBSERVATION_CONCEPT_UUID = "1284AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
+
+	public  static final String FHIR_MEDIA = "Media";
 }
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMediaDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMediaDaoImpl.java
@@ -1,0 +1,78 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.dao.impl;
+
+import lombok.AccessLevel;
+import lombok.Setter;
+import org.hibernate.Criteria;
+import org.openmrs.Obs;
+import org.openmrs.api.APIException;
+import org.openmrs.module.fhir2.FhirConstants;
+import org.openmrs.module.fhir2.api.dao.FhirMediaDao;
+import org.openmrs.module.fhir2.api.search.param.SearchParameterMap;
+import org.openmrs.obs.ComplexObsHandler;
+import org.springframework.stereotype.Component;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+
+@Component
+@Setter(AccessLevel.PACKAGE)
+public class FhirMediaDaoImpl extends BaseFhirDao<Obs> implements FhirMediaDao, ComplexObsHandler {
+
+    @Override
+    protected void setupSearchParams(Criteria criteria, SearchParameterMap theParams) {
+        theParams.getParameters().forEach(entry ->{
+            switch (entry.getKey()) {
+
+            }
+        });
+    }
+
+    @Override
+    public Obs saveObs(Obs obs) throws APIException {
+        return null;
+    }
+
+    @Override
+    public Obs getObs(Obs obs, String s) {
+        return null;
+    }
+
+    @Override
+    public boolean purgeComplexData(Obs obs) {
+        return false;
+    }
+
+    @Override
+    public String[] getSupportedViews() {
+        return new String[0];
+    }
+
+    @Override
+    public boolean supportsView(String s) {
+        return false;
+    }
+
+    @Override
+    public Obs createOrUpdate(@Nonnull Obs newEntry) {
+        return super.createOrUpdate(newEntry);
+    }
+
+    @Override
+    public Obs delete(@Nonnull String uuid) {
+        return super.delete(uuid);
+    }
+
+    @Override
+    public List<String> getSearchResultUuids(@Nonnull SearchParameterMap theParams) {
+        return super.getSearchResultUuids(theParams);
+    }
+}
--- a/api/src/main/resources/moduleApplicationContext.xml
+++ b/api/src/main/resources/moduleApplicationContext.xml
@@ -35,4 +35,13 @@
     <bean name="fhirR4" class="ca.uhn.fhir.context.FhirContext" factory-method="forR4">
         <property name="restfulClientFactory" ref="restfulClientFactoryR4"/>
     </bean>
+    <bean name="handler" class="org.openmrs.module.fhir2.api.handler.MediaComplexObsHandler">
+        <property name="handlers">
+            <map>
+                <entry>
+                    <key><value>mediaHandler</value></key>
+                </entry>
+            </map>
+        </property>
+    </bean>
 </beans>
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/FhirMediaDao.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/FhirMediaDao.java
@@ -1,0 +1,44 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.dao;
+
+import lombok.AccessLevel;
+import lombok.Setter;
+import org.openmrs.Obs;
+
+import org.openmrs.annotation.Authorized;
+import org.openmrs.module.fhir2.api.search.param.SearchParameterMap;
+import org.openmrs.obs.ComplexObsHandler;
+import org.openmrs.util.PrivilegeConstants;
+import org.springframework.stereotype.Component;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+
+public interface FhirMediaDao extends FhirDao<Obs> {
+
+    @Override
+    @Authorized(PrivilegeConstants.GET_OBS)
+    Obs get(@Nonnull String uuid);
+
+
+    @Override
+    @Authorized(PrivilegeConstants.ADD_OBS)
+    Obs createOrUpdate(@Nonnull Obs newEntry);
+
+    @Override
+    @Authorized(PrivilegeConstants.DELETE_OBS)
+    Obs delete(@Nonnull String uuid);
+
+    @Override
+    @Authorized(PrivilegeConstants.GET_OBS)
+    List<String> getSearchResultUuids(@Nonnull SearchParameterMap theParams);
+
+}
--- a/api/src/main/java/org/openmrs/module/fhir2/api/handler/MediaComplexObsHandler.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/handler/MediaComplexObsHandler.java
@@ -1,0 +1,82 @@
+
+/**
+ * The contents of this file are subject to the OpenMRS Public License
+ * Version 1.0 (the "License"); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the License at
+ * http://license.openmrs.org
+ *
+ * Software distributed under the License is distributed on an "AS IS"
+ * basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See the
+ * License for the specific language governing rights and limitations
+ * under the License.
+ *
+ * Copyright (C) OpenMRS, LLC.  All Rights Reserved.
+ */
+package org.openmrs.module.fhir2.api.handler;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.openmrs.Obs;
+import org.openmrs.Patient;
+import org.openmrs.api.APIException;
+import org.openmrs.api.PatientService;
+import org.openmrs.api.context.Context;
+import org.openmrs.customdatatype.CustomDatatypeHandler;
+import org.openmrs.obs.ComplexData;
+import org.openmrs.obs.ComplexObsHandler;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+@Component
+@Order(Ordered.LOWEST_PRECEDENCE)
+public class MediaComplexObsHandler implements ComplexObsHandler, CustomDatatypeHandler {
+
+    public static final Log log = LogFactory.getLog(Obs.class);
+
+    public static final String HANDLER_TYPE = "MediaHandler";
+
+    public MediaComplexObsHandler(){
+        super();
+    }
+
+    @Override
+    public Obs saveObs(Obs obs) throws APIException {
+        PatientService patientService = Context.getPatientService();
+        Patient patient = patientService.getPatient(Integer.parseInt(obs.getValueComplex()));
+        if(patient == null){
+            throw new APIException("Cannot save complex obs where obsId=" + obs.getObsId() + "Desired Patient id :"
+            +Integer.parseInt(obs.getValueComplex()) + "cannot be found");
+        }
+        
+        ComplexData complexdata = obs.getComplexData();
+        obs.setValueComplex(complexdata.getTitle());
+
+        return null;
+    }
+
+    @Override
+    public Obs getObs(Obs obs, String s) {
+        return null;
+    }
+
+    @Override
+    public boolean purgeComplexData(Obs obs) {
+        return true;
+    }
+
+    @Override
+    public String[] getSupportedViews() {
+        return new String[0];
+    }
+
+    @Override
+    public boolean supportsView(String s) {
+        return false;
+    }
+
+    @Override
+    public void setHandlerConfiguration(String s) {
+
+    }
+}


tendomart:
@ibacher  @pmanko  

1 .I just want to know wether this Handler has the required method implementation for doing  the required business. 

2.Would it make sense to design multiple ComplexObsHandler implementations for various MIME Types ?

api/src/main/java/org/openmrs/module/fhir2/api/FhirMediaService.java
```diff
@@ -0,0 +1,33 @@
...
+import ca.uhn.fhir.rest.api.server.IBundleProvider;
+import ca.uhn.fhir.rest.param.DateRangeParam;
+import ca.uhn.fhir.rest.param.ReferenceAndListParam;
+import ca.uhn.fhir.rest.param.StringAndListParam;
+import ca.uhn.fhir.rest.param.TokenAndListParam;
+import org.hl7.fhir.r4.model.Observation;
+
+public interface FhirMediaService extends FhirService<Observation> {
+	
+	Observation get(@Nonnull String uuid);
```
ibacher, line 27: This seems to be unnecessary as it's already defined by the parent interface.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMediaDaoImpl.java
```diff
@@ -0,0 +1,107 @@
...
+import org.openmrs.api.APIException;
+import org.openmrs.module.fhir2.FhirConstants;
+import org.openmrs.module.fhir2.api.dao.FhirMediaDao;
+import org.openmrs.module.fhir2.api.search.param.SearchParameterMap;
+import org.openmrs.obs.ComplexObsHandler;
+import org.springframework.stereotype.Component;
+
+@Component
+@Setter(AccessLevel.PACKAGE)
+public class FhirMediaDaoImpl extends BaseFhirDao<Obs> implements FhirMediaDao, ComplexObsHandler {
```
ibacher, line 29: This class definitely shouldn't be a `ComplexObsHandler`. That interface is only for classes which implement a specific type of `ComplexObs`, e.g., `ImageHandler` or `PdfHandler` or something like that.

api/src/main/java/org/openmrs/module/fhir2/api/handler/FhirMediaComplexObsHandler.java
```diff
@@ -0,0 +1,115 @@
...
+import org.openmrs.obs.ComplexData;
+import org.openmrs.obs.ComplexObsHandler;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+@Component
+@Setter(AccessLevel.PUBLIC)
+@Order(Ordered.LOWEST_PRECEDENCE)
+public class FhirMediaComplexObsHandler implements ComplexObsHandler, CustomDatatypeHandler {
```
ibacher, line 32: We really, really don't need a FHIR-specific `ComplexObsHandler` we can rely on the `ObsService` to populate the `ComplexObs` data (if any)

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirMediaServiceImpl.java
```diff
@@ -0,0 +1,110 @@
...
+import org.openmrs.module.fhir2.api.translators.MediaTranslator;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.annotation.Transactional;
+
+@Component
+@Transactional
+@Setter(AccessLevel.PACKAGE)
+@Getter(AccessLevel.PROTECTED)
+public class FhirMediaServiceImpl implements FhirMediaService {
```
ibacher, line 46: Is there a reason this doesn't extend `BaseFhirService` as that provides default implementations for all of the methods except search.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/MediaTranslatorImpl.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import javax.annotation.Nonnull;
+
+import static org.apache.commons.lang.Validate.notNull;
+
+@Component
+@Setter(AccessLevel.PACKAGE)
+public class MediaTranslatorImpl implements MediaTranslator {
```
ibacher, line 18: For the mappings to be added here, look at the ticket and the way we generally map those properties. Most of them have overlap with Obs, so the `ObservationTranslatorImpl` class would be a good place to start.

api/src/main/java/org/openmrs/module/fhir2/providers/r4/MediaFhirResourceProvider.java
```diff
@@ -0,0 +1,34 @@
...
+import org.hl7.fhir.r4.model.Media;
+import org.openmrs.module.fhir2.api.FhirMediaService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.stereotype.Component;
+
+@Component("mediaFhirR4ResourceProvider")
+@Qualifier("fhirResources")
+@Setter(AccessLevel.PACKAGE)
+public class MediaFhirResourceProvider implements IResourceProvider {
```
ibacher, line 25: We need some actual resource methods here, similar to the other `FhirResourceProvider` classes.

api/src/main/resources/moduleApplicationContext.xml
```diff
@@ -35,4 +35,13 @@
...
     </bean>
+    <bean name="handler" class="org.openmrs.module.fhir2.api.handler.FhirMediaComplexObsHandler">
+        <property name="handlers">
+            <map>
+                <entry>
+                    <key><value>mediaHandler</value></key>
+                </entry>
+            </map>
+        </property>
+    </bean>
```
ibacher, line 46:
```suggestion
```

ibacher (changes requested):
@tendomart Thanks for this! I've left you some comments here and some additional pointers on the Jira issue. Please read through them and let me know if there are any points you need me to clarify, etc.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirMediaServiceImpl.java
```diff
@@ -0,0 +1,82 @@
...
+		
+		SearchParameterMap theParams = new SearchParameterMap().addParameter(FhirConstants.MEDIA_STATUS, status)
+		        .addParameter(FhirConstants.MEDIA_TYPE, type).addParameter(FhirConstants.MEDIA_SUBJECT, subject)
+		        .addParameter(FhirConstants.MEDIA_ENCOUNTER_REFERENCE, encounterReference)
+		        .addParameter(FhirConstants.MEDIA_CREATED_DATE_TIME, createdDateTime)
+		        .addParameter(FhirConstants.MEDIA_CONTENT_TYPE, contentType)
+		        .addParameter(FhirConstants.CONTENT_DATA, contentDataType)
+		        .addParameter(FhirConstants.CONTENT_TITLE, contentTitle)
+		        .addParameter(FhirConstants.CONTENT_DATE_OF_CREATION, contentCreated)
+		        .addParameter(FhirConstants.COMMON_SEARCH_HANDLER, FhirConstants.ID_PROPERTY, id)
```
theanandankit, line 75: @tendomart `id` is not defined.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMediaDaoImpl.java
```diff
@@ -0,0 +1,144 @@
...
+		}
+	}
+
+	private void handleMediaCreatedDate(Criteria criteria, DateRangeParam mediaCreatedDate) {
+		if(mediaCreatedDate != null){
+			if(lacksAlias(criteria, "dt")){
+					criteria.createAlias("mediaCreatedDate", "dt");
+			}
+		}
+	}
```
corneliouzbett, line 110: This is superfluous. Ideally, you should re-use already existing functionality. i.e instead of this 👇🏾
```java
case FhirConstants.MEDIA_CREATED_DATE_TIME:
     entry.getValue().forEach(createdDateTime -> handleDateRange(createdDateTime.getPropertyName(),
	(DateRangeParam) createdDateTime.getParam()).ifPresent(criteria::add));
    break;
```
re-use `handleDateRange()` which somehow you are using it. Probably, you should replace with;
```java
case FhirConstants.DATE_RANGE_SEARCH_HANDLER:
    entry.getValue().forEach(dateRangeParam -> handleDateRange(dateRangeParam.getPropertyName(),
	(DateRangeParam) dateRangeParam.getParam()).ifPresent(criteria::add));
    break;
```
No need for the new constant and new method `handleMediaCreatedDate()`. This applies to `encounterReference` and the `subjectReference` - patient reference. Try re-using already existing functionality as much as possible.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDao.java
```diff
@@ -156,7 +156,7 @@
  *     AND crt.code in (?, ?, ?)
  *   ) OR (
  *         crt.concept_source_id = (select concept_source_id from fhir_concept_source where url = ?)
- *     AND crt.code = ?
+ *     AND crt.code = ?fhir_concept_source
```
corneliouzbett, line 159: This doesn't seem right to me. Why did you change the example query?

api/src/main/java/org/openmrs/module/fhir2/api/search/SearchQueryInclude.java
```diff
@@ -24,17 +24,7 @@
...
-import org.hl7.fhir.r4.model.DiagnosticReport;
-import org.hl7.fhir.r4.model.Encounter;
-import org.hl7.fhir.r4.model.Location;
-import org.hl7.fhir.r4.model.MedicationRequest;
-import org.hl7.fhir.r4.model.Observation;
-import org.hl7.fhir.r4.model.Person;
-import org.hl7.fhir.r4.model.Reference;
-import org.hl7.fhir.r4.model.RelatedPerson;
-import org.hl7.fhir.r4.model.ServiceRequest;
+import org.hl7.fhir.r4.model.*;
```
corneliouzbett, line 27: Avoid the use of import wildcards. I think you should change your IDE settings for Java imports

corneliouzbett (changes requested):

tendomart:
oops true .Thanks @corneliouzbett  this happened in my last commit yesterday . Prior to that all was well .

api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
```diff
@@ -296,4 +298,25 @@
...
+	public static final String MEDIA_CREATED_ON = "media.created.on";
+	
+	public static final String MEDIA_CONTENT_TYPE = "media.content.type";
+	
+	public static final String CONTENT_DATA = "content.data";
+	
+	public static final String CONTENT_TITLE = "content.title";
+	
+	public static final String CONTENT_DATE_OF_CREATION = "content.creation";
+
```
ibacher, line 321: We probably don't need all these constants, since we're generally not parsing the JSON ourselves. We usually just need the resource name as a constant, e.g., `Media`

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirMediaServiceImpl.java
```diff
@@ -0,0 +1,83 @@
...
+import org.openmrs.module.fhir2.api.translators.MediaTranslator;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.annotation.Transactional;
+
+@Component
+@Transactional
+@Setter(AccessLevel.PACKAGE)
+@Getter(AccessLevel.PROTECTED)
+public class FhirMediaServiceImpl extends BaseFhirService<Observation, Obs> implements FhirMediaService {
```
ibacher, line 43: The problem here is that you've declared this class to be `BaseFhirService<Observation, Obs>`, which implies that it's a service connecting between OMRS `Obs` and FHIR `Observation`s, but the translator (correctly) declares itself to be a translator between OMRS `Obs` and FHIR `Media`, so you get a type mismatch.

Change this to `BaseFhirService<Media, Obs>` and the compile error should go away. should work.

tendomart:
Thanks @ibacher so then what will  we use to map the remaining searchParams as most of the Fhir Media Params have no direct correlations with the Openmrs Obs and no handler implementations from fhir as well .

ibacher:
@tendomart Hopefully it's pretty straight-forward to create `handle...()` methods modelled on those from Obs or elsewhere in the code base.

tendomart:
Alright thanks alot .Sorry for late response , been doing some house cleaning on my Jira Dashboard .

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/MediaTranslatorImplTest.java
```diff
@@ -0,0 +1,54 @@
...
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.translators.impl;
+
+import static org.hamcrest.Matchers.notNullValue;
+import static org.springframework.test.util.MatcherAssertionErrors.assertThat;
```
ibacher, line 13:
```suggestion
import static org.hamcrest.MatcherAssert.assertThat;
```

api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImplTest.java
```diff
@@ -10,9 +10,7 @@
 package org.openmrs.module.fhir2.api.dao.impl;
 
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.notNullValue;
-import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.Matchers.*;
```
ibacher, line 13: Don't use star imports

api/src/main/java/org/openmrs/module/fhir2/providers/r3/MediaFhirResourceProvider.java
```diff
@@ -0,0 +1,28 @@
...
+package org.openmrs.module.fhir2.providers.r3;
+
+import ca.uhn.fhir.rest.server.IResourceProvider;
+import lombok.AccessLevel;
+import lombok.Setter;
+import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.stereotype.Component;
+
+@Component("mediaFhirR4ResourceProvider")
```
ibacher, line 19: This needs to have a different name, e.g., `mediaFhirR3ResourceProvider`. This is one of the things causing testing errors.

api/src/main/java/org/openmrs/module/fhir2/providers/r3/MediaFhirResourceProvider.java
```diff
@@ -0,0 +1,28 @@
...
+@Qualifier("fhirResources")
```
ibacher, line 20:
```suggestion
@R3Provider
```

api/src/main/java/org/openmrs/module/fhir2/providers/r4/MediaFhirResourceProvider.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.Setter;
+import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.hl7.fhir.r4.model.Media;
+import org.openmrs.module.fhir2.api.FhirMediaService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.stereotype.Component;
+
+@Component("mediaFhirR4ResourceProvider")
+@Qualifier("fhirResources")
```
ibacher, line 23:
```suggestion
@R4Provider
```

tendomart:
@ibacher  I just want to ask  ,is this the new way of , to replace  `@Qualifier("fhirResources")` with  `@R4Provider`

tendomart:
Hi @ibacher  i have tried to look around for a solution but failed to figure out why

	```
@Override
	public Obs get(@Nonnull String uuid) {
		return obsService.getObsByUuid(uuid);
	}
	
	@Override
	public Obs createOrUpdate(@Nonnull Obs newEntry) {
		return obsService.saveObs(newEntry, FhirConstants.SAVED_SUCCESSFULLY);
	}
	
	@Override
	public Obs delete(@Nonnull String uuid) {
		return super.delete(uuid);
	}
```
either throws a NPE or returns a Null Object when actually the same implementation in `FhirConceptDaoImpl` works .Any thoughts.

File: api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirMediaDaoImplTest.java [hunk 19]
```diff
@@ -0,0 +1,105 @@
...
+	@Before
+	public void setup() throws Exception {
+		executeDataSet(OBS_DATA_XML);
+		dao = new FhirMediaDaoImpl();
+		dao.setObsService(obsService);
+	}
+	
+	@Test
+	public void get_shouldGetComplexObsByUuid() {
+		assertThat(dao.get("OBS_UUID"), notNullValue());
```
ibacher, line 63:
```suggestion
		assertThat(dao.get(OBS_UUID), notNullValue());
```

api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirMediaDaoImplTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.test.context.ContextConfiguration;
+
+@ContextConfiguration(classes = TestFhirSpringConfiguration.class, inheritLocations = false)
+public class FhirMediaDaoImplTest extends BaseModuleContextSensitiveTest {
+	
+	private static final String OBS_DATA_XML = "org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImplTest_initial_data_suppl.xml";
+	
+	private static final String OBS_UUID = "759a0d9e-ccf8-4f00-a045-6a94c43fbd6b";
```
ibacher, line 44: Where did this UUID come from? In order to be able to load it from the database, it needs to be an Obs defined correctly in data that are loaded for this test. You probably want to ensure [this test data is loaded](https://github.com/openmrs/openmrs-core/blob/master/api/src/test/resources/org/openmrs/api/include/ObsServiceTest-complex.xml) and use the UUID from that.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMediaDaoImpl.java
```diff
@@ -0,0 +1,131 @@
...
+	
+	@Override
+	public List<Obs> getSearchResults(@Nonnull SearchParameterMap theParams, @Nonnull List<String> resourceUuids) {
+		return resourceUuids.stream().map(obsService::getObsByUuid).collect(Collectors.toList());
+	}
+	
+	@Override
+	protected void setupSearchParams(Criteria criteria, SearchParameterMap theParams) {
+		theParams.getParameters().forEach(entry -> {
```
ibacher, line 66: Here, we'll want to ensure that we're only searching ComplexObs and not just any Obs. This looks like it could be done as something like:

```java
if (lacksAlias(criteria, "c")) {
	criteria.createAlias("concept", "c");
}

if (lacksAlias(criteria, "cdt")) {
    criteria.createAlias("c.datatype", "cdt");
}

criteria.add(eq("cdt.hl7_abbreviation", "ED"));
```

tendomart:
Thanks @ibacher  for bringing this up . Iam still finding some trouble defining the search param . I have already identified them at https://www.hl7.org/fhir/DSTU2/media.html but i really don't understand the meaning of the aliases used here in any case .

Any clear documentation for this ?

ibacher:
@tendomart Sorry... I somehow missed the above question. The "c" and "cd" above are really just aliases for tables inside the criteria query. Just like in SQL you could write something like (this is similar to the query my code above is intended to generate):

```sql
select *
from concept c
join concept_data_type cdt on
  c.data_type_id = cdt.concept_data_type_id
where cdt.hl7_abbreviation = "ED"
```

Having an alias just allows us to refer to the property name in a shorter fashion.

The "ED" part actually comes from the code we use to determine whether or not a concept is a complex concept. See [here](https://github.com/openmrs/openmrs-core/blob/b7c03a28c64493023a8211c53bb60d4d944b39b7/api/src/main/java/org/openmrs/ConceptComplex.java#L83).

PR #334, opened 2021-03-16
FM2-352: Add support for creating encounters via the FHIR API

Medhavi-16:
## Issue I worked on

see https://issues.openmrs.org/browse/FM2-352

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
@@ -53,7 +53,7 @@ public class FhirConstants {
 	        + "/condition-ver-status";
 	
 	public static final String CLINICAL_FINDINGS_SYSTEM_URI = HL7_FHIR_CODE_SYSTEM_PREFIX + "/clinical-findings";
-
+	
 	public static final String TASK_STATUS_VALUE_SET_URI = "http://hl7.org/fhir/task-status";
 	
 	public static final String OBSERVATION_REFERENCE_RANGE_SYSTEM_URI = HL7_FHIR_CODE_SYSTEM_PREFIX
--- a/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
@@ -52,7 +52,7 @@ public class FhirConstants {
 	public static final String CONDITION_VERIFICATION_STATUS_SYSTEM_URI_R3 = HL7_FHIR_CODE_SYSTEM_PREFIX_R3
 	        + "/condition-ver-status";
 	
-	public static final String CLINICAL_FINDINGS_SYSTEM_URI = HL7_FHIR_CODE_SYSTEM_PREFIX + "/clinical-findings";	
+	public static final String CLINICAL_FINDINGS_SYSTEM_URI = HL7_FHIR_CODE_SYSTEM_PREFIX + "/clinical-findings";
 
 	public static final String TASK_STATUS_VALUE_SET_URI = "http://hl7.org/fhir/task-status";
 	
--- a/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
@@ -52,8 +52,8 @@ public class FhirConstants {
 	public static final String CONDITION_VERIFICATION_STATUS_SYSTEM_URI_R3 = HL7_FHIR_CODE_SYSTEM_PREFIX_R3
 	        + "/condition-ver-status";
 	
-	public static final String CLINICAL_FINDINGS_SYSTEM_URI = HL7_FHIR_CODE_SYSTEM_PREFIX + "/clinical-findings";
-	
+	public static final String CLINICAL_FINDINGS_SYSTEM_URI = HL7_FHIR_CODE_SYSTEM_PREFIX + "/clinical-findings";	
+
 	public static final String TASK_STATUS_VALUE_SET_URI = "http://hl7.org/fhir/task-status";
 	
 	public static final String OBSERVATION_REFERENCE_RANGE_SYSTEM_URI = HL7_FHIR_CODE_SYSTEM_PREFIX
--- a/api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
@@ -93,7 +93,6 @@ public class FhirEncounterServiceImpl extends BaseFhirService<Encounter, org.ope
 		}
 		
 		throw new InvalidRequestException("Invalid type of request");
-		
 	}
 	
 	@Override
@@ -136,7 +135,6 @@ public class FhirEncounterServiceImpl extends BaseFhirService<Encounter, org.ope
 		}
 		
 		return result;
-		
 	}
 	
 	@Override
--- a/api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
@@ -59,9 +59,6 @@ public class FhirEncounterServiceImpl extends BaseFhirService<Encounter, org.ope
 	
 	@Override
 	public Encounter get(@Nonnull String uuid) {
-		if (uuid == null) {
-			throw new InvalidRequestException("Uuid cannot be null.");
-		}
 		
 		Encounter result;
 		try {
--- a/api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
@@ -31,6 +31,7 @@ import org.openmrs.module.fhir2.api.search.SearchQuery;
 import org.openmrs.module.fhir2.api.search.SearchQueryInclude;
 import org.openmrs.module.fhir2.api.search.param.SearchParameterMap;
 import org.openmrs.module.fhir2.api.translators.EncounterTranslator;
+import org.openmrs.module.fhir2.api.util.FhirUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 import org.springframework.transaction.annotation.Transactional;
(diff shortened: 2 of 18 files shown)


Medhavi-16:
@ibacher I added support for adding encounters through FHIR API. PTAL.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
```diff
@@ -73,6 +70,57 @@ public Encounter get(@Nonnull String uuid) {
 		return result;
 	}
 	
+	@Override
+	public Encounter create(@Nonnull Encounter encounter) {
+		
+		if (encounter.getTypeFirstRep().getCodingFirstRep().getSystem() == null) {
+			throw new InvalidRequestException("Id type must be specified to update");
+		}
+		if (encounter.getTypeFirstRep().getCodingFirstRep().getSystem().equals(FhirConstants.ENCOUNTER_TYPE_SYSTEM_URI)) {
```
ibacher, line 79: So, rather than relying on the *first* type coding (especially since we want to ignore other coding types), I'd refactor this into a utility method, something like:

```java
private enum OpenmrsEncounterType {
	ENCOUNTER,
	VISIT,
	AMBIGUOUS
}

// ...

private Optional<OpenmrsEncounterType> getOpenmrsEncounterType(Encounter encounter) {
	List<OpenmrsEncounterType> openmrsEncounterTypes = encounter.getType().stream().flatMap(it -> 
		it.getCoding().stream()
			.filter(coding -> FhirConstants.ENCOUNTER_TYPE_SYSTEM_URI.equals(coding.getSystem()) ||
				FhirConstants.VISIT_TYPE_SYSTEM_URI.equals(coding.getSystem()))
			.map(coding -> {
				if (FhirConstants.ENCOUNTER_TYPE_SYSTEM_URI.equals(coding.getSystem())) {
					return OpenmrsEncounterType.ENCOUNTER;
				} else {
					return OpenmrsEncounterType.VISIT;
				}
			})
	).distinct().collect(Collectors.toList());
	
	if (openmrsEncounterTypes.isEmpty()) {
		return Optional.empty();
	} else if (openmrsEncounterTypes.size() == 1) {
		return Optional.of(openmrsEncounterTypes.get(0));
	} else {
		return Optional.of(OpenmrsEncounterType.AMBIGUOUS);
	}
}
```

Now as we end up with three states:

1. We known the encounter type of the encounter and can process it
2. No encounter type was supplied and we can throw an exception with a message to that effect
3. An encounter invalidly specified that it is *both* an OpenMRS Encounter and an OpenMRS visit and we can send an appropriate error message for that.

Plus, we don't depend on the order of the types or do anything else in response to codings we don't care about.

api/src/test/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImplTest.java
```diff
@@ -102,7 +106,12 @@
 	
 	@Before
 	public void setUp() {
-		encounterService = new FhirEncounterServiceImpl();
+		encounterService = new FhirEncounterServiceImpl() {
+			
+			@Override
+			protected void validateObject(Encounter object) {
```
ibacher, line 112: Why do we need to stub this out?

api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
```diff
@@ -50,7 +50,7 @@
 	        + "/condition-ver-status";
 	
 	public static final String CLINICAL_FINDINGS_SYSTEM_URI = HL7_FHIR_CODE_SYSTEM_PREFIX + "/clinical-findings";
-
+
```
ibacher, line 53: Can we exclude this change from this PR? It prevents it from interfering with other PRs that might make the same change.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
```diff
@@ -58,9 +58,6 @@
 	
 	@Override
 	public Encounter get(@Nonnull String uuid) {
-		if (uuid == null) {
```
ibacher, line 61: Why did you remove this check? Note that `@Nonnull` is a hint to the programmer; it doesn't guarantee that the value here is never null (and we should have the same checks we do [in the base class](https://github.com/openmrs/openmrs-module-fhir2/blob/master/api/src/main/java/org/openmrs/module/fhir2/api/impl/BaseFhirService.java)).

ibacher (changes requested):
@Medhavi-16 Thanks for this. It mostly looks good. I've added a few comments on some changes I'd like to see. In addition, it would be good to re-create the encounter related tests that were removed in [this commit](https://github.com/openmrs/openmrs-module-fhir2/commit/c6fc65a105b7a875cf9973966274c1880933c842#diff-d29233104de9d96ca6e8af497f647719400e78b850b206e4dddbbd5d74b22a76) or equivalents of them so we can be sure the changes actually work.

api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
Medhavi-16: @ibacher Yes definitely. I will remove it.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
Medhavi-16: @ibacher I mistook the Nonnull annotation. I will add it back to it and the base class.

ibacher: Perfectly understandable... My IDE also tells me those checks are useless, but sometimes you have to know better than your tools.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
Medhavi-16: @ibacher I will change the code accordingly.

Medhavi-16:
@ibacher  I made the said changes. PTAL.

theanandankit:
@Medhavi-16, thanks for this. I think you should add some more tests to cover all the changes.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
```diff
@@ -73,6 +74,74 @@ public Encounter get(@Nonnull String uuid) {
...
+		
+		throw new InvalidRequestException("Invalid type of request");
+	}
+	
+	@Override
+	public Encounter delete(@Nonnull String uuid) {
+		
+		if (uuid == null) {
+			throw new InvalidRequestException("Uuid cannot be null.");
+		}
```
theanandankit, line 131: Please cover this line with tests.

Medhavi-16: @theanandankit i will add the test.

Medhavi-16:
@ibacher I added the missing tests and changed things according to your review. PTAL.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
```diff
@@ -73,6 +74,74 @@ public Encounter get(@Nonnull String uuid) {
 		return result;
 	}
 	
+	@Override
+	public Encounter create(@Nonnull Encounter encounter) {
+
```
ibacher, line 79:
```suggestion
```

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
```diff
@@ -73,6 +74,74 @@ public Encounter get(@Nonnull String uuid) {
...
+		if (result.equals(FhirUtils.OpenmrsEncounterType.ENCOUNTER)) {
+			return super.create(encounter);
+		}
+		
+		if (result.equals(FhirUtils.OpenmrsEncounterType.VISIT)) {
+			return visitService.create(encounter);
+		}
+		
+		throw new InvalidRequestException("Invalid type of request");
+
```
ibacher, line 99:
```suggestion
```

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
ibacher: Can you please generally remove the blank lines at the start and end of functions?

api/src/main/java/org/openmrs/module/fhir2/api/util/FhirUtils.java
```diff
@@ -100,4 +110,25 @@ public static OperationOutcome createExceptionErrorOperationOutcome(String diagn
 		issue.setDiagnostics(diagnostics);
 		return outcome;
 	}
+	
+	public static Optional<OpenmrsEncounterType> getOpenmrsEncounterType(Encounter encounter) {
+		List<OpenmrsEncounterType> openmrsEncounterTypes = encounter.getType().stream().flatMap(
+		    it -> it.getCoding().stream().filter(coding -> FhirConstants.ENCOUNTER_TYPE_SYSTEM_URI.equals(coding.getSystem())
```
ibacher, line 116: Whitespace:

```suggestion
			it -> it.getCoding().stream().filter(coding -> FhirConstants.ENCOUNTER_TYPE_SYSTEM_URI.equals(coding.getSystem())
```

ibacher (changes requested):
@Medhavi-16 Thanks for this. First of all, I think there's a slight conflict with the other PR I just merged, so if you could fix that, that'd be great.

Also, we need a few more tests here. Yes, they're a little repetitive, but it makes it easier for us to know when and where we're breaking something, so ideally, we'd have some tests in the OMOD for the resource providers in addition to the integration tests and service-layer tests.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
Medhavi-16: @ibacher I will remove them.

ibacher:
@Medhavi-16 Do you need any help with this ticket? Thanks again!

Medhavi-16:
> @Medhavi-16 Do you need any help with this ticket? Thanks again!

@ibacher I am waiting for FM2-366 to be resolved and merged. Without that I am facing the error `failed to validate with reason: encounterDatetime: Encounter.datetime.required` in the OMOD tests

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r4/EncounterFhirResourceProviderIntegrationTest.java
```diff
@@ -347,4 +358,101 @@ public void shouldReturnCountForEncounterAsXml() throws Exception {
...
+		assertThat(response, isOk());
+		
+		Encounter newEncounter = readResponse(response);
+		
+		assertThat(newEncounter.getId(), equalTo(encounter.getId()));
+		
+	}
+	
+	@Test
+	public void shouldCreateEncounterFromOpenMrsVisitAsJson() throws Exception {
```
Medhavi-16, line 425: @ibacher This is the integration test for creating Visit type Encounter

ibacher: @Medhavi-16: I'd wager the error is caused by [this attribute type](https://github.com/openmrs/openmrs-module-fhir2/blob/16ff04925718bc6b0c718bef0439f48751e81bf9/test-data/src/test/resources/org/openmrs/module/fhir2/api/dao/impl/FhirVisitDaoImplTest_initial_data.xml#L22). Can we try not loading that file and finding another existing visit in the standard data set to work with>?

Medhavi-16: @ibacher Thank you! Yes I'll try to use the standard dataset.

Medhavi-16:
@ibacher PTAL.

ibacher (approved):
@Medhavi-16 Thanks for this! LGTM!

PR #335, opened 2021-03-20
FM2-347:Add support for _has and _has:...:not for ServiceRequest and …

gitcliff:
LINK:https://issues.openmrs.org/browse/FM2-347


api/src/main/java/org/openmrs/module/fhir2/api/FhirServiceRequestService.java
```diff
@@ -14,13 +14,14 @@
...
 import ca.uhn.fhir.rest.param.DateRangeParam;
+import ca.uhn.fhir.rest.param.HasParam;
 import ca.uhn.fhir.rest.param.ReferenceAndListParam;
 import ca.uhn.fhir.rest.param.TokenAndListParam;
 import org.hl7.fhir.r4.model.ServiceRequest;
 
 public interface FhirServiceRequestService extends FhirService<ServiceRequest> {
 	
-	IBundleProvider searchForServiceRequests(ReferenceAndListParam patientReference, TokenAndListParam code,
+	IBundleProvider searchForServiceRequests(ReferenceAndListParam patientReference, TokenAndListParam code, HasParam has,
```
ibacher, line 24: nit: I'd prefer that the `has` param is near the end around where `includes` are (they are somewhat similar ideas.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirServiceRequestServiceImpl.java
```diff
@@ -52,12 +53,12 @@
...
 	public IBundleProvider searchForServiceRequests(ReferenceAndListParam patientReference, TokenAndListParam code,
-	        ReferenceAndListParam encounterReference, ReferenceAndListParam participantReference, DateRangeParam occurrence,
-	        TokenAndListParam uuid, DateRangeParam lastUpdated, HashSet<Include> includes) {
+	        HasParam has, ReferenceAndListParam encounterReference, ReferenceAndListParam participantReference,
+	        DateRangeParam occurrence, TokenAndListParam uuid, DateRangeParam lastUpdated, HashSet<Include> includes) {
 		
 		SearchParameterMap theParams = new SearchParameterMap()
 		        .addParameter(FhirConstants.PATIENT_REFERENCE_SEARCH_HANDLER, patientReference)
-		        .addParameter(FhirConstants.CODED_SEARCH_HANDLER, code)
+		        .addParameter(FhirConstants.CODED_SEARCH_HANDLER, code).addParameter(FhirConstants.HAS_SEARCH_HANDLER, has)
```
ibacher, line 61:
```suggestion
		        .addParameter(FhirConstants.CODED_SEARCH_HANDLER, code)
		        .addParameter(FhirConstants.HAS_SEARCH_HANDLER, has)
```

ibacher: nit: I would also move this near the `includes` handling

api/src/main/java/org/openmrs/module/fhir2/providers/r3/ProcedureRequestFhirResourceProvider.java
```diff
@@ -113,7 +114,7 @@ public IBundleProvider searchForProcedureRequests(
 	                Practitioner.SP_NAME }, targetTypes = Practitioner.class) ReferenceAndListParam participantReference,
 	        @OptionalParam(name = ProcedureRequest.SP_OCCURRENCE) DateRangeParam occurrence,
 	        @OptionalParam(name = ProcedureRequest.SP_RES_ID) TokenAndListParam uuid,
-	        @OptionalParam(name = "_lastUpdated") DateRangeParam lastUpdated,
+	        @OptionalParam(name = "_lastUpdated") DateRangeParam lastUpdated, @OptionalParam(name = "_has") HasParam has,
```
ibacher, line 117:
```suggestion
	        @OptionalParam(name = "_lastUpdated") DateRangeParam lastUpdated,
	        @OptionalParam(name = "_has") HasParam has,
```

api/src/main/java/org/openmrs/module/fhir2/providers/r4/ServiceRequestFhirResourceProvider.java
```diff
@@ -108,8 +119,9 @@ public IBundleProvider searchForServiceRequests(
 	                Practitioner.SP_GIVEN, Practitioner.SP_FAMILY,
 	                Practitioner.SP_NAME }, targetTypes = Practitioner.class) ReferenceAndListParam participantReference,
 	        @OptionalParam(name = ServiceRequest.SP_OCCURRENCE) DateRangeParam occurrence,
+	        
 	        @OptionalParam(name = ServiceRequest.SP_RES_ID) TokenAndListParam uuid,
-	        @OptionalParam(name = "_lastUpdated") DateRangeParam lastUpdated,
+	        @OptionalParam(name = "_lastUpdated") DateRangeParam lastUpdated, @OptionalParam(name = "_has") HasParam has,
```
ibacher, line 124:
```suggestion
	        @OptionalParam(name = "_lastUpdated") DateRangeParam lastUpdated,
	        @OptionalParam(name = "_has") HasParam has,
```

api/src/main/java/org/openmrs/module/fhir2/providers/r4/ServiceRequestFhirResourceProvider.java
```diff
@@ -49,6 +58,8 @@
 @Setter(AccessLevel.PACKAGE)
 public class ServiceRequestFhirResourceProvider implements IResourceProvider {
 	
+	private final FhirContext cntxt = FhirContext.forR4();
```
ibacher, line 61: Don't do this. If you really need a FHIRContext, you can get one from Spring:

```java
@Autowired
@Qualifier("fhirR4")
private FhirContext fhirContext;
```

api/src/main/java/org/openmrs/module/fhir2/providers/r4/ServiceRequestFhirResourceProvider.java
```diff
@@ -121,7 +133,45 @@ public IBundleProvider searchForServiceRequests(
...
 		
-		return serviceRequestService.searchForServiceRequests(patientReference, code, encounterReference,
+		hasParameter(has);
+		
+		return serviceRequestService.searchForServiceRequests(patientReference, code, has, encounterReference,
 		    participantReference, occurrence, uuid, lastUpdated, includes);
 	}
+	
+	private void hasParameter(HasParam has) {
+		List<List<IQueryParameterType>> theHasParameters = Lists.newArrayList();
```
ibacher, line 143: It seems like a bad idea to implement the `hasParameter` here. `_has` is basically asking for a JOIN, so it would be better to add this at the DAO level where we can just add the join. We also have some nifty ways to handle these embedded parameters without so many nested loops.

ibacher: I also don't see any real advantage of this over `new ArrayList<>()`...

api/src/main/java/org/openmrs/module/fhir2/providers/r4/ServiceRequestFhirResourceProvider.java
```diff
@@ -121,7 +133,45 @@ public IBundleProvider searchForServiceRequests(
...
+		for (List<? extends IQueryParameterType> nextOrList : theHasParameters) {
```
ibacher, line 144: Isn't this looping over the empty list we just created?

gitcliff:
@ibacher kindly review

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDao.java
```diff
@@ -443,6 +453,49 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
+		}
+		ArrayList<IQueryParameterType> orValues = new ArrayList<>();
+		String targetResourceType = null;
+		String paramReference = null;
+		String parameterName = null;
+		
+		String paramName = null;
+		targetResourceType = hasParam.getTargetResourceType();
+		
+		if (targetResourceType.getClass().equals(Observation.class)) {
```
ibacher, line 475: `targetResourceType` is a `String` so `targetResourceType.getClass()` will always return `java.lang.String`. What you actually want is:

```suggestion
		if (FhirConstants.OBSERVATION.equals(targetResourceType)) {
```

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDao.java
```diff
@@ -443,6 +453,49 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
 	}
 	
+	/**
+	 * A handler for a {@link HasParam}
+	 *
+	 * @param propertyName the name of the property in the query to use
+	 * @param hasParam the {@link HasParam} to handle
+	 * @return a {@link Criterion} to be added to the query for the indicate has param
+	 */
+	protected Optional<Criterion> handleHas(String propertyName, HasParam hasParam) {
```
ibacher, line 463: Since we're *only* defining `_has` for `ServiceRequest`, this function should just be in the `FhirServiceRequestDaoImpl` class.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDao.java
```diff
@@ -443,6 +453,49 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
+		if (hasParam == null) {
+			return Optional.empty();
+		}
+		ArrayList<IQueryParameterType> orValues = new ArrayList<>();
+		String targetResourceType = null;
+		String paramReference = null;
+		String parameterName = null;
+		
+		String paramName = null;
```
ibacher, line 472: I'm not sure we actually need any of these variables.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDao.java
```diff
@@ -443,6 +453,49 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
+		targetResourceType = hasParam.getTargetResourceType();
+		
+		if (targetResourceType.getClass().equals(Observation.class)) {
+			paramReference = hasParam.getReferenceFieldName();
+			parameterName = hasParam.getParameterName();
+			paramName = parameterName.replaceAll("\\..*", "");
```
ibacher, line 478: Can you explain what this is trying to solve? I'm not sure I get it.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDao.java
```diff
@@ -443,6 +453,49 @@ protected boolean containsAlias(Iterator<CriteriaImpl.Subcriteria> subcriteriaIt
...
+			hasParam = new HasParam(targetResourceType, paramReference, parameterName, hasParam.getParameterValue());
+			orValues.add(hasParam);
+			orValues.add(
+			    (IQueryParameterType) QualifiedParamList.singleton(null, hasParam.getValueAsQueryToken(fhirContext)));
+		}
+		String qualifier = paramName.substring(4);
+		hasParam.setValueAsQueryToken(fhirContext, FhirConstants.HAS_PROPERTY, qualifier,
+		    hasParam.getValueAsQueryToken(fhirContext));
+		orValues.add(hasParam);
+		return Optional.of(eq(propertyName, orValues));
```
ibacher, line 488: Does this actually work? It seems to me like this is never going to match.

ibacher:
PS I think the handling of the `_has` parameter should look something like this:

```java
private void handleHasAndParam(Criteria criteria, HasAndListParam hasAndListParam) {
		if (hasAndListParam == null) {
			return;
		}

		DetachedCriteria detachedCriteria = DetachedCriteria.forClass(Obs.class);

		hasAndListParam.getValuesAsQueryTokens().stream().map(this::handleHasOrParam).filter(Optional::isPresent)
				.flatMap(Optional::get).forEach(detachedCriteria::add);
		
		
		criteria.add(Subqueries.exists(detachedCriteria));
	}

	private Optional<Stream<Criterion>> handleHasOrParam(HasOrListParam hasOrListParam) {
		if (hasOrListParam == null) {
			return Optional.empty();
		}

		return Optional.of(hasOrListParam.getValuesAsQueryTokens().stream().map(this::handleHasParam).filter(
				Optional::isPresent).map(Optional::get));
	}

	private Optional<Criterion> handleHasParam(HasParam hasParam) {
		if (hasParam == null) {
			return Optional.empty();
		}
		
		if (!FhirConstants.OBSERVATION.equals(hasParam.getTargetResourceType())) {
			return Optional.empty();
		}
	}
```

(I've left out the mapping of the fields to the right criteria.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDao.java
gitcliff: true ,,made some changes

gitcliff:
@ibacher thanks for the outline ,,,
i have added the field mapping

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
```diff
@@ -160,6 +169,27 @@ private void handleValueCodedConcept(Criteria criteria, TokenAndListParam valueC
...
+		handleAndListParam(reference, param -> {
+			if (validReferenceParam(param)) {
+				if (lacksAlias(criteria, alias)) {
+					criteria.createAlias(property, alias);
+				}
+				
+				List<Optional<Criterion>> criterionList = new ArrayList<>();
+				criterionList.add(Optional.of(eq(String.format("%s.reference", alias), param.getIdPart())));
+				criterionList.add(Optional.of(eq(String.format("%s.type", alias), param.getResourceType())));
+				return Optional.of(and(toCriteriaArray(criterionList)));
```
ibacher, line 186: Do we actually need this?

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
```diff
@@ -59,6 +63,10 @@ protected void setupSearchParams(Criteria criteria, SearchParameterMap theParams
 					entry.getValue()
 					        .forEach(category -> handleConceptClass(criteria, (TokenAndListParam) category.getParam()));
 					break;
+				case FhirConstants.BASED_ON_REFERENCE_SEARCH_HANDLER:
```
ibacher, line 66: So, for `basedOnReferences` we don't want to add *that* to the Hibernate search, because Hibernate won't know what that means at all. What we need to do is tell Hibernate to join the relevant table. For `basedOn`, that would be the orders table. So, what we actually want to do here is ensure that we call something like:

```java
if (lacksCriteria(criteria, "ord") {
    criteria.createAlias("order", "ord");
}

// handle any properties of the order here
```

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
```diff
@@ -71,6 +79,7 @@ protected void setupSearchParams(Criteria criteria, SearchParameterMap theParams
 					entry.getValue().forEach(hasMemberReference -> handleHasMemberReference(criteria,
 					    (ReferenceAndListParam) hasMemberReference.getParam()));
 					break;
+
```
ibacher, line 82:
```suggestion
```

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
```diff
@@ -160,6 +169,27 @@ private void handleValueCodedConcept(Criteria criteria, TokenAndListParam valueC
 		}
 	}
 	
+	private Boolean validReferenceParam(ReferenceParam ref) {
+		return (ref != null && ref.getIdPart() != null && ref.getResourceType() != null);
+	}
+
```
ibacher, line 175: Is there any value to having this as a function? Also, if we leave the function in here, the return type should be `boolean` rather than `Boolean` (it's usually better to use primitives where possible).

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
ibacher: Also, note that these are being translated into an actual Hibernate query; is Hibernate going to know how to handle `basedOnReference.reference = <id part>`?

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirServiceRequestDaoImpl.java
```diff
@@ -83,5 +101,54 @@ private void handleCodedConcept(Criteria criteria, TokenAndListParam code) {
 		    Optional.of(or(toCriteriaArray(Stream.of(handleDate("dateStopped", dateRangeParam.getUpperBound()),
 		        handleDate("autoExpireDate", dateRangeParam.getUpperBound())))))))));
 	}
+
+	private void handleHasAndParam(Criteria criteria, HasAndListParam hasAndListParam) {
+		if (hasAndListParam != null) {
+		
+		if (lacksAlias(criteria, "obss")) {
```
ibacher, line 108: Why not `obs`?

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirServiceRequestDaoImpl.java
```diff
@@ -83,5 +101,54 @@ private void handleCodedConcept(Criteria criteria, TokenAndListParam code) {
...
+				criteria.createAlias("obs", "aliasObss");
```
ibacher, line 109: This alias should match the alias checked for in `lacksAlias` above.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirServiceRequestDaoImpl.java
```diff
@@ -83,5 +101,54 @@ private void handleCodedConcept(Criteria criteria, TokenAndListParam code) {
...
+			}
+
+		DetachedCriteria detachedCriteria = DetachedCriteria.forClass(Obs.class);
+
+		 handleAndListParam(hasAndListParam, param -> {
+			if ( param.getValueAsQueryToken(fhirContext) != null) {
+				if (lacksAlias(criteria, "o")) {
+					criteria.createAlias("aliasObss.category", "o");
```
ibacher, line 117: So this is going to try to generate something like:

```sql
INNER JOIN aliasObss_.category o
```

Which probably isn't going to work. Is there a way we could maybe call `setupSearchParams()` in the `FhirObsDao`?

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirServiceRequestDaoImpl.java
```diff
@@ -83,5 +101,54 @@ private void handleCodedConcept(Criteria criteria, TokenAndListParam code) {
...
+				}
+			}
+			
+			return Optional.of(eq("o.category", param.getValueAsQueryToken(fhirContext)));
```
ibacher, line 121: This is going to generate something like:

```sql
WHERE o.category = "category=blah"
```

ibacher (commented):
@gitcliff Sorry it took me so long to add some comments here. I've tried to do so, if you're willing to take another look. One thing that's clearly missing here is some tests for the methods in question, which would be ideal.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirServiceRequestDaoImpl.java
gitcliff: @ibacher are you talking about [setupSearchParam(](https://github.com/openmrs/openmrs-module-fhir2/blob/3a085ce12511f1efae4f99d6b6ed54f950975099/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java#L44)) in `FhirObservationImpl `

Secondly i think the `INNER JOIN aliasObss_.category o` can be avoided by doing somthing like 
```
if (lacksAlias(criteria, "obs")) {
	criteria.createAlias("obs", "obs1");
			
}
```

ibacher: So here's what I was thinking: the way in which we handle the FHIR `Observation.category` is a bit... unique (the implementation is [here](https://github.com/openmrs/openmrs-module-fhir2/blob/495c6666e51c1d2f7416a7b0c6f729fe8d8d98b6/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java#L134); note that it involves creating a subquery and then checking if the `Obs.conceptClass.uuid` field is in a set returned from `categoryMap.queryConceptClassByCategory()`. That's a lot of logic to replicate here. Plus, we might want to filter the `Obs` based on other properties rather than just `category`. So the question is is there some way we can reuse the logic that already exists in `FhirObservationDaoImpl` here?

ibacher: What we ultimately want to end up with is something like this query:

```sql
SELECT *
FROM drug_order do_
JOIN obs ob_ ON ob_.order_id = do_.order_id
JOIN concept c_ ON ob_.concept_id = c_.concept_id
JOIN concept_class cc_ on  c_.class_id = cc_.concept_class_id
WHERE cc_.uuid IN (?, ?, ?)
```

Or, somewhat equivalently:

```sql
SELECT *
FROM drug_order do_
WHERE do_.id IN (
    SELECT ob_.order_id
    FROM obs ob_
    JOIN concept c_ ON ob_.concept_id = c_.concept_id
    JOIN concept_class cc_ on  c_.class_id = cc_.concept_class_id
	WHERE cc_.uuid IN (?, ?, ?)
)
```

ibacher: I'm not exactly sure how to generate the first of those (it's a little tricky because the linkage is only one way), but the second could be created by something like:

```java
Criteria criteria = sessionFactory.getCurrentSession().createCriteria(DrugOrder.class);
DetachedCriteria criteria2 = DetachedCriteria.forClass(Obs.class, "obs");
criteria2.createAlias("concept", "c");
criteria2.createAlias("c.conceptClass", "cc");
criteria2.add(Subqueries.propertyIn("cc.uuid", categoryMap.queryConceptClassByCategory("laboratory")));
criteria.add(Subqueries.propertyIn("id", criteria2));
```

So the question is whether we can (somehow) use `FhirObservationDaoImpl#setupSearchParams()` to populate the criteria for `criteria2` above and somewhat simplify the code we need to implement in the `FhirServiceRequestDaoImpl` class.

gitcliff: hello @ibacher
am sorry for the delayed response ,,i have been looking into the commnets you made above and your suggestion, i would say implementing`  FhirObservationDaoImpl#setupSearchParams(`) in `FhirServiceRequestDaoImpl `would work since it will cover the Observation.category issue with the concept class mapping plus also catering the other filtering on the obs ie patient, encounter references something like `/ServiceRequest?_has:Observation:based-on:patient=uuid `(please correct me if am wrong on this). The tricky bit here has been implementing `FhirObservationDaoImpl#setupSearchParams()` after `Auto-wiring` it in the `FhirServiceRequestDaoImpl `class because of its void nature and taking in a `SearchParameterMap` argument

gitcliff:
hello @ibacher,,, i added some unit tests

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirServiceRequestDaoImpl.java
```diff
@@ -84,4 +103,70 @@ private void handleCodedConcept(Criteria criteria, TokenAndListParam code) {
 		        handleDate("autoExpireDate", dateRangeParam.getUpperBound())))))))));
 	}
 	
+	private void handleHasAndParam(Criteria criteria, HasAndListParam hasAndListParam) {
+		if (hasAndListParam == null) {
+			return;
+		}
+		
+		HasParam param = hasAndListParam.getValuesAsQueryTokens().get(0).getValuesAsQueryTokens().get(0);
```
ibacher, line 111: So, there's a quite a bit of refactoring that should be done around this. The reason there are so many handlers is really to try to simplify the code. So something like this might work better (this replaces basically `handleHasAndParam`, `handleHasOrParam`, and `handleHasParam`)

```java
	private void handleHasAndParam(Criteria criteria, HasAndListParam hasAndListParam) {
		if (hasAndListParam == null || hasAndListParam.size() == 0) {
			return;
		}

		DetachedCriteria obsSubquery = DetachedCriteria.forClass(Obs.class, "obs");
		handleAndListParam(hasAndListParam, hasParam -> handleHasParam(obsSubquery, hasParam)).ifPresent(obsSubquery::add);
		obsSubquery.setProjection(property("obs.order.orderId"));
		criteria.add(exists(obsSubquery));
	}

	private Optional<Criterion> handleHasParam(DetachedCriteria criteria, HasParam hasParam) {
		switch (hasParam.getParameterName()) {
			case Observation.SP_ENCOUNTER:
				if (lacksAlias(criteria, "e")) {
					criteria.createAlias("encounter", "e");
				}
				return Optional.of(eq("e.uuid", hasParam.getParameterValue()));
			break;
			case Observation.SP_PATIENT:
				// etc;
		}
	}
```

Note that this implementation relies on a couple of functions I added to this PR in `BaseDao`. This is intended to be something of a pattern for how things are handled, made just a bit more complicated by the need to do this for a subquery.

ibacher (commented):
@gitcliff Thanks for the update on this. I've provided some updated thoughts on the central function to this account. Weirdly, the unit tests still don't seem to be actually using this new code. Maybe a good idea to look into why that is. Anyway, thanks a lot! 😁

PR #353, opened 2021-06-18
FM2-400: Support for $lastn-encounters in R4

Medhavi-16:
## Issue I worked on

see https://issues.openmrs.org/browse/FM2-400

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
@@ -10,7 +10,7 @@
 package org.openmrs.module.fhir2.api.dao.impl;
 
 import static org.hibernate.criterion.Projections.property;
-import static org.openmrs.module.fhir2.api.util.LastnResults.getTopNRankedUuids;
+import static org.openmrs.module.fhir2.api.util.LastnOperationUtils.getTopNRankedUuids;
 
 import javax.annotation.Nonnull;
 
@@ -28,7 +28,7 @@ import org.openmrs.Encounter;
 import org.openmrs.module.fhir2.FhirConstants;
 import org.openmrs.module.fhir2.api.dao.FhirEncounterDao;
 import org.openmrs.module.fhir2.api.search.param.SearchParameterMap;
-import org.openmrs.module.fhir2.api.util.LastnResults;
+import org.openmrs.module.fhir2.api.util.LastnResult;
 import org.springframework.stereotype.Component;
 
 @Component
@@ -45,7 +45,7 @@ public class FhirEncounterDaoImpl extends BaseFhirDao<Encounter> implements Fhir
 			criteria.setProjection(Projections.projectionList().add(property("uuid")).add(property("encounterDatetime")));
 			
 			@SuppressWarnings("unchecked")
-			List<LastnResults> results = ((List<Object[]>) criteria.list()).stream().map(LastnResults::new)
+			List<LastnResult> results = ((List<Object[]>) criteria.list()).stream().map(LastnResult::new)
 			        .collect(Collectors.toList());
 			
 			return getTopNRankedUuids(results, getMaxParameter(theParams));
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
@@ -10,6 +10,7 @@
 package org.openmrs.module.fhir2.api.dao.impl;
 
 import static org.hibernate.criterion.Projections.property;
+import static org.openmrs.module.fhir2.providers.util.LastnResults.getTopNRankedUuids;
 
 import javax.annotation.Nonnull;
 
@@ -49,7 +50,7 @@ public class FhirEncounterDaoImpl extends BaseFhirDao<Encounter> implements Fhir
 			List<LastnResults> results = ((List<Object[]>) criteria.list()).stream().map(LastnResults::new)
 			        .collect(Collectors.toList());
 			
-			return LastnResults.getTopNRankedUuids(results, getMaxParameter(theParams));
+			return getTopNRankedUuids(results, getMaxParameter(theParams));
 		}
 		
 		return super.getSearchResultUuids(theParams);
@@ -85,36 +86,5 @@ public class FhirEncounterDaoImpl extends BaseFhirDao<Encounter> implements Fhir
 	private int getMaxParameter(SearchParameterMap theParams) {
 		return ((NumberParam) theParams.getParameters(FhirConstants.MAX_SEARCH_HANDLER).get(0).getParam()).getValue()
 		        .intValue();
-	}
-	
-	private List<String> getTopNRankedUuids(List<LastnResults> list, int max) {
-		
-		list.sort((a, b) -> b.getDatetime().compareTo(a.getDatetime()));
-		List<String> results = new ArrayList<>(list.size());
-		int currentRank = 0;
-		
-		for (int var = 0; var < list.size() && currentRank < max; var++) {
-			currentRank++;
-			results.add(list.get(var).getUuid());
-			Date currentEncounterDate = list.get(var).getDatetime();
-			
-			if (var == list.size() - 1) {
-				return results;
-			}
-			
-			//Adding all Encounters which have the same Datetime as the current Obs Datetime since they will have the same rank
-			Date nextEncounterDate = list.get(var + 1).getDatetime();
-			while (nextEncounterDate.equals(currentEncounterDate)) {
-				results.add(list.get(var + 1).getUuid());
-				var++;
-				
-				if (var + 1 == list.size()) {
-					return results;
-				}
-				nextEncounterDate = list.get(var + 1).getDatetime();
-			}
-		}
-		
-		return results;
 	}
 }
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
@@ -10,12 +10,10 @@
 package org.openmrs.module.fhir2.api.dao.impl;
 
 import static org.hibernate.criterion.Projections.property;
-import static org.openmrs.module.fhir2.providers.util.LastnResults.getTopNRankedUuids;
+import static org.openmrs.module.fhir2.api.util.LastnResults.getTopNRankedUuids;
 
 import javax.annotation.Nonnull;
 
-import java.util.ArrayList;
-import java.util.Date;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -30,7 +28,7 @@ import org.openmrs.Encounter;
 import org.openmrs.module.fhir2.FhirConstants;
 import org.openmrs.module.fhir2.api.dao.FhirEncounterDao;
 import org.openmrs.module.fhir2.api.search.param.SearchParameterMap;
-import org.openmrs.module.fhir2.providers.util.LastnResults;
+import org.openmrs.module.fhir2.api.util.LastnResults;
 import org.springframework.stereotype.Component;
 
 @Component
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
@@ -9,11 +9,23 @@
  */
 package org.openmrs.module.fhir2.api.dao.impl;
 
+import static org.hibernate.criterion.Projections.property;
+
+import javax.annotation.Nonnull;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.stream.Collectors;
+
 import ca.uhn.fhir.rest.param.DateRangeParam;
+import ca.uhn.fhir.rest.param.NumberParam;
 import ca.uhn.fhir.rest.param.ReferenceAndListParam;
 import lombok.AccessLevel;
 import lombok.Setter;
+import lombok.Value;
 import org.hibernate.Criteria;
+import org.hibernate.criterion.Projections;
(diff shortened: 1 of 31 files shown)


Medhavi-16:
@ibacher @varung-31 PTAL.

api/src/test/java/org/openmrs/module/fhir2/api/search/ObservationSearchQueryTest.java
```diff
@@ -1585,6 +1585,112 @@ public void searchForLastnObs_shouldHandleNoCodeRequest() {
...
+		        .addParameter(FhirConstants.LASTN_ENCOUNTERS_SEARCH_HANDLER, new StringParam());
+		
+		IBundleProvider results = search(theParams);
+		
+		assertThat(results, notNullValue());
+		List<IBaseResource> resultList = get(results);
+		
+		assertThat(resultList.size(), equalTo(2));
+		assertThat(resultList, everyItem(anyOf(allOf(is(instanceOf(Observation.class))))));
+
```
varung-31, line 1616: In the tests for search query, it might be worthwhile to check the number of distinct Encounter uuids related to the output Observations.
`observation.getEncounter().getReferenceElement().getIdPart()`

api/src/test/java/org/openmrs/module/fhir2/api/search/ObservationSearchQueryTest.java
```diff
@@ -1585,6 +1585,112 @@ public void searchForLastnObs_shouldHandleNoCodeRequest() {
...
+		
+		TokenAndListParam categories = new TokenAndListParam().addAnd(new TokenParam().setValue("laboratory"));
+		
+		TokenAndListParam code = new TokenAndListParam().addAnd(
+		    new TokenParam().setSystem(FhirTestConstants.LOINC_SYSTEM_URL).setValue(LOINC_SYSTOLIC_BP),
+		    new TokenParam().setSystem(FhirTestConstants.CIEL_SYSTEM_URN).setValue(OBS_CONCEPT_ID));
+		
+		SearchParameterMap theParams = new SearchParameterMap().addParameter(FhirConstants.CODED_SEARCH_HANDLER, code)
+		        .addParameter(FhirConstants.CATEGORY_SEARCH_HANDLER, categories)
+		        .addParameter(FhirConstants.MAX_SEARCH_HANDLER, new NumberParam(2))
```
varung-31, line 1605: One test can be added for max = 1 to test the difference in outputs.

varung-31 (changes requested):

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
```diff
@@ -69,6 +75,28 @@
...
+			Criteria criteria = getSessionFactory().getCurrentSession().createCriteria(typeToken.getRawType());
+			
+			setupSearchParams(criteria, theParams);
+			
+			criteria.setProjection(property("uuid"));
+			
+			@SuppressWarnings("unchecked")
+			List<String> lastnEncountersObservationsUuids = criteria.list();
+			
+			return lastnEncountersObservationsUuids;
```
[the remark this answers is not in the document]
Medhavi-16: Yes we can do that. I'll modify it.

api/src/test/java/org/openmrs/module/fhir2/api/search/ObservationSearchQueryTest.java
Medhavi-16: Yes, I'll add those assertions.

api/src/test/java/org/openmrs/module/fhir2/api/search/ObservationSearchQueryTest.java
Medhavi-16: Yes, I'll add it

api/src/test/java/org/openmrs/module/fhir2/api/search/ObservationSearchQueryTest.java
Medhavi-16: Should we check distinct Encounters or rather extract the `encounterDatetime` and check the count of distinct `encounterDatetime`?

varung-31: Yeah, we can do that as well.

Medhavi-16:
@varung-31 @ibacher PTAL

api/src/test/java/org/openmrs/module/fhir2/api/search/ObservationSearchQueryTest.java
```diff
@@ -1585,11 +1592,183 @@ public void searchForLastnObs_shouldHandleNoCodeRequest() {
...
+	
 	private IBundleProvider search(SearchParameterMap theParams) {
 		return searchQuery.getQueryResults(theParams, dao, translator, searchQueryInclude);
 	}
 	
 	private List<IBaseResource> get(IBundleProvider results) {
 		return results.getResources(START_INDEX, END_INDEX);
 	}
+	
+	int getDistinctEncounterDatetime(List<IBaseResource> resultList) {
```
varung-31, line 1745:
```suggestion
	private int getDistinctEncounterDatetime(List<IBaseResource> resultList) {
```

varung-31 (commented):
@Medhavi-16 Thanks for your work on this. Mostly LGTM, we'll wait for @ibacher to give it a look as well.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirEncounterDaoImpl.java
```diff
@@ -50,4 +81,53 @@ protected void setupSearchParams(Criteria criteria, SearchParameterMap theParams
 			}
 		});
 	}
+	
+	private int getMaxParameter(SearchParameterMap theParams) {
```
ibacher, line 85: It would maybe be a good idea to try to refactor this and `getTopNRankedUuids()` into utility classes so we can prevent code duplication. Maybe instead of having a `LastnEncountersResults` and `LastnObservationResult` we can get away with a `LastnResult` object which holds the UUID and relevant date then something like a `HashMap` to store other properties (e.g., concept)

ibacher (commented):
@Medhavi-16 Mostly looks good to me. I've added a request to do some refactoring so we can keep as much of this implementation in one place as possible.

Medhavi-16:
@ibacher @varung-31 PTAL

api/src/main/java/org/openmrs/module/fhir2/api/util/LastnResults.java
```diff
@@ -0,0 +1,89 @@
...
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.util;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+
+public class LastnResults {
```
varung-31, line 17: Please add a javadoc here.

api/src/main/java/org/openmrs/module/fhir2/api/util/LastnResults.java
```diff
@@ -0,0 +1,89 @@
...
+	
+	String uuid;
+	
+	Date datetime;
+	
+	Map<String, Object> attributes;
```
varung-31, line 23:
```suggestion
	private final String uuid;
	
	private final Date datetime;
	
	private final Map<String, Object> attributes;
```

api/src/main/java/org/openmrs/module/fhir2/api/util/LastnResults.java
```diff
@@ -0,0 +1,89 @@
...
+		this.datetime = (Date) datetime;
+		this.attributes = attributes;
+	}
+	
+	public LastnResults(Object[] objects) {
+		uuid = (String) objects[0];
+		datetime = (Date) objects[1];
+	}
+	
+	public Map<String, Object> getAttributes() {
```
varung-31, line 36: Instead of explicitly writing getter and setter for each field, it's easier to use the `@Data` annotation from Lombok on the class. It automatically provides us with the getter, setter, toString, hashCode, equals of all the final fields.

api/src/main/java/org/openmrs/module/fhir2/api/util/LastnResults.java
```diff
@@ -0,0 +1,89 @@
...
+	
+	public void setDatetime(Date datetime) {
+		this.datetime = datetime;
+	}
+	
+	public void setAttributes(Map<String, Object> attributes) {
+		this.attributes = attributes;
+	}
+	
+	public static List<String> getTopNRankedUuids(List<LastnResults> list, int max) {
```
varung-31, line 60: It's not a good idea to have this `LastnResult` model inside the utils class. It would be better to define the model in the dao.impl package whereas this method can reside in a separate utils class (public final) called maybe `LastnOperationUtils.java`.

Medhavi-16: Can we have a package called `model` in `dao` package for `LastnResult` ?

ibacher: I'm actually in favour of having this in the `util` folder. `model` is primarily intended for the Hibernate objects that we've needed to add over time (`GroupMember` being a bit of an oddity). I don't think we need a sub-package for dao...

api/src/main/java/org/openmrs/module/fhir2/api/util/LastnResults.java
ibacher: Even better would be a Java 14 Record class... maybe someday.

api/src/test/java/org/openmrs/module/fhir2/api/search/ObservationSearchQueryTest.java
```diff
@@ -1585,11 +1592,183 @@ public void searchForLastnObs_shouldHandleNoCodeRequest() {
...
+	}
+	
 	private IBundleProvider search(SearchParameterMap theParams) {
 		return searchQuery.getQueryResults(theParams, dao, translator, searchQueryInclude);
 	}
 	
 	private List<IBaseResource> get(IBundleProvider results) {
 		return results.getResources(START_INDEX, END_INDEX);
 	}
+
```
Medhavi-16, line 1744: @ibacher @varung-31 `getDistinctEncounterDatetime` is only checking for distinct `encounterDatetime` . Should I refactor this to a custom matcher as well?

api/src/test/java/org/openmrs/module/fhir2/api/search/ObservationSearchQueryTest.java
```diff
@@ -1585,11 +1592,183 @@ public void searchForLastnObs_shouldHandleNoCodeRequest() {
...
+	
 	private IBundleProvider search(SearchParameterMap theParams) {
 		return searchQuery.getQueryResults(theParams, dao, translator, searchQueryInclude);
 	}
 	
 	private List<IBaseResource> get(IBundleProvider results) {
 		return results.getResources(START_INDEX, END_INDEX);
 	}
+	
+	private int getDistinctEncounterDatetime(List<IBaseResource> resultList) {
```
Medhavi-16, line 1745: @ibacher @varung-31 `getDistinctEncounterDatetime` is only getting the count of distinct `encounterDatetime`, should I refactor it to a custom matcher as well?

ibacher: It's not a bad idea, but I don't think it's essential. It's important to keep in mind what a failed test will look like. The pre-matcher version of `IsSortedAndWithinMax` would've just resulted in "Expected true but got false", which isn't very helpful in figuring out *why* it failed the check. `getDistinctEncounterDatetime`, the way it's used will result in something like "Expected a result less than or equal to 2, but got 4", which isn't bad. I'd only do so if you think you can add a meaningfully better failure message.

Medhavi-16:
@ibacher @varung-31 PTAL

api/src/main/java/org/openmrs/module/fhir2/api/util/LastnOperationUtils.java
```diff
@@ -0,0 +1,48 @@
...
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.util;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+
+public final class LastnOperationUtils {
```
varung-31, line 16: A javadoc can be added here as well.

api/src/main/java/org/openmrs/module/fhir2/api/util/LastnResult.java
```diff
@@ -0,0 +1,52 @@
...
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.util;
+
+/**
+ * The class stores the result of $lastn type operations
+ */
```
varung-31, line 14: It's perhaps better if we can place it just above the class definition. Putting it above an import looks misleading.

api/src/main/java/org/openmrs/module/fhir2/api/util/LastnResult.java
```diff
@@ -0,0 +1,52 @@
...
+public class LastnResult {
+	
+	private final String uuid;
+	
+	private final Date datetime;
+	
+	private final Map<String, Object> attributes;
+	
+	/**
+	 * Constructor for observations in $lastn operation
```
varung-31, line 30: We should have a more generic description for these constructors since they might be used for other resources as well in the future.

varung-31 (changes requested):
@Medhavi-16, Just a few minor comments and we should be good to go!

Medhavi-16:
@ibacher @varung-31 PTAL

varung-31 (approved):
Thanks @Medhavi-16 for the great work. LGTM!

ibacher (approved):

PR #375, opened 2021-11-18
FM2-437: Update Location Support for core changes and mCSD profile

pmanko:
- Added support for addition of the `type` attribute to the Location model in core 2.5.x
- Implemented mappings for FHIR `Location.type` field 

## Issue I worked on
see https://issues.openmrs.org/browse/FM2-437

  No? -> execute above command


api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
```diff
@@ -159,7 +159,11 @@ private FhirConstants() {
 	public static final String LOCATION_CONTACT_POINT_ATTRIBUTE_TYPE = "fhir2.locationContactPointAttributeTypeUuid";
 	
 	public static final String PROVIDER_CONTACT_POINT_ATTRIBUTE_TYPE = "fhir2.providerContactPointAttributeTypeUuid";
-	
+
+	public static final String LOCATION_TYPE_ATTRIBUTE_TYPE = "fhir2.locationTypeAttributeTypeUuid";
+
+	public static final String LOCATION_TYPE_SYSTEM_URL = "fhir2.locationTypeSystemUrl";
```
ibacher, line 165: I absolutely do not think this is a good idea to put in a global property. We should either have the system hard-coded or have a global property that refers to the concept source and have the mapping to the system URL in the `fhir_concept_source` table.

api/src/main/java/org/openmrs/module/fhir2/api/translators/LocationTypeTranslator.java
```diff
@@ -0,0 +1,41 @@
...
+public interface LocationTypeTranslator extends ToFhirTranslator<Location, List<CodeableConcept>>, UpdatableOpenmrsTranslator<Location, List<CodeableConcept>> {
+	
+	/**
+	 * Maps an {@link Location} to a type {@link CodeableConcept}
+	 *
+	 * @param omrsLocation the location to translate
+	 * @return the FHIR CodeableConcept resource for the location's type
+	 */
+	@Override
+	List<CodeableConcept> toFhirResource(@Nonnull Location omrsLocation);
```
ibacher, line 30: I think it's better to just build this for `CodeableConcept`. We can refactor things when we need to support several of them, but for now, we really only support a single type anyways.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/LocationTypeTranslatorImpl.java
```diff
@@ -0,0 +1,96 @@
...
+	
+	@Override
+	public Location toOpenmrsType(@Nonnull Location location, @Nonnull List<CodeableConcept> types) {
+		LocationAttributeType typeAttributeType = locationService.getLocationAttributeTypeByUuid(
+		    globalPropertyService.getGlobalProperty(FhirConstants.LOCATION_TYPE_ATTRIBUTE_TYPE));
+		String locationTypeSystem = globalPropertyService.getGlobalProperty(FhirConstants.LOCATION_TYPE_SYSTEM_URL);
+
+		Concept typeConcept = null;
+
+		// Only save the type from this code system as per mCSD IG: https://terminology.hl7.org/2.1.0/CodeSystem-v3-RoleCode.html
```
ibacher, line 77: This is a business rule we should probably not enforce here. The binding type on the resource is "extensible" rather than "required", which means we should be able to support other possible codes.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/LocationTypeTranslatorImpl.java
```diff
@@ -0,0 +1,96 @@
...
+		Optional<CodeableConcept> type = types.stream().filter(t -> t.hasCoding()
+						&& t.getCodingFirstRep().getSystem().equals(locationTypeSystem))
+				.findFirst();
+
+		if (type.isPresent()) {
+			typeConcept = conceptTranslator.toOpenmrsType(type.get());
+		}
+
+		if (typeConcept != null) {
```
ibacher, line 86: We should definitely have a check in here to see if an attribute of this attribute type already exists, right? Otherwise, we'll just keep adding attributes, which doesn't seem right. (One of the reasons support for multiple codes becomes hard is we then need to add logic to determine which codes, if any, were changed, etc.)

ibacher (commented):
@pmanko Thanks for this. Some initial thoughts on the draft here.

pmanko (commented):
@ibacher let me know what else you'd like changed!

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/LocationTranslatorImplTest.java
```diff
@@ -24,16 +24,12 @@
...
 import java.util.List;
 
 import org.exparity.hamcrest.date.DateMatchers;
-import org.hl7.fhir.r4.model.Coding;
-import org.hl7.fhir.r4.model.ContactPoint;
-import org.hl7.fhir.r4.model.IdType;
-import org.hl7.fhir.r4.model.Identifier;
-import org.hl7.fhir.r4.model.Provenance;
-import org.hl7.fhir.r4.model.Reference;
+import org.hl7.fhir.r4.model.*;
```
theanandankit, line 32: @pmanko please change this * to specific imports.

PR #376, opened 2021-11-30
FM2-444: Add the ability to save Location Tags

mozzy11:
## Issue I worked on

see https://issues.openmrs.org/browse/FM2-444

  No? -> execute above command

### DIFF
--- a/test-data/src/test/resources/org/openmrs/module/fhir2/providers/LocationWebTest_create.json
+++ b/test-data/src/test/resources/org/openmrs/module/fhir2/providers/LocationWebTest_create.json
@@ -2,14 +2,8 @@
   "resourceType": "Location",
   "meta": {
     "versionId": "1",
-    "lastUpdated": "2021-11-30T08:43:54.105+00:00",
-    "source": "#DNbZg5doCf3wLAkB",
-    "profile": [
-      "http://ihe.net/fhir/StructureDefinition/IHE.mCSD.FacilityLocation"
-    ],
     "tag": [
       {
-        "system": "http://facility_registry/mcsd_location",
         "code": "mCSD_Location",
         "display": "mCSD_Location"
       }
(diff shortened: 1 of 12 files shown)


mozzy11:
cc @pmanko

ibacher:
@mozzy11 I think we need to reconsider this. In particular, I think the existing code translating location tags is... not right. It seems to me that the way this will work will be to create a tag per location created by FHIR. So, e.g., instead of ending up with one "mCSD Location" tag, we will end up with one per-location.

I also think we should break out the responsibilities into different parts

1. UUID on new items should be assigned at the service create method (actually ideally the one in the BaseFhirService... I thought we did that there already but apparently not).
2. We probably should have a translator for tags that will look up an existing tag (if any) and assign that to the location. I would not create a new tag while creating a new location... we should ignore non-existent tags.

test-data/src/test/resources/org/openmrs/module/fhir2/providers/LocationWebTest_create.json
```diff
@@ -1,5 +1,20 @@
...
+  "meta": {
+    "versionId": "1",
+    "lastUpdated": "2021-11-30T08:43:54.105+00:00",
+    "source": "#DNbZg5doCf3wLAkB",
+    "profile": [
+      "http://ihe.net/fhir/StructureDefinition/IHE.mCSD.FacilityLocation"
+    ],
+    "tag": [
+      {
+        "system": "http://facility_registry/mcsd_location",
```
ibacher, line 12: It's best we don't include a system here... there's no meaningful system for this. Really, in this case, it should just be "code", I think...

ibacher (approved):
LGTM. One final thing though, could you add a similar integration test for the R3 version of the same provider?

ibacher (approved):

PR #388, opened 2022-03-07
FM2-450: Create a basic mapping between OMRS ConceptSets and FHIR ValueSets

Richa-09:
FM2-450: Create a basic mapping between OMRS ConceptSets and FHIR ValueSets

Mapped OMRS ConceptSets to FHIR ValueSets

## Issue I worked on
see https://issues.openmrs.org/browse/FM2-450

  No? -> execute above command


api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ValueSetTranslatorImpl.java
```diff
@@ -0,0 +1,91 @@
...
+		valueSet.setDescription(description);
+		
+		if (!concept.getSet()) {
+			return valueSet;
+		}
+		
+		ValueSet.ValueSetComposeComponent compose = new ValueSet.ValueSetComposeComponent();
+		
+		for (ConceptSet conceptSet : concept.getConceptSets()) {
+			Concept conceptSetMember = conceptSet.getConcept();
```
ibacher, line 57: There is one tricky spot here: what happens if the concept set member is itself a concept set? Here it seems the thing to do is to include a reference in the `compose.include.valueSet` instead of adding it as a normal concept. Sorry that should've been in the original ticket.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ValueSetTranslatorImpl.java
```diff
@@ -0,0 +1,91 @@
...
+			
+			for (ConceptMap mapping : conceptSetMember.getConceptMappings()) {
+				ConceptReferenceTerm crt = mapping.getConceptReferenceTerm();
```
ibacher, line 60:
```suggestion
				ConceptReferenceTerm crt = mapping.getConceptReferenceTerm();
				if (crt == null || crt.getConceptSource() == null) {
					continue;
				}
				
```

api/src/main/java/org/openmrs/module/fhir2/providers/r3/ValueSetFhirResourceProvider.java
```diff
@@ -0,0 +1,45 @@
...
+public class ValueSetFhirResourceProvider implements IResourceProvider {
+	
+	@Autowired
+	FhirValueSetService fhirValueSetService;
+	
+	@Override
+	public Class<? extends IBaseResource> getResourceType() {
+		return ValueSet.class;
+	}
+
```
ibacher, line 40: We should also at least have a `@Read` method that gets the value set by UUID.

test-data/src/test/resources/org/openmrs/module/fhir2/api/dao/impl/FhirConceptDaoImplTest_initial_data.xml
```diff
@@ -13,4 +13,7 @@
     <concept_reference_source concept_source_id="6" name="LOINC" creator="1" date_created="2005-01-01 00:00:00.0" retired="0" uuid="2b3c1ff8-768a-102f-83f4-12313b04a615" />
     <concept_reference_term concept_reference_term_id="1" concept_source_id="6" code="1000-1" creator="1" date_created="2005-01-01 00:00:00.0" uuid="d471950d-0ef5-424c-b72a-3fff29af6b0b" />
     <concept_reference_map concept_map_id="1" concept_reference_term_id="1" concept_map_type_id="1" creator="1" date_created="2005-01-01 00:00:00.0" concept_id="3" uuid="2f7e0cd7-5866-4345-a24a-9c75cd3a876f" />
+
+    <concept_name concept_id="23" name="DEMO" locale="en_GB" creator="1" date_created="2008-08-18 12:38:58.0" concept_name_id="2315" concept_name_type="FULLY_SPECIFIED" locale_preferred="1" voided="false" uuid="b3ebe4f6-50b4-4e78-82b1-5G792a7ede92"/>
+
```
ibacher, line 18:
```suggestion
    <concept_name concept_id="23" name="DEMO" locale="en_GB" creator="1" date_created="2008-08-18 12:38:58.0" concept_name_id="2315" concept_name_type="FULLY_SPECIFIED" locale_preferred="1" voided="false" uuid="b3ebe4f6-50b4-4e78-82b1-5G792a7ede92"/>
```

ibacher (commented):
@Richa-09 Thanks for this! It mostly looks great and well done. I've got a few small points which are largely things I forgot to think about while writing up the ticket.

Richa-09:
@ibacher PTAL! will be adding more tests.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ValueSetTranslatorImpl.java
```diff
@@ -0,0 +1,108 @@
...
+		valueSet.setDescription(description);
+		
+		ValueSet.ValueSetComposeComponent compose = new ValueSet.ValueSetComposeComponent();
+		
+		for (ConceptSet conceptSet : concept.getConceptSets()) {
+			Concept conceptSetMember = conceptSet.getConcept();
+			
+			if (conceptSetMember != null && conceptSetMember.getSet()) {
+				ValueSet.ConceptSetComponent conceptSetComponent = new ValueSet.ConceptSetComponent();
+				conceptSetComponent.addValueSet(valueSetReferenceTranslator.toFhirResource(conceptSetMember).getReference());
```
Richa-09, line 71: @ibacher is it correct to add reference like this or it should be an absolute URL?

ibacher: @Richa-09 Normally, it would be fine to use a relative reference. However, in this case because the FHIR spec says that the value is "canonical" rather than a "reference", I do think we need to use the full URL.

Richa-09: @ibacher Since base URL will vary for different implementers, is there any way to get the base URL? or should I hard code?

ibacher: @Richa-09 Ooph... good point. Let's leave it as a standard reference for now. There is a way to this, but it will be messy.

The "right way" would be to move the [OpenmrsFhirAddressStrategy](https://github.com/openmrs/openmrs-module-fhir2/blob/master/omod/src/main/java/org/openmrs/module/fhir2/web/util/OpenmrsFhirAddressStrategy.java) class into the API module, capture the raw `HttpServletRequest` as a parameter in the `ResourceProvider` and then call [`determineServerBase`](https://github.com/openmrs/openmrs-module-fhir2/blob/290688fc9189b0fa108c77f09f46c2560292e84d/omod/src/main/java/org/openmrs/module/fhir2/web/util/OpenmrsFhirAddressStrategy.java#L83) with the request. Let me think if there's a better way to implement things.

ibacher:
@Richa-09 Thanks for this! One last bit of work here: would you be able to add corresponding integration-tests for the ValueSet resource?

Richa-09:
@ibacher Yeah, sure. I'm working on it.

Richa-09:
@ibacher PTAL! I have added integration-tests for the ValueSet resource.

ibacher (approved):

PR #392, opened 2022-04-12
FM2-469: Remove the FhirDurationUnitMap

mseaton:
See: https://issues.openmrs.org/projects/FM2/issues/FM2-469


ibacher (approved):

ibacher:
Thanks @mseaton! This is great!

mseaton:
Looks like I needed to run integration tests explicitly.  Doing that now and will fix.  @ibacher

ibacher:
Ah, yes, I should probably add a note to the README. I find it helpful to be able to just fix the unit tests and not worry about the integration tests, so they're default disabled. Alternatively, maybe I should invert that and I can just run with `-p !integration-test`?

mseaton:
Updated the integration test data, building now...

PR #394, opened 2022-04-15
FM2-470: Allow MedicationRequestOrderFrequency to support more frequencies

mseaton:
…ger frequency

See: https://issues.openmrs.org/browse/FM2-470
See: https://talk.openmrs.org/t/mapping-openmrs-order-frequencies-fhir/36421


mogoodrich:
Sounds like there's consensus around switching to timing.code?

ibacher (approved):

PR #395, opened 2022-04-21
FM2-463: Map DrugOrder Dose and Dose Units to MedicationRequest

mseaton:
See:  https://issues.openmrs.org/projects/FM2/issues/FM2-463

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
@@ -314,7 +314,7 @@ public class FhirConstants {
 	
 	public static final String CONDITION_OBSERVATION_CONCEPT_UUID = "1284AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
 	
-	public static final String SNOMED_CT_CONCEPT_SOURCE_HL7_CODE = "SCT";
-	
 	public static final String SNOMED_SYSTEM_URI = "http://snomed.info/sct";
+	
+	public static final String RX_NORM_SYSTEM_URI = "http://www.nlm.nih.gov/research/umls/rxnorm";
 }
--- a/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
@@ -313,4 +313,8 @@ public class FhirConstants {
 	public static final String EVERYTHING_SEARCH_HANDLER = "everything.search.handler";
 	
 	public static final String CONDITION_OBSERVATION_CONCEPT_UUID = "1284AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
+	
+	public static final String SNOMED_CT_CONCEPT_SOURCE_HL7_CODE = "SCT";
+	
+	public static final String SNOMED_SYSTEM_URI = "http://snomed.info/sct";
 }
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConceptTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConceptTranslatorImpl.java
@@ -18,8 +18,9 @@ import org.hl7.fhir.r4.model.CodeableConcept;
 import org.hl7.fhir.r4.model.Coding;
 import org.openmrs.Concept;
 import org.openmrs.ConceptMap;
-import org.openmrs.ConceptName;
 import org.openmrs.ConceptReferenceTerm;
+import org.openmrs.Duration;
+import org.openmrs.module.fhir2.FhirConstants;
 import org.openmrs.module.fhir2.api.FhirConceptService;
 import org.openmrs.module.fhir2.api.FhirConceptSourceService;
 import org.openmrs.module.fhir2.api.translators.ConceptTranslator;
@@ -61,6 +62,11 @@ public class ConceptTranslatorImpl implements ConceptTranslator {
 			
 			String sourceUrl = conceptSourceToURL(crt.getConceptSource().getName());
 			if (sourceUrl == null) {
+				if (Duration.SNOMED_CT_CONCEPT_SOURCE_HL7_CODE.equals(crt.getConceptSource().getHl7Code())) {
+					sourceUrl = FhirConstants.SNOMED_SYSTEM_URI;
+				}
+			}
+			if (sourceUrl == null) {
 				continue;
 			}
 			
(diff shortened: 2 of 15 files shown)


api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
```diff
@@ -313,4 +313,8 @@ private FhirConstants() {
 	public static final String EVERYTHING_SEARCH_HANDLER = "everything.search.handler";
 	
 	public static final String CONDITION_OBSERVATION_CONCEPT_UUID = "1284AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
+	
+	public static final String SNOMED_SYSTEM_URI = "http://snomed.info/sct";
+	
+	public static final String RX_NORM_SYSTEM_URI = "http://www.nlm.nih.gov/research/umls/rxnorm";
```
mseaton, line 319: @ibacher can you confirm the above are correct and good to code in as constants?

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConceptTranslatorImpl.java
```diff
@@ -52,14 +54,21 @@ public CodeableConcept toFhirResource(@Nonnull Concept concept) {
 				continue;
 			}
 			
-			String mappingType = mapping.getConceptMapType().getName();
-			if (mappingType == null || !mappingType.equalsIgnoreCase("SAME-AS")) {
+			ConceptMapType mapType = mapping.getConceptMapType();
+			boolean sameAs = mapType.getUuid() != null && mapType.getUuid().equals(ConceptMapType.SAME_AS_MAP_TYPE_UUID);
+			sameAs = sameAs || (mapType.getName() != null && mapType.getName().equalsIgnoreCase("SAME-AS"));
+			if (!sameAs) {
```
mseaton, line 60: There is an OpenMRS core constant for the uuid of the "same-as" concept map type.  So we should compare against this.  We can also look to see if the name is "SAME-AS" as was happening before - seems safe.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConceptTranslatorImpl.java
```diff
@@ -52,14 +54,21 @@ public CodeableConcept toFhirResource(@Nonnull Concept concept) {
...
 			}
 			
 			ConceptReferenceTerm crt = mapping.getConceptReferenceTerm();
 			
 			String sourceUrl = conceptSourceToURL(crt.getConceptSource().getName());
+			if (sourceUrl == null) {
+				if (Duration.SNOMED_CT_CONCEPT_SOURCE_HL7_CODE.equals(crt.getConceptSource().getHl7Code())) {
+					sourceUrl = FhirConstants.SNOMED_SYSTEM_URI;
+				}
+			}
```
mseaton, line 71: Since there is an OpenMRS constant for the snomed-ct hl7 code, we should attempt to add a snomed-ct mapping if not otherwise found by url in the mapping table.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConceptTranslatorImpl.java
```diff
@@ -85,6 +94,11 @@ public Concept toOpenmrsType(@Nonnull CodeableConcept concept) {
 			}
 			
 			String codingSource = conceptURLToSource(coding.getSystem());
+			if (codingSource == null) {
+				if (FhirConstants.SNOMED_SYSTEM_URI.equals(coding.getSystem())) {
+					codingSource = Duration.SNOMED_CT_CONCEPT_SOURCE_HL7_CODE;
+				}
+			}
```
mseaton, line 101: Likewise, we should be able to translate a concept back from snomed-ct, even if there is no uri mapping in the fhir table.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConceptTranslatorImpl.java
```diff
@@ -105,10 +119,7 @@ public Concept toOpenmrsType(@Nonnull CodeableConcept concept) {
 	private void addConceptCoding(Coding coding, String system, String code, Concept concept) {
 		coding.setSystem(system);
 		coding.setCode(code);
-		
-		ConceptName conceptName = concept.getName();
-		String display = (conceptName == null || conceptName.getName() == null) ? "" : conceptName.getName();
-		coding.setDisplay(display);
+		coding.setDisplay(concept.getDisplayString());
```
mseaton, line 122: The above is handled by concept.getDisplayString() in a standard way.  It also allows for easier mocking.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/DosageTranslatorImpl.java
```diff
@@ -43,6 +50,32 @@ public Dosage toFhirResource(@Nonnull DrugOrder drugOrder) {
 		dosage.setRoute(conceptTranslator.toFhirResource(drugOrder.getRoute()));
 		dosage.setTiming(timingTranslator.toFhirResource(drugOrder));
 		
+		if (drugOrder.getDose() != null || drugOrder.getDoseUnits() != null) {
+			Dosage.DosageDoseAndRateComponent doseAndRate = new Dosage.DosageDoseAndRateComponent();
+			Quantity dose = new SimpleQuantity();
+			dose.setValue(drugOrder.getDose());
+			if (drugOrder.getDoseUnits() != null) {
+				CodeableConcept doseUnits = conceptTranslator.toFhirResource(drugOrder.getDoseUnits());
+				if (doseUnits != null) {
```
mseaton, line 59: Per ticket comments, this will use the RxNorm mapping if present, then the snomed ct if present, and then the uuid with null system otherwise.

mogoodrich (approved):
Sweet, LGTM, thanks @mseaton !

api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
ibacher: LGTM

ibacher (commented):
Thanks @mseaton! This mostly looks good. Two suggestions and one question.

mseaton:
@ibacher I made the 2 changes you requested above.  I left the url -> source method returning either the source name or hl7 code as the service / dao method that is called accepts either source name or hl7code in the query.

mseaton:
@ibacher - the checks on this PR had been failing due to integration test failures, due to the need to limit the MedicationRequest template a bit between v3 and v4.  I think I fixed that.  I'm going to merge this in.

PR #396, opened 2022-04-25
FM2-473:  Map DrugOrder Quantity, Quantity Units and Num Refills to M…

mseaton:
…edicationRequest

See:  https://issues.openmrs.org/browse/FM2-473

NOTE:  This PR is issued against the FM2-463 branch, as it builds upon work done in that branch.  @ibacher / we should consider this when figuring out how to merge these in.

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/MedicationQuantityCodingTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/MedicationQuantityCodingTranslatorImpl.java
@@ -1,0 +1,83 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.fhir2.api.translators.impl;
+
+import javax.annotation.Nonnull;
+
+import lombok.AccessLevel;
+import lombok.Setter;
+import org.hl7.fhir.r4.model.CodeableConcept;
+import org.hl7.fhir.r4.model.Coding;
+import org.hl7.fhir.r4.model.ICoding;
+import org.openmrs.Concept;
+import org.openmrs.module.fhir2.FhirConstants;
+import org.openmrs.module.fhir2.api.translators.CodingTranslator;
+import org.openmrs.module.fhir2.api.translators.ConceptTranslator;
+import org.openmrs.util.OpenmrsUtil;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+@Setter(AccessLevel.PACKAGE)
+/**
+ * This is an implementation of Coding Translator that maps a Medication Quantity Concept to/from a
+ * Coding in FHIR with a preferred set of systems and codes to prioritize. This will first favor
+ * using RxNorm as the coding system. If this is not present on the OpenMRS concept, it will next
+ * favor SNOMED-CT. Finally, if neither are present, it will favor the Concept UUID with a null
+ * system.
+ */
+public class MedicationQuantityCodingTranslatorImpl implements CodingTranslator {
+	
+	@Autowired
+	private ConceptTranslator conceptTranslator;
+	
+	@Override
+	public Coding toFhirResource(@Nonnull Concept concept) {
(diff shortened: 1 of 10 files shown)


api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/MedicationRequestDispenseRequestComponentTranslatorImpl.java
```diff
@@ -0,0 +1,71 @@
...
+	
+	@Override
+	public MedicationRequest.MedicationRequestDispenseRequestComponent toFhirResource(@Nonnull DrugOrder drugOrder) {
+		MedicationRequest.MedicationRequestDispenseRequestComponent dispenseRequestComponent = new MedicationRequest.MedicationRequestDispenseRequestComponent();
+		if (drugOrder.getQuantity() != null) {
+			Quantity quantity = new Quantity();
+			quantity.setValue(drugOrder.getQuantity());
+			if (drugOrder.getQuantityUnits() != null) {
+				CodeableConcept units = conceptTranslator.toFhirResource(drugOrder.getQuantityUnits());
+				Coding coding = units.getCodingFirstRep();
```
mogoodrich, line 42: I assume you've researched/thought about this, but how is the "first rep" determined?
Do we want to consider using the the same strategy we use here for DosageUnits (using RX_Norm if found, then SNOMED, otherwise just uuid)?

mseaton: I've added a new commit to address this @mogoodrich .  Please have a look.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/MedicationQuantityCodingTranslatorImpl.java
```diff
@@ -0,0 +1,83 @@
...
+@Component
+@Setter(AccessLevel.PACKAGE)
+/**
+ * This is an implementation of Coding Translator that maps a Medication Quantity Concept to/from a
+ * Coding in FHIR with a preferred set of systems and codes to prioritize. This will first favor
+ * using RxNorm as the coding system. If this is not present on the OpenMRS concept, it will next
+ * favor SNOMED-CT. Finally, if neither are present, it will favor the Concept UUID with a null
+ * system.
+ */
+public class MedicationQuantityCodingTranslatorImpl implements CodingTranslator {
```
mogoodrich, line 36: I'm wondering if we want to call this "SimpleQuantityCodingTranslator" or even just "BaesCodingTranslator" or "CodingTranslator" until we determine that what we want do here is specific to medication and need to do something different for the coding of another simple quantity?

Or are there already cases/examples in the code where we already do something differnet?

In particular, it was a little confusing to me because my assumption based on the name would be that this translator was only used for translating the quantity of medication ordered/dispensed, and not also the dose (whicih, of course is "a quantity associated with a medication", but wasn't my first reading of the meaning of the class name).

mseaton: @mogoodrich I added an alternative idea on the ticket, but on reflection I think I'm in favor of keeping this as is, pending @ibacher 's thoughts.

mogoodrich: Makes sense, I adding my thoughts on the ticket.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/MedicationRequestDispenseRequestComponentTranslatorImpl.java
```diff
@@ -0,0 +1,64 @@
...
+			dispenseRequestComponent.setNumberOfRepeatsAllowed(drugOrder.getNumRefills());
+		}
+		return dispenseRequestComponent;
+	}
+	
+	@Override
+	public DrugOrder toOpenmrsType(@Nonnull DrugOrder drugOrder,
+	        @Nonnull MedicationRequest.MedicationRequestDispenseRequestComponent resource) {
+		Quantity quantity = resource.getQuantity();
+		if (quantity != null && quantity.getValue() != null) {
```
ibacher, line 56: This is somewhat defensive, but the HL7 core libraries have a bad habit of doing essentially this:

```java
class MedicationRequest.MedicationRequestDispenseRequestComponent {
	public Quantity getQuantity() {
		if (this.quantity == null) {
			this.quantity = new Quantity();
		}
		return this.quantity;
	}
}
```


```suggestion
		Quantity quantity;
		if (resource.hasQuantity() && (quantity = resource.getQuantity()).hasValue()) {
```

ibacher (commented):
Thanks @mseaton! One small adjustment (hope I got the code sort of right).

mseaton:
@ibacher - I've pushed an update based on your feedback.  Let me know if this is what you had in mind.

PR #397, opened 2022-05-03
Encounter has medication request spike

mseaton:
Draft PR for supporting "nested reverse chaining" - i.e. the ":has..." parameters, with an objective of getting just those Encounter resources that have associated MedicationRequests.


mseaton:
Some additional questions beyond this:

What's the best way to get back only "true" Encounters - i.e. exclude Visits represented as Encounters?  Based on what I'm [reading here](http://hl7.org/fhir/R4/search.html#tag), it sounds like this might be something we'd support like this:

http://localhost:8080/openmrs/ws/fhir2/R4/Encounter_tag=http://fhir.openmrs.org/ext/encounter-tag|encounter

Is that right @ibacher ?  This doesn't appear to currently work, so we may need to ticket something like this independently...

PR #398, opened 2022-05-05
FM2-477 and FM2-478: Support limiting to only Encounters that have medication requests

mseaton:
Both FM2-477 and FM2-478 are written with a common goal - to be able to return encounters (only encounters, not visits) that contain medication requests.  Those 2 tickets break this down into supporting the _has:MedicationRequest:encounter parameter on the Encounter resource (FM2-477), and in supporting the _tag=http://fhir.openmrs.org/ext/encounter-tag|encounter on the Encounter resource.

See:  https://issues.openmrs.org/browse/FM2-477
See:  https://issues.openmrs.org/browse/FM2-478

Both of these tickets hit the same blocks of code, and both benefited from some refactoring I did as a part of this on the FhirEncounterService and associated unit tests, so I bundled these together so as not to have to deal with a mess of merge issues.

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/FhirEncounterService.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/FhirEncounterService.java
@@ -15,6 +15,7 @@ import ca.uhn.fhir.model.api.Include;
 import ca.uhn.fhir.rest.api.SortSpec;
 import ca.uhn.fhir.rest.api.server.IBundleProvider;
 import ca.uhn.fhir.rest.param.DateRangeParam;
+import ca.uhn.fhir.rest.param.HasAndListParam;
 import ca.uhn.fhir.rest.param.ReferenceAndListParam;
 import ca.uhn.fhir.rest.param.TokenAndListParam;
 import ca.uhn.fhir.rest.param.TokenParam;
@@ -25,7 +26,7 @@ public interface FhirEncounterService extends FhirService<Encounter> {
 	IBundleProvider searchForEncounters(DateRangeParam date, ReferenceAndListParam location,
 	        ReferenceAndListParam participant, ReferenceAndListParam subject, TokenAndListParam encounterType,
 	        TokenAndListParam id, DateRangeParam lastUpdated, SortSpec sort, HashSet<Include> includes,
-	        HashSet<Include> revIncludes);
+	        HashSet<Include> revIncludes, HasAndListParam hasMedicationRequestParam);
 	
 	IBundleProvider getEncounterEverything(TokenParam identifier);
 }
--- a/api/src/main/java/org/openmrs/module/fhir2/api/FhirEncounterService.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/FhirEncounterService.java
@@ -26,7 +26,7 @@ public interface FhirEncounterService extends FhirService<Encounter> {
 	IBundleProvider searchForEncounters(DateRangeParam date, ReferenceAndListParam location,
 	        ReferenceAndListParam participant, ReferenceAndListParam subject, TokenAndListParam encounterType,
 	        TokenAndListParam id, DateRangeParam lastUpdated, SortSpec sort, HashSet<Include> includes,
-	        HashSet<Include> revIncludes, HasAndListParam hasMedicationRequestParam);
+	        HashSet<Include> revIncludes, HasAndListParam hasAndListParam);
 	
 	IBundleProvider getEncounterEverything(TokenParam identifier);
 }
--- a/api/src/main/java/org/openmrs/module/fhir2/api/FhirEncounterService.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/FhirEncounterService.java
@@ -9,24 +9,14 @@
  */
 package org.openmrs.module.fhir2.api;
 
-import java.util.HashSet;
-
-import ca.uhn.fhir.model.api.Include;
-import ca.uhn.fhir.rest.api.SortSpec;
 import ca.uhn.fhir.rest.api.server.IBundleProvider;
-import ca.uhn.fhir.rest.param.DateRangeParam;
-import ca.uhn.fhir.rest.param.HasAndListParam;
-import ca.uhn.fhir.rest.param.ReferenceAndListParam;
-import ca.uhn.fhir.rest.param.TokenAndListParam;
 import ca.uhn.fhir.rest.param.TokenParam;
 import org.hl7.fhir.r4.model.Encounter;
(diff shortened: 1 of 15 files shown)


api/src/main/java/org/openmrs/module/fhir2/api/FhirEncounterService.java
```diff
@@ -9,23 +9,14 @@
...
 import org.hl7.fhir.r4.model.Encounter;
+import org.openmrs.module.fhir2.api.search.param.EncounterSearchParams;
 
 public interface FhirEncounterService extends FhirService<Encounter> {
 	
-	IBundleProvider searchForEncounters(DateRangeParam date, ReferenceAndListParam location,
-	        ReferenceAndListParam participant, ReferenceAndListParam subject, TokenAndListParam encounterType,
-	        TokenAndListParam id, DateRangeParam lastUpdated, SortSpec sort, HashSet<Include> includes,
-	        HashSet<Include> revIncludes);
+	IBundleProvider searchForEncounters(EncounterSearchParams encounterSearchParams);
```
mseaton, line 20: It was getting a bit unwieldy to keep adding parameters to this list, and then having to change all of the mocks in the unit tests that called this method, so I took the opportunity to change to a different pattern with a single parameter bean.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseEncounterDao.java
```diff
@@ -46,10 +57,70 @@ protected void setupSearchParams(Criteria criteria, SearchParameterMap theParams
 				case FhirConstants.COMMON_SEARCH_HANDLER:
 					handleCommonSearchParameters(entry.getValue()).ifPresent(criteria::add);
 					break;
+				case FhirConstants.HAS_SEARCH_HANDLER:
+					entry.getValue().forEach(param -> handleHasAndListParam(criteria, (HasAndListParam) param.getParam()));
+					break;
```
mseaton, line 62: I expect the way FHIR handles the _has parameters is that there will always be only 0 or 1 of these (multiple params are combined into the "andList" portion of this).

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseEncounterDao.java
```diff
@@ -46,10 +57,70 @@ protected void setupSearchParams(Criteria criteria, SearchParameterMap theParams
...
+					hasOrListParam.getValuesAsQueryTokens().forEach(orParam -> values.add(orParam.getParameterValue()));
+					
+					log.debug("Handling hasParam = " + hasParam.getQueryParameterQualifier());
+					log.debug("With value in " + values);
+					
+					boolean handled = false;
+					
+					// Support constraining encounter resources to those that contain only certain Medication Requests
+					if (FhirConstants.MEDICATION_REQUEST.equals(hasParam.getTargetResourceType())) {
+						if (MedicationRequest.SP_ENCOUNTER.equals(hasParam.getReferenceFieldName())) {
```
mseaton, line 90: This will effectively add the inner join restriction, and only return encounters that have associated drug orders at this point, even if there are no valid/matching parameter values found.  If this isn't ok, we can adapt, but this was simplest.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseEncounterDao.java
```diff
@@ -46,10 +57,70 @@ protected void setupSearchParams(Criteria criteria, SearchParameterMap theParams
...
+							// Constrain only on non-voided Drug Orders
+							criteria.add(Restrictions.eq("orders.class", DrugOrder.class));
+							criteria.add(Restrictions.eq("orders.voided", false));
+							
+							String paramName = hasParam.getParameterName();
+							String paramValue = hasParam.getParameterValue();
+							if (MedicationRequest.SP_INTENT.equals(paramName)) {
+								if (values.contains(MedicationRequest.MedicationRequestIntent.ORDER.toCode())) {
+									// No additional constraints needed, all Orders are given/assumed intent=order
+									handled = true;
```
mseaton, line 110: There is nothing really to handle here, I just wanted to add explicit support for this parameter name/value combination.  I'm doing that by setting a flag, and then logging a warning if any other parameter combination was found.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
```diff
@@ -143,22 +143,28 @@ public void delete(@Nonnull String uuid) {
...
+		
+		IBundleProvider visitBundle = new SimpleBundleProvider();
+		IBundleProvider encounterBundle = new SimpleBundleProvider();
+		
+		if (shouldIncludeResourceBasedOnEncounterTag(esp.getTag(), "visit")) {
+			visitBundle = visitService.searchForVisits(theParams);
+		}
+		if (shouldIncludeResourceBasedOnEncounterTag(esp.getTag(), "encounter")) {
+			encounterBundle = searchQuery.getQueryResults(theParams, dao, translator, searchQueryInclude);
+		}
```
mseaton, line 167: The above lines is the code responsible for handling the _tag parameter on Encounter.   Basically, if no tags restrict the encounter, or if tags that restrict match appropriately, include matching resources.  Otherwise, use an empty bundle provider.

api/src/main/java/org/openmrs/module/fhir2/api/impl/FhirEncounterServiceImpl.java
```diff
@@ -169,6 +175,24 @@ public IBundleProvider searchForEncounters(DateRangeParam date, ReferenceAndList
...
+		if (tokenAndListParam != null && tokenAndListParam.getValuesAsQueryTokens() != null) {
+			tokenAndListParam.getValuesAsQueryTokens().forEach(tokenOrListParam -> {
+				tokenOrListParam.getValuesAsQueryTokens().forEach(tokenParam -> {
+					if (tokenParam.getSystem().equals(FhirConstants.OPENMRS_FHIR_EXT_ENCOUNTER_TAG)) {
+						tagsToInclude.add(tokenParam.getValue());
+					}
+				});
+			});
+		}
+		return tagsToInclude.isEmpty() || tagsToInclude.contains(valueToCheck);
```
mseaton, line 193: Per above, if any tags are found for the OPENMRS_FHIR_EXT_ENCOUNTER_TAG system, then restrict based on that, otherwise do not.

ibacher (approved):
LGTM. Thanks @mseaton!

PR #400, opened 2022-05-10
FM2-462: Add support for MedicationDispense domain object

mseaton:
This is quite a substantial commit with a lot of additions and a few changes to existing functions, notably the concept translator.  This adds support for the new MedicationDispense domain object introduced in OpenMRS 2.6.0 via implementing the R3 and R4 MedicationDispense FHIR resource.  @ibacher I'm going to need your review on this, if possible.

See:  https://issues.openmrs.org/browse/FM2-462


mseaton:
@ibacher - just a ping on this.  I'm worried about this PR getting too stale, given the size.

ibacher:
@mseaton I haven't forgotten about this. I was just off last week 😄.

ibacher:
@mseaton Plan here: we've had a number of commits, so I'd like to release 1.4.0 and then incorporate these changes, given their dependency on 2.6.0-SNAPSHOT. Does that seem ok?

PR #404, opened 2022-06-02
FM2-481: add ObservationSearchParams class and refactor Observation service search method

pmanko2:
## Issue I worked on

  No? -> execute above command


api/src/main/java/org/openmrs/module/fhir2/api/search/param/BaseResourceSearchParams.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.HashSet;
+
+import ca.uhn.fhir.model.api.Include;
+import ca.uhn.fhir.rest.api.SortSpec;
+import ca.uhn.fhir.rest.param.DateRangeParam;
+import ca.uhn.fhir.rest.param.TokenAndListParam;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import lombok.experimental.SuperBuilder;
```
pmanko2, line 22: not thrilled about SuperBuilder still being part of the experimental package. may look into refactoring to no longer use SuperBuilder if possible. can probably use the alternative method described here https://www.baeldung.com/lombok-builder-inheritance to still make use of inheritance for resource search params

api/src/main/java/org/openmrs/module/fhir2/api/search/param/BaseResourceSearchParams.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Class containing shared search (and search result) parameters that apply to all resources See
+ * https://www.hl7.org/fhir/search.html#Summary for a comprehensive description of these parameters
+ */
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@SuperBuilder
+public class BaseResourceSearchParams implements Serializable {
```
pmanko2, line 32: did a first pass at pulling out params that are used in most resources, although it does seem like resources explicitly define which params they accept within the resource Provider class directly. wondering if this means that not all resources will support the common parameters defined in https://www.hl7.org/fhir/search.html#Summary ?

api/src/main/java/org/openmrs/module/fhir2/api/search/param/ObservationSearchParams.java
```diff
@@ -0,0 +1,69 @@
...
+	
+	private TokenAndListParam code;
+	
+	private TokenAndListParam category;
+	
+	/**
+	 * Custom AllArgsConstructor (instead of using @AllArgsConstructor annotation) that supports super
+	 * class instantiation
+	 */
+	public ObservationSearchParams(ReferenceAndListParam encounterReference, ReferenceAndListParam patientReference,
```
pmanko2, line 50: lombok doesn't support super class field instantiation via the AllArgsConstructor annotation (https://github.com/projectlombok/lombok/issues/1229), so if we want to go this route we'll need to define our own, for each class that inherits from BaseResourceSearchParams

api/src/main/java/org/openmrs/module/fhir2/api/search/param/BaseResourceSearchParams.java
pmanko2: also technically we have common search params and common search result params, not sure if it's better to distinguish between those two and have BaseResourceSearchParams/BaseResourceSearchResultParams

api/src/main/java/org/openmrs/module/fhir2/api/search/param/SearchParameterMap.java
```diff
@@ -56,10 +56,6 @@ public SearchParameterMap addParameter(@Nonnull String key, @Nonnull Serializabl
 	 * @return {@link org.openmrs.module.fhir2.api.search.param.SearchParameterMap}
 	 */
 	public SearchParameterMap addParameter(@Nonnull String key, String propertyName, @Nonnull Serializable param) {
-		if (key == null || param == null) {
```
pmanko2, line 59: ide may have accidentally changed this, but it does seem like it's an unnecessary if statement with the Nonnull annotations (unless those aren't guaranteed to throw an exception in some cases?)

omod/src/test/java/org/openmrs/module/fhir2/providers/r3/ObservationFhirResourceProviderWebTest.java
```diff
@@ -38,6 +38,7 @@
 import java.util.Objects;
 
 import ca.uhn.fhir.model.api.Include;
+import ca.uhn.fhir.rest.api.SortSpec;
```
pmanko2, line 41: these test changes were not fun

pmanko2: also seems like a lot of the r3/r4 tests are duplicated, wondering if it might be worth pulling out the duplicates into a base `ObservationFhirResourceProviderWebTest` test class and only include r3 vs r4 specific testing in these classes?

api/src/main/java/org/openmrs/module/fhir2/api/search/param/BaseResourceSearchParams.java
pmanko2: ^ ended up going this route

api/src/main/java/org/openmrs/module/fhir2/api/search/SearchQueryInclude.java
```diff
@@ -216,13 +217,14 @@ private IBundleProvider handleLocationReverseInclude(ReferenceAndListParam param
...
 	}
 	
-	private IBundleProvider handleObservationReverseInclude(ReferenceAndListParam params, String targetType) {
+	private IBundleProvider handleObservationReverseInclude(ReferenceAndListParam param, String targetType) {
 		switch (targetType) {
 			case FhirConstants.OBSERVATION:
-				return observationService.searchForObservations(null, null, params, null, null, null, null, null, null, null,
-				    null, null, null, null, null);
+				ObservationSearchParams observationSearchParams = new ObservationSearchParams();
+				observationSearchParams.setHasMember(param);
```
pmanko2, line 224: verify this is the right search param to set

pmanko (approved):
Looks good, thanks @pmanko2! 

The logic for processing and mapping search parameters still is quite complex, since we seem to be doing a subsequent mappings between different representations of the parameters for each Resource. 

This is my first time taking a close look at the search functionality, so perhaps it would be useful to bring this up and discuss in the next FHIR meeting a larger group and get some perspective from others.

pmanko2:
> Looks good, thanks @pmanko2!
> 
> The logic for processing and mapping search parameters still is quite complex, since we seem to be doing a subsequent mappings between different representations of the parameters for each Resource.
> 
> This is my first time taking a close look at the search functionality, so perhaps it would be useful to bring this up and discuss in the next FHIR meeting a larger group and get some perspective from others.

Yeah agreed, i'd love to discuss whether we need both "SearchParameter" classes and the SearchParameterMap class that is used within the daos to create criteria queries. This implementation allows us to define a single place that a SearchParameterMap object is configured for each Resource. We don't have to re-specify the different key/property constants required by SearchParamMap anytime we want to make a Resource search request. However, there are already multiple services that accept a SearchParameterMap as the only parameter for their search method, so it might be worth exploring standardizing around that.

api/src/main/java/org/openmrs/module/fhir2/api/search/param/BaseResourceSearchParams.java
ibacher: Hmmmm... yeah. It might make sense to implement some common parameters, but I'd want to discuss how we do it. The only two common search parameters that are unproblematically acceptable for all types are `_id` and `_lastUpdated`.

api/src/main/java/org/openmrs/module/fhir2/api/FhirObservationService.java
```diff
@@ -9,35 +9,21 @@
...
-	        SortSpec sort, HashSet<Include> includes, HashSet<Include> revIncludes);
+	IBundleProvider searchForObservations(ObservationSearchParams observationSearchParams);
 	
-	IBundleProvider getLastnObservations(NumberParam max, ReferenceAndListParam patientReference, TokenAndListParam category,
-	        TokenAndListParam code);
+	IBundleProvider getLastnObservations(NumberParam max, ObservationSearchParams observationSearchParams);
 	
-	IBundleProvider getLastnEncountersObservations(NumberParam max, ReferenceAndListParam patientReference,
-	        TokenAndListParam category, TokenAndListParam code);
+	IBundleProvider getLastnEncountersObservations(NumberParam max, ObservationSearchParams observationSearchParams);
```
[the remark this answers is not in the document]
ibacher: That's fine. For `$lastn` I think we only implemented [the minimum requirements](https://www.hl7.org/fhir/observation-operation-lastn.html).

ibacher:
> i'd love to discuss whether we need both "SearchParameter" classes and the SearchParameterMap class

We don't, but I thought I'd make this a series of smaller steps, e.g., we add the `SearchParameter` classes and once those are all in place we can kibosh `SearchParameterMap` altogether. It has the bonus of having real types.

api/src/main/java/org/openmrs/module/fhir2/api/search/param/BaseResourceSearchParams.java
pmanko2: are you ok with me leaving the common params as is for now? sounds like we'll at least have _id and _lastUpdated in that class, so imo it can stick around even if we pull out some params that are currently defined there in the future.

pmanko2:
@ibacher fixed up the import issues, let me know if there's anything else that needs to be looked at

PR #409, opened 2022-06-08
FM2-476: Add support for searching Medication Requests based on status

mherman22:

Updated the following in order to support searching for MedicationRequests based on status(https://www.hl7.org/fhir/medicationrequest-definitions.html#MedicationRequest.status);-

- MedicationRequestFhirResourceProvider(s).
- Service.
-  Dao.
- unit tests.

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-476

  No? -> execute above command


mherman22:
Still a WIP

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMedicationRequestDaoImpl.java
```diff
@@ -47,13 +53,33 @@ protected void setupSearchParams(Criteria criteria, SearchParameterMap theParams
...
 			}
 		});
 	}
 	
+	private Optional<Criterion> handleMedicationRequestStatus(TokenAndListParam tokenAndListParam) {
+		return handleAndListParam(tokenAndListParam, token -> {
+			if (token.getValue() != null) {
+				try {
+					return Optional.of(
+					    eq("status", MedicationRequest.MedicationRequestStatus.valueOf(token.getValue().toUpperCase())));
```
ibacher, line 72: Is there a `status` field on the OpenMRS model for `DrugOrder`?

ibacher (commented):
So, first @mherman22 thanks for the effort here.

There are a few problems:

1. There is no `status` field in the OpenMRS data model. In 2.2+ there is a `fulfilerStatus`. So, at the very least, while we should add a search parameter for `MedicationRequest`, the implementation needs to go in a folder targetted at the 2.2+ release of OpenMRS.
2. It's not clear that `fulfillerStatus` really maps to the FHIR MedicaitonRequest `status` field. The only value they have in common is `completed`. This implies that the `status` we need to search on needs to be generated from more data than just a single column search.
3. Ultimately, I don't think this ticket should've been marked "Ready for Work" because it really isn't given 2 above.

For that reason, I'm going to close this PR and move the ticket to "Needs Design."

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMedicationRequestDaoImpl.java
mherman22: > Is there a `status` field on the OpenMRS model for `DrugOrder`?

not really , drug order doesn't seem to have a status anywhere on the [data model](https://docs.openmrs.org/datamodel/)

mherman22:
> So, first @mherman22 thanks for the effort here.
> 
> There are a few problems:
> 
> 1. There is no `status` field in the OpenMRS data model. In 2.2+ there is a `fulfilerStatus`. So, at the very least, while we should add a search parameter for `MedicationRequest`, the implementation needs to go in a folder targetted at the 2.2+ release of OpenMRS.
> 2. It's not clear that `fulfillerStatus` really maps to the FHIR MedicaitonRequest `status` field. The only value they have in common is `completed`. This implies that the `status` we need to search on needs to be generated from more data than just a single column search.
> 3. Ultimately, I don't think this ticket should've been marked "Ready for Work" because it really isn't given 2 above.
> 
> For that reason, I'm going to close this PR and move the ticket to "Needs Design."

- After looking through the openmrs model, it is now clear to me that the Drug order doesn't have a status as you have stated above.
- However i beg to be corrected if am wrong @ibacher , according to the [medicationRequestStatusTranslator implementation class](https://github.com/openmrs/openmrs-module-fhir2/blob/master/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/MedicationRequestStatusTranslatorImpl.java), in the [MedicationRequest.MedicationRequestStatus toFhirResource](https://github.com/openmrs/openmrs-module-fhir2/blob/f20accf25fc497b518c9d22ca4027fe9a773d241/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/MedicationRequestStatusTranslatorImpl.java#L23) function, i thought the whole point there is that;-

1. if the Drug Order is **active** then it can be translated as **ACTIVE** in MedicationRequest.MedicationRequestStatus.
2. if it **isDiscontinuedRightNow** then it can be translated as **CANCELLED** to MedicationRequest.MedicationRequestStatus.
3. if it is null then subsquently it can be translated as NULL.

ibacher:
@mherman22 Yep, that's the logic I was looking for! So, can you get the search to work with the same criteria? Also, we'll need some tests that actually prove that the search works.

mherman22:
> @mherman22 Yep, that's the logic I was looking for! So, can you get the search to work with the same criteria? Also, we'll need some tests that actually prove that the search works. PS, I wouldn't worry about searching for `null` as that's not a meaningful search, just if the value is `CANCELLED` it will match requests that are cancelled (according to the translator) and if the value is `ACTIVE` if will match requests that are active (according to the translator).

jumping into that work now, thanks for the clarity

mherman22:
Paving way for another contributor to bring this home

PR #410, opened 2022-06-08
Task service should persist all the Task.basedOn Refs and Task.output Values

mozzy11:
## Issue I worked on

see https://issues.openmrs.org/browse/FM2-483

  No? -> execute above command


integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceIntegrationTest.java
```diff
@@ -154,6 +154,7 @@ public void shouldCreateNewTaskAsJson() throws Exception {
 		assertThat(task.getAuthoredOn(), within(1, ChronoUnit.MINUTES, new Date()));
 		assertThat(task.getLastModified(), within(1, ChronoUnit.MINUTES, new Date()));
 		assertThat(task, validResource());
+		assertThat(task.getBasedOn().size(),is(2));
```
kdaud, line 157:
```suggestion
		assertThat(task.getBasedOn().size(), is(2));
```

I am learning from your experience @mozzy11. Though have made a minor suggestion my intention is to learn something here concerning the number of tasks that should be expected by the test. The test is expecting 2 tasks and I see only one order has been added in the xml file, how does this add up to 2?

mozzy11: Thanks @kdaud 
we both learn from each other.  :roll_eyes:
regarding the number of oders , one oder already existed , i just added a second one when fixing this issue

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceIntegrationTest.java
```diff
@@ -154,7 +154,7 @@ public void shouldCreateNewTaskAsJson() throws Exception {
 		assertThat(task.getAuthoredOn(), within(1, ChronoUnit.MINUTES, new Date()));
 		assertThat(task.getLastModified(), within(1, ChronoUnit.MINUTES, new Date()));
 		assertThat(task, validResource());
-		assertThat(task.getBasedOn().size(),is(2));
+		assertThat(task.getBasedOn().size(), is(2));
```
kdaud, line 157: > regarding the number of oders , one oder already existed , i just added a second one when fixing this issue

Oh I see! Then I think this would mean the test should fail if the expected value isn't 2. Have been playing around with the test and found out that it passes the build on whatever value is passed in the matcher object ie `is(x)`.

mozzy11: its beacuse youre simply running `mvn clean install` locally.
by default the [maven-surefire-plugin plugin](https://github.com/openmrs/openmrs-module-fhir2/blob/master/pom.xml#L555-L562) is configured to skip the [ intergration tests](https://github.com/openmrs/openmrs-module-fhir2/blob/master/pom.xml#L560) .

try running with this [profile](https://github.com/openmrs/openmrs-module-fhir2/blob/master/pom.xml#L801).
or run the test directly in your IDE  .see  screenshot of my IDE 
image

mozzy11: Note that the Intergration tests are ran in the [github action CI build ](https://github.com/openmrs/openmrs-module-fhir2/blob/master/.github/workflows/main.yml#L36)

api/src/main/java/org/openmrs/module/fhir2/model/FhirReference.java
```diff
@@ -50,6 +50,7 @@ public FhirReference() {
 	@Column(name = "target_uuid")
 	private String targetUuid;
 	
+	@EqualsAndHashCode.Include
```
ibacher, line 53: I'd prefer if we didn't do this. Honestly, I'm not even all that keen on having the `id` field annotated this way.

```suggestion
```

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceIntegrationTest.java
mozzy11: Tty this `mvn verify --batch-mode -P integration-test --file pom.xml `locally , youll see the failure.
image

test-data/src/test/resources/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImplTest_initial_data.xml
```diff
@@ -24,6 +24,7 @@
...
 	<orders order_id="101" order_type_id="100" order_number="ORD-101" urgency="ROUTINE" order_action="NEW" concept_id="5497" orderer="1" date_activated="2008-11-19 09:24:10.0" patient_id="2" care_setting="1" encounter_id="6" uuid="7d96f25c-4949-4f72-9931-d808fbc226de" date_created="2008-11-19 09:24:10.0" creator="1" voided="0" />
 	<orders order_id="102" order_type_id="100" order_number="ORD-101" urgency="ROUTINE" order_action="NEW" concept_id="5497" orderer="1" date_activated="2008-11-19 09:24:10.0" patient_id="2" care_setting="1" encounter_id="6" uuid="cbcb84f3-4576-452f-ba74-7cdeaa9aa602" date_created="2008-11-19 09:24:10.0" creator="1" voided="0" />
+	<orders order_id="103" order_type_id="100" order_number="ORD-101" urgency="ROUTINE" order_action="NEW" concept_id="5497" orderer="1" date_activated="2008-11-19 09:24:10.0" patient_id="2" care_setting="1" encounter_id="6" uuid="7d96f25c-4949-4f72-9931-d808fbc226dd" date_created="2008-11-19 09:24:10.0" creator="1" voided="0" />
```
ibacher, line 27: I tend to prefer that we generate completely new random UUIDs rather than just modify one character from one. This tends to make debugging more obvious, so could we use `a7fd2d2a-e068-4c45-b613-8261d70b29b7` where appropriate?

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceIntegrationTest.java
```diff
@@ -154,6 +154,7 @@ public void shouldCreateNewTaskAsJson() throws Exception {
 		assertThat(task.getAuthoredOn(), within(1, ChronoUnit.MINUTES, new Date()));
 		assertThat(task.getLastModified(), within(1, ChronoUnit.MINUTES, new Date()));
 		assertThat(task, validResource());
+		assertThat(task.getBasedOn().size(), is(2));
```
ibacher, line 157:
```suggestion
		assertThat(task.getBasedOn(), hasSize(2));
```

ibacher (changes requested):
Thanks @mozzy11! A couple of small points and this should be fine to merge.

api/src/main/java/org/openmrs/module/fhir2/model/FhirReference.java
mozzy11: @ibacher removing this annoation makes things fail .
it literally means the object hashcode is simply generated based on the Object [Id](https://github.com/openmrs/openmrs-module-fhir2/blob/master/api/src/main/java/org/openmrs/module/fhir2/model/FhirReference.java#L41).
and given that id is null before the object is persisted , the [set](https://github.com/openmrs/openmrs-module-fhir2/blob/master/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImpl.java#L151-L152) would assume all these Reference Objects are equls so would deduplicate the References into one

mozzy11: The other alternative is to change the [Set](https://github.com/openmrs/openmrs-module-fhir2/blob/master/api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java#L85) to List

mozzy11: > Then we really should just get rid of the other EqualsAndHashCode.Include and fall back to using the UUID hash code...

Alright , let get rid of them

mozzy11:
cc @ibacher

api/src/main/java/org/openmrs/module/fhir2/model/FhirReference.java
```diff
@@ -27,7 +27,7 @@
  * FHIR Reference - https://www.hl7.org/fhir/references.html
  */
 @Data
-@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = false)
+@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
```
ibacher, line 30:
```suggestion
```

api/src/main/java/org/openmrs/module/fhir2/model/FhirTaskOutput.java
```diff
@@ -25,7 +25,7 @@
  */
 @Data
 @NoArgsConstructor
-@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = false)
+@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
```
ibacher, line 28:
```suggestion
```

mozzy11: @ibacher , removing `@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)` doesnt solve the issue. for some reason it still fails.
setting `callSuper = true ` solves it.

PR #412, opened 2022-06-12
FM2-485: Add District Mapping to person address translator

slubwama:
FM2-485 Add District Mapping to person address translator
https://issues.openmrs.org/browse/FM2-485


ibacher (approved):

ibacher:
It would also be a good idea to do this for `LocationAddress`, but I'll leave that aside for now. Thanks!

PR #414, opened 2022-06-19
FM2-64: add support for UCUM system units in Observation quantities

pmanko2:
move CodingTranslator from interface -> abstract class. use for UCUM system resolving

test for observation quantity coding translator

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-64

  No? -> execute above command


api/src/main/java/org/openmrs/module/fhir2/api/translators/CodingTranslator.java
```diff
@@ -33,5 +41,26 @@ public interface CodingTranslator extends ToFhirTranslator<Concept, Coding>, ToO
 	 * @return the corresponding OpenMRS {@link Concept}
 	 */
 	@Override
-	Concept toOpenmrsType(@Nonnull ICoding coding);
+	public Concept toOpenmrsType(@Nonnull ICoding coding) {
```
pmanko2, line 44: changed CodingTranslator from an interface -> abstract class to avoid duplicating the toOpenmrsType method. It feels like this toOpenmrsType() implementation is generic enough that it can be used by most CodingTranslator extending classes.

`getCodingForSystem` is another method that i think could be usefule for any CodingTranslator

pmanko2: happy to revert back to the previous interface definition if need be, though

api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
```diff
@@ -319,4 +319,6 @@ private FhirConstants() {
 	public static final String SNOMED_SYSTEM_URI = "http://snomed.info/sct";
 	
 	public static final String RX_NORM_SYSTEM_URI = "http://www.nlm.nih.gov/research/umls/rxnorm";
+	
+	public static final String UCUM_SYSTEM_URI = "http://unitsofmeasure.org";
```
pmanko2, line 323: don't love defining these as constants in the fhir2 module. it feels like the source of truth for these system uris should be somewhere else (first thought is core openmrs?)

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationQuantityCodingTranslatorImpl.java
```diff
@@ -0,0 +1,68 @@
...
+		
+		for (UcumService ucumService : ucumServices) {
+			String code = conceptNumeric.getUnits();
+			if (code != null && ucumService.validate(code) == null) {
+				coding = new Coding(UCUM_SYSTEM_URI, code, ucumService.getCommonDisplay(code));
+				break;
+			}
+		}
+		
+		// attempt to fall back to the concept uuid with a null coding concept
```
pmanko2, line 60: wasn't sure what the correct behavior should be here. initially had it return null if a UCUM valid code is not provided, but seems like the default fallback for the MedicationCodingTranslator was uuid with null system so implemented that here as well. easy enough to change if we want to only return UCUM system codes.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationValueTranslatorImpl.java
```diff
@@ -60,6 +58,15 @@ public Type toFhirResource(@Nonnull Obs obs) {
 			if (obs.getConcept() instanceof ConceptNumeric) {
 				ConceptNumeric cn = (ConceptNumeric) obs.getConcept();
 				result.setUnit(cn.getUnits());
+				
+				// only set the coding system if unit conforms to UCUM standard
+				Coding coding = quantityCodingTranslator.toFhirResource(cn);
+				if (coding.hasSystem() && coding.getSystem().equals(UCUM_SYSTEM_URI)) {
```
pmanko2, line 64: ignoring every system but UCUM. is this the correct approach or do we want to use a fallback system for observation quantity?

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationValueTranslatorImpl.java
```diff
@@ -60,6 +58,15 @@ public Type toFhirResource(@Nonnull Obs obs) {
...
+					String codingSystem = coding.getSystem();
+					result.setCode(coding.getCode());
+					result.setSystem(codingSystem);
+					result.getCodeElement().setSystem(codingSystem);
```
pmanko2, line 68: not sure if CodeElement.system needs to be set as well but figured i'd be thorough about it

api/src/main/resources/moduleApplicationContext.xml
```diff
@@ -34,4 +35,12 @@
     <bean name="fhirR4" class="ca.uhn.fhir.context.FhirContext" factory-method="forR4">
         <property name="restfulClientFactory" ref="restfulClientFactoryR4"/>
     </bean>
+    <!--    base ucum-essence file packaged with fhir-ucum-->
```
pmanko2, line 38: couple of things here:

1. the fhir-ucum ucum-essence.xml file is currently outdated, i'll put up a PR to update to the latest provided file from 2017

2. imo we should treat that module as the main source of truth for UCUM translation instead of duplicating the core ucum-essence file in fhir2, but open to a discussion on that approach

api/src/main/resources/ucum-fhir-essence.xml
```diff
@@ -0,0 +1,17 @@
...
+    Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+    graphic logo is a trademark of OpenMRS Inc.
+-->
+<root version="2.1" revision="$Revision: 442 $" revision-date="$Date: 2017-11-21 19:04:52 -0500 (Tue, 21 Nov 2017) $">
+    <unit Code="mmHg" CODE="MHG" isMetric="yes" class="clinical">
+        <name>meter of mercury column</name>
+        <printSymbol>mmHg</printSymbol>
+        <property>pressure</property>
+        <value Unit="kPa" UNIT="KPAL" value="133322">133322</value>
+    </unit>
```
pmanko2, line 16: update this

api/src/main/resources/ucum-fhir-essence.xml
```diff
@@ -0,0 +1,17 @@
+<!--
+    This Source Code Form is subject to the terms of the Mozilla Public License,
+    v. 2.0. If a copy of the MPL was not distributed with this file, You can
+    obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+    the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+
+    Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+    graphic logo is a trademark of OpenMRS Inc.
+-->
+<root version="2.1" revision="$Revision: 442 $" revision-date="$Date: 2017-11-21 19:04:52 -0500 (Tue, 21 Nov 2017) $">
```
pmanko2, line 10: this additional ucum-essence.xml is where we can define any UCUM definitions that aren't covered by the standard file provided by unitsofmeasure.org. currently we only define mmHg since it's used frequently in observation quantities (and the UCUM standard representation is mm[Hg])

Not a huge fan of this approach though since it feels like we're creating our own standardized library that now doesn't conform to UCUM definitions. I think maybe the better solution might be to extend ucum-fhir parsing to validate/parse via the <printSymbol> definition as well. but that project doesn't seem to be maintained regularly and this approach has a smaller barrier to entry so 🤷‍♂️ . Would be interested to hear opinions though

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImplTest.java
```diff
@@ -11,7 +11,10 @@
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.*;
```
pmanko2, line 14: tsk tsk @pmanko

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/ObservationQuantityCodingTranslatorImplTest.java
```diff
@@ -0,0 +1,196 @@
...
+		conceptTranslator.setConceptSourceService(conceptSourceService);
+		
+		quantityCodingTranslator.setConceptTranslator(conceptTranslator);
+	}
+	
+	@Test
+	public void toFhirResource_shouldTranslateConceptWithCommonUCUMUnitCodes() {
+		ConceptNumeric cn = new ConceptNumeric();
+		
+		cn.setUnits("cm");
```
pmanko2, line 66: maybe we should extend this to cover all the units defined in https://www.hl7.org/fhir/valueset-ucum-common.html , but i just included some common units to start us off (mostly those called out as important in FM2-64)

api/src/main/java/org/openmrs/module/fhir2/api/translators/CodingTranslator.java
```diff
@@ -33,5 +41,26 @@ public interface CodingTranslator extends ToFhirTranslator<Concept, Coding>, ToO
...
+			codeableConcept.addCoding(new Coding(coding.getSystem(), coding.getCode(), coding.getDisplay()));
+			return conceptTranslator.toOpenmrsType(codeableConcept);
+		}
+		return null;
+	}
+	
+	/**
+	 * @return the coding on the CodeableConcept with the given system, or null if none found.
+	 */
+	protected Coding getCodingForSystem(CodeableConcept codeableConcept, String system) {
```
pmanko2, line 56: hmm wonder if this can be a public static method since its not really tied to an instance of a CodingTranslator/might be useful from outside of a CodingTranslator as well?

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/ObservationQuantityCodingTranslatorImplTest.java
```diff
@@ -0,0 +1,166 @@
...
+	}
+	
+	/**
+	 * Tests all common fhir ucum codes specified in https://www.hl7.org/fhir/valueset-ucum-common.html
+	 */
+	@Test
+	public void toFhirResource_shouldTranslateConceptWithCommonUCUMUnitCodes() throws IOException {
+		ConceptNumeric cn = new ConceptNumeric();
+		
+		try (CSVReader csvReader = new CSVReader(
```
pmanko2, line 74: i'd love to run this with a ParameterizedTest but i think we need to either update to Junit5 or Spring 4.2 which introduces the SpringClassRule junit4 rule

api/src/main/java/org/openmrs/module/fhir2/api/translators/CodingTranslator.java
```diff
@@ -11,11 +11,19 @@
...
+import org.hl7.fhir.r4.model.CodeableConcept;
 import org.hl7.fhir.r4.model.Coding;
 import org.hl7.fhir.r4.model.ICoding;
 import org.openmrs.Concept;
+import org.openmrs.util.OpenmrsUtil;
+import org.springframework.beans.factory.annotation.Autowired;
 
-public interface CodingTranslator extends ToFhirTranslator<Concept, Coding>, ToOpenmrsTranslator<Concept, ICoding> {
+@Setter
+public abstract class CodingTranslator implements ToFhirTranslator<Concept, Coding>, ToOpenmrsTranslator<Concept, ICoding> {
```
ibacher, line 23: Do we need to turn this into an abstract class? I'd rather leave it as an interface and move the coding here into the actual implementation class.

api/src/main/java/org/openmrs/module/fhir2/api/translators/CodingTranslator.java
ibacher: Yes, actually. I think adding this as a utility method somewhere would be useful.

api/src/main/java/org/openmrs/module/fhir2/api/translators/CodingTranslator.java
ibacher: The usual pattern we use for this would be to add a `BaseCodingTranslator` class that we could extend for the actual implementations. (The idea of the interfaces is to keep things as agnostic of the underlying code as possible; this enables implementations to swap out our implementations for... w/e they want—otherwise, we wouldn't have all these interfaces!)

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationQuantityCodingTranslatorImpl.java
ibacher: This seems good to me!

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationValueTranslatorImpl.java
ibacher: I think having the hard-coded to UCUM seems like it will work. Most real-world uses probably only care about what's in the `unit` field... The coded representation of the unit is basically useful if you want to use a coding system to translate into another unit... and the only coding system I'm aware of that has those kinds of unit conversion rules is UCUM.

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/ObservationQuantityCodingTranslatorImplTest.java
ibacher: Heh... we're kind of stuck in the base package here with OpenMRS 2.0.0 which is Spring 4.1 IIRC and JUnit4. Newer versions provide newer versions of those libraries, but then we'd have to drop support for older versions of the OpenMRS platform. One thought here: do we even need Spring in this test at all? I.e., can we drop the `extends BaseModuleContextSensitiveTest` and `@ContextConfiguration` and just make this a plain JUnit 4 test, in which case we can use the `Parameterized` runner...

ibacher (commented):
Hi @pmanko2! Thanks for this! PR... I've filled in a bunch of mostly questions. Let me know what you think!

api/src/main/resources/ucum-fhir-essence.xml
```diff
@@ -0,0 +1,17 @@
...
+    This Source Code Form is subject to the terms of the Mozilla Public License,
+    v. 2.0. If a copy of the MPL was not distributed with this file, You can
+    obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+    the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+
+    Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+    graphic logo is a trademark of OpenMRS Inc.
+-->
+<root version="1.0" revision="$Revision: 1 $" revision-date="$Date: 2022-06-21 16:00:00 -0500 (Tue, 21 June 2022) $">
+    <unit Code="mmHg" CODE="MMHG" isMetric="yes" class="clinical">
```
pmanko2, line 11: @ibacher let me know if there are any other custom definitions we need to add

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/BaseCodingTranslator.java
```diff
@@ -0,0 +1,55 @@
...
+			codeableConcept.addCoding(new Coding(coding.getSystem(), coding.getCode(), coding.getDisplay()));
+			return conceptTranslator.toOpenmrsType(codeableConcept);
+		}
+		return null;
+	}
+	
+	/**
+	 * @return the coding on the CodeableConcept with the given system, or null if none found.
+	 */
+	public static Coding getCodingForSystem(CodeableConcept codeableConcept, String system) {
```
pmanko2, line 45: initial thought was to add this utility function here since its related to coding translation, but we could also move it over to FhirUtils instead

api/src/main/java/org/openmrs/module/fhir2/api/translators/CodingTranslator.java
pmanko2: so i went ahead and switched to a BaseCodingTranslator abstract class and CodingTranslator interface since that's the standard we've been using. let me know if that's what you had in mind.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/BaseCodingTranslator.java
ibacher: I think it's fine to keep it here for now... shouldn't be too disruptive to move it later. Should this method be `protected` though?

api/src/main/resources/ucum-fhir-essence.xml
ibacher: Heh... I think that's something we'll have to figure out from practice. This is a good starting point though!

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/BaseCodingTranslator.java
pmanko2: hmm not sure to be honest. the only benefit of a protected static method in this case would be so we restrict its use to within its package. is that something we want to do? 

i think i'd prefer to keep it public

PR #424, opened 2022-08-09
update Concept translator to map FHIR codes to concepts with any source mapping

mozzy11:
## Issue I worked on

see https://issues.openmrs.org/browse/FM2-501

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConceptDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConceptDaoImpl.java
@@ -70,4 +70,30 @@ public class FhirConceptDaoImpl extends BaseFhirDao<Concept> implements FhirConc
 		
 		return Optional.ofNullable((Concept) criteria.uniqueResult());
 	}
+	
+	@Override
+	public Optional<Concept> getConceptWithAnyMappingInSource(@Nonnull ConceptSource conceptSource,
+	        @Nonnull String mappingCode) {
+		if (conceptSource == null || mappingCode == null) {
+			return Optional.empty();
+		}
+		
+		Criteria criteria = getSessionFactory().getCurrentSession().createCriteria(ConceptMap.class);
+		criteria.setProjection(property("concept"));
+		criteria.createAlias("conceptReferenceTerm", "term");
+		criteria.createAlias("conceptMapType", "mapType");
+		criteria.createAlias("concept", "concept");
+		
+		if (Context.getAdministrationService().isDatabaseStringComparisonCaseSensitive()) {
+			criteria.add(eq("term.code", mappingCode).ignoreCase());
+		} else {
+			criteria.add(eq("term.code", mappingCode));
+		}
+		
+		criteria.add(eq("term.conceptSource", conceptSource));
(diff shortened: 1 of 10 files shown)


api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConceptDaoImpl.java
```diff
@@ -70,4 +70,30 @@ public Optional<Concept> getConceptWithSameAsMappingInSource(@Nonnull ConceptSou
...
+			criteria.add(eq("term.code", mappingCode).ignoreCase());
+		} else {
+			criteria.add(eq("term.code", mappingCode));
+		}
+		
+		criteria.add(eq("term.conceptSource", conceptSource));
+		criteria.addOrder(asc("concept.retired"));
+		criteria.setResultTransformer(DistinctRootEntityResultTransformer.INSTANCE);
+		
+		return Optional.ofNullable((Concept) criteria.uniqueResult());
```
ibacher, line 97: `uniqueResult()` throws an exception if you get more than one result which is, I think, not what we want here.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConceptTranslatorImpl.java
```diff
@@ -103,6 +103,11 @@ public Concept toOpenmrsType(@Nonnull CodeableConcept concept) {
 							    coding.getCode());
 							if (c.isPresent()) {
 								return c.get();
+							} else {
+								c = conceptService.getConceptWithAnyMappingInSource(conceptSource.get(), coding.getCode());
+								if (c.isPresent()) {
+									return c.get();
+								}
```
ibacher, line 110: Do we need two database look-ups here? Maybe we can look into return a DTO that includes both the concept and the mapping type then change the logic around here: if we have a SAME-AS map use that, etc. as before?

ibacher (commented):
Thanks @mozzy11! A couple of thoughts on what we're doing here.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConceptTranslatorImpl.java
mozzy11: Thanks @ibacher .
updated

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirConceptDaoImpl.java
```diff
@@ -70,4 +72,30 @@ public Optional<Concept> getConceptWithSameAsMappingInSource(@Nonnull ConceptSou
 		
 		return Optional.ofNullable((Concept) criteria.uniqueResult());
 	}
+	
+	@Override
+	public List<Concept> getConceptsWithAnyMappingInSource(@Nonnull ConceptSource conceptSource,
```
ibacher, line 77: Can you add a test or two for this DAO method? Then I think this is good to merge.

mozzy11: Thanks @ibacher , had forgoten to add tests for that one .
added

PR #431, opened 2022-09-29
FM2-514 - Add back FhirConceptSourceDao.getFhirConceptSourceByConcept…

mseaton:


mseaton:
Actually, looks like you beat me to it :)
I did change the method implementation somewhat in this commit, to be more consistent with the behavior found in the get by hl7 code method (handling case sensitive comparisons, allowing for sorting by retired, etc)

I'll close out this PR since you've already done it @ibacher , but it's here for posterity :)

PR #438, opened 2022-10-26
FM2-526: MedicationRequest: support revInclude of Medication Dispense

mogoodrich:


api/src/main/java/org/openmrs/module/fhir2/api/FhirMedicationRequestService.java
```diff
@@ -28,5 +28,5 @@ public interface FhirMedicationRequestService extends FhirService<MedicationRequ
 	IBundleProvider searchForMedicationRequests(ReferenceAndListParam patientReference,
 	        ReferenceAndListParam encounterReference, TokenAndListParam code, ReferenceAndListParam participantReference,
 	        ReferenceAndListParam medicationReference, TokenAndListParam id, DateRangeParam lastUpdated,
-	        HashSet<Include> includes);
```
mogoodrich, line 31: I've changed the signature of this service method... is this okay (ie is this service expected to be used anywhere outside this module)?  If not, easy enough to create a new service method with an additional parameter as well.

api/src/main/java/org/openmrs/module/fhir2/providers/r3/MedicationRequestFhirResourceProvider.java
```diff
@@ -119,7 +120,9 @@ public IBundleProvider searchForMedicationRequests(
 	        @IncludeParam(allow = { "MedicationRequest:" + MedicationRequest.SP_MEDICATION,
 	                "MedicationRequest:" + MedicationRequest.SP_REQUESTER,
 	                "MedicationRequest:" + MedicationRequest.SP_PATIENT,
-	                "MedicationRequest:" + MedicationRequest.SP_CONTEXT }) HashSet<Include> includes) {
+	                "MedicationRequest:" + MedicationRequest.SP_CONTEXT }) HashSet<Include> includes,
+	        @IncludeParam(reverse = true, allow = {
+	                "MedicationDispense:" + MedicationDispense.SP_PRESCRIPTION }) HashSet<Include> revIncludes) {
```
mogoodrich, line 125: Note that MedicationDispense is not available in OpenMRS until 2.6.  If you try to make a request that includes this revinclude request on an earlier version of OpenMRS, you get a cryptic error. 

However, I didn't see a pattern to introduce a "MedicationRequestFhirResourceProvider2_6"... I'm willing to work through how to do this if we think this is a better practice.

api/src/main/java/org/openmrs/module/fhir2/providers/r4/MedicationRequestFhirResourceProvider.java
```diff
@@ -115,7 +116,9 @@ public IBundleProvider searchForMedicationRequests(
 	        @IncludeParam(allow = { "MedicationRequest:" + MedicationRequest.SP_MEDICATION,
 	                "MedicationRequest:" + MedicationRequest.SP_REQUESTER,
 	                "MedicationRequest:" + MedicationRequest.SP_PATIENT,
-	                "MedicationRequest:" + MedicationRequest.SP_ENCOUNTER }) HashSet<Include> includes) {
+	                "MedicationRequest:" + MedicationRequest.SP_ENCOUNTER }) HashSet<Include> includes,
+	        @IncludeParam(reverse = true, allow = {
+	                "MedicationDispense:" + MedicationDispense.SP_PRESCRIPTION }) HashSet<Include> revIncludes) {
```
mogoodrich, line 121: Same issue as comment above.

mogoodrich (commented):
Ready for review... I added a couple comments that I was interested in your thoughts on @ibacher

mogoodrich:
FWIW, all the lines that are flagged above as not covered by tests are covered at the integration level by searchForMedicationRequest_shouldReturnMedicationRequestByPatientUuidAndRevIncludeAssociatedMedicationDispense in MedicationRequestSearchQuery2_6Test.  I didn't see any existing unit tests for SearchQueryInclude?

api/src/main/java/org/openmrs/module/fhir2/providers/r3/MedicationRequestFhirResourceProvider.java
mseaton: What kind of cryptic error @mogoodrich ?  I would think that - in pre-2.6 systems that the request to "get all associated MedicationDispenses" would just return an empty set, since none are there.  I don't see anything in this particular interface that is OpenMRS-version dependent.  If there is an issue, I assume it is in the implementation/services of getting and translation the MedicationDispenses, and I would assume we can just do a check there to return empty if they aren't supported?

api/src/main/java/org/openmrs/module/fhir2/providers/r4/MedicationRequestFhirResourceProvider.java
mseaton: Same response as above.

api/src/main/java/org/openmrs/module/fhir2/api/search/SearchQueryInclude.java
```diff
@@ -278,13 +286,23 @@ private IBundleProvider handleEncounterReverseInclude(ReferenceAndListParam para
...
 		
 		return null;
 	}
 	
+	private IBundleProvider handlePrescriptionReverseInclude(ReferenceAndListParam params, String targetType) {
+		switch (targetType) {
+			case FhirConstants.MEDICATION_DISPENSE:
+				MedicationDispenseSearchParams medicationDispenseSearchParams = new MedicationDispenseSearchParams();
+				medicationDispenseSearchParams.setMedicationRequest(params);
+				return medicationDispenseService.searchMedicationDispenses(medicationDispenseSearchParams);
```
mseaton, line 301: Presumably this would be the place to decide if MedicationDispense is supported or not and to do the right thing?  Is the service not capable of doing that?

mseaton (approved):
A few comments, looks mostly good to me.

api/src/main/java/org/openmrs/module/fhir2/providers/r3/MedicationRequestFhirResourceProvider.java
mogoodrich: Good call @mseaton re: looking into the error... this may be an easy fix, will explore more tomorrow.

api/src/main/java/org/openmrs/module/fhir2/api/search/SearchQueryInclude.java
mogoodrich: So you were correct @mseaton this *was* returning an empty bundle if MedicationDispense is not supported.... I modified line 157 above to check if the provider is empty and this fixes the issue... now if you request a MedicationRequest with a MedicationDispense revinclude against a pre 2.6.x core, it doesn't crash but simply doesn't return any MedicationDispense events.

api/src/main/java/org/openmrs/module/fhir2/api/FhirMedicationRequestService.java
ibacher: The general rule is that these service classes are not general service classes. For the most part they are supposed to the the things called by the ResourceProviders. Basically, the ResourceProviders provide the FHIR REST API and the service classes translate that into other calls. As long as the functionality still works, it should be fine to change the signatures like this.

api/src/main/java/org/openmrs/module/fhir2/api/search/SearchQueryInclude.java
ibacher: Yeah, this is my interpretation of how the FHIR API is meant to be implemented, i.e., in general we should accept as many requests as possible and just ignore parts of the request we can't process.

ibacher (approved):
This looks good to me. It would be nice, though not essential, if this included an IntegrationTest (the difference being that the IntegrationTest doesn't mock out the other layers, so checks against the H2 database).

mogoodrich:
> This looks good to me. It would be nice, though not essential, if this included an IntegrationTest (the difference being that the IntegrationTest doesn't mock out the other layers, so checks against the H2 database).

Oh, definitely agree we want an Integration Test, but that's what MedicationRequestSearchQueryImpl_2_6Test.java is meant to cover... is there something I should add somewhere else as well? @ibacher

PR #443, opened 2022-11-09
FM2-530 - MedicationRequestTimingTranslator should support OrderFrequ…

mseaton:
…ency reference

Not sure why the formatter changed the imports on a few files.  That came from building the code.

@mogoodrich - try this out an confirm it meets your use case?

@ibacher anything else needed here?


mogoodrich:
Thanks @mseaton  I will test and confirm this works...

mogoodrich (approved):
LGTM, and I confirmed that is works, thanks @mseaton !

PR #447, opened 2022-11-17
FM2-532: MedicationRequest endpoint should not return Discontinue Orders

mogoodrich:
## Issue I worked on

  No? -> execute above command

### DIFF
--- a/api-2.6/src/test/java/org/openmrs/module/fhir2/api/search/EncounterSearchQueryImpl_2_6Test.java
+++ b/api-2.6/src/test/java/org/openmrs/module/fhir2/api/search/EncounterSearchQueryImpl_2_6Test.java
@@ -9,16 +9,6 @@
  */
 package org.openmrs.module.fhir2.api.search;
 
-import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.iterableWithSize;
-import static org.hamcrest.Matchers.notNullValue;
-
-import java.util.HashSet;
-import java.util.List;
-import java.util.stream.Collectors;
-
 import ca.uhn.fhir.model.api.Include;
 import ca.uhn.fhir.rest.api.server.IBundleProvider;
 import ca.uhn.fhir.rest.param.TokenAndListParam;
@@ -37,6 +27,16 @@ import org.openmrs.module.fhir2.api.translators.EncounterTranslator;
 import org.openmrs.test.BaseModuleContextSensitiveTest;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.test.context.ContextConfiguration;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.iterableWithSize;
+import static org.hamcrest.Matchers.notNullValue;
 
 @ContextConfiguration(classes = TestFhirSpringConfiguration.class, inheritLocations = false)
 public class EncounterSearchQueryImpl_2_6Test extends BaseModuleContextSensitiveTest {
@@ -93,11 +93,11 @@ public class EncounterSearchQueryImpl_2_6Test extends BaseModuleContextSensitive
 		
 		assertThat(results, notNullValue());
 		
-		assertThat(resultList.size(), equalTo(10)); // reverse included resources added as part of the result list
+		assertThat(resultList.size(), equalTo(8)); // reverse included resources added as part of the result list
 		assertThat(resultList.stream().filter(result -> result instanceof Encounter).collect(Collectors.toList()),
 		    is(iterableWithSize(1))); // the actual matched encounter
 		assertThat(resultList.stream().filter(result -> result instanceof MedicationRequest).collect(Collectors.toList()),
-		    is(iterableWithSize(8))); // 8 requests that reference that encounter
+		    is(iterableWithSize(6))); // 8 requests that reference that encounter
 		assertThat(resultList.stream().filter(result -> result instanceof MedicationDispense).collect(Collectors.toList()),
 		    is(iterableWithSize(1))); // 1 dispense that references the above requests
 	}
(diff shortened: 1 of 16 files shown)


api-2.6/src/test/java/org/openmrs/module/fhir2/api/search/EncounterSearchQueryImpl_2_6Test.java
```diff
@@ -93,11 +93,11 @@ public void searchForEncounters_shouldReverseIncludeMedicationRequestsAndAssocia
 		
 		assertThat(results, notNullValue());
 		
-		assertThat(resultList.size(), equalTo(10)); // reverse included resources added as part of the result list
```
mogoodrich, line 96: there were two orders of type "DISCONTINUE" in the core test data set, so there are a lot of test cases where the expected result is reduced by 2

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseEncounterDao.java
```diff
@@ -101,6 +102,7 @@ protected void handleHasAndListParam(Criteria criteria, HasAndListParam hasAndLi
 							// Constrain only on non-voided Drug Orders
 							criteria.add(Restrictions.eq("orders.class", DrugOrder.class));
 							criteria.add(Restrictions.eq("orders.voided", false));
+							criteria.add(Restrictions.ne("orders.action", Order.Action.DISCONTINUE));
```
mogoodrich, line 105: this makes sure the Encounter "has" MedicationRequests functionality doesn't pull back encounters than only have "discontinue" orders.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMedicationRequestDaoImpl.java
```diff
@@ -14,16 +14,59 @@
...
+import javax.annotation.Nonnull;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
 
 @Component
 @Setter(AccessLevel.PACKAGE)
 public class FhirMedicationRequestDaoImpl extends BaseFhirDao<DrugOrder> implements FhirMedicationRequestDao {
 	
+	@Override
```
mogoodrich, line 35: For the two "get" and the "getSearcResults" methods I went back and forth on replacing the BaseFhirDAO entirely with a new query versus delegating to them and then stripping out any "discontinue" orders.  I went with the second approach (delegating and stripping) because it seemed less intrusive to the original code, and, also, the session factory currently has private access and so wasn't available in this extension class.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMedicationRequestDaoImpl.java
```diff
@@ -52,6 +95,7 @@ protected void setupSearchParams(Criteria criteria, SearchParameterMap theParams
 					break;
 			}
 		});
+		excludeDiscontinueOrders(criteria);
```
mogoodrich, line 98: The "getSearchResultsUuid" method relied on the search params, so instead of overriding it to filter out discontinue orders, I could modify the criteria directly in the setupSearchParams function.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMedicationRequestDaoImpl.java
```diff
@@ -64,4 +108,8 @@ private void handleCodedConcept(Criteria criteria, TokenAndListParam code) {
 		}
 	}
 	
+	private void excludeDiscontinueOrders(Criteria criteria) {
+		// exclude "discontinue" orders, see: https://issues.openmrs.org/browse/FM2-532
+		criteria.add(Restrictions.ne("action", Order.Action.DISCONTINUE));
+	}
```
mogoodrich, line 114: I didn't override the "delete(uuid)" or the "createOrUpdate(DrugOrder)" methods... the delete does a get(uuid) to fetch, so it will now just do nothing if passed the uuid of a discontinue order (which is the same as the current behavior if the uuid didn't match a order at all).

For createOrUpdate, since I didn't touch this, but someone can't create a DISCONTINUE order via FHIR because there's currently no way to represent DISCONTINUE in FHIR.  Someone technically *could* call an update, but this seemed like too much of an edge case to had some convoluted code around... the createOrUpdate would have to check to see if there's an existing drug order with that uuid, and not perform the update if it was type DISCONTINUE.

api/src/test/java/org/openmrs/module/fhir2/api/search/MedicationRequestSearchQueryTest.java
```diff
@@ -497,12 +518,10 @@ public void searchForMedicationRequest_shouldReturnMedicationRequestByParticipan
 		IBundleProvider results = search(theParams);
```
mogoodrich, line 519: This whole test class was kind of strange... each of the tests I modified below first tested that there were the proper number of elements in the result set, but then called a utility get method that fetched the first ten elements in that result set... and then confirmed that return list did have, indeed, 10 elements.

I changed the utility method to return all the elements (see below) and just changed it to  only test once that the proper number of elements were present (generally  2 less than previously because of the 2 DISCONTINUE orders in the test dataset)

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMedicationRequestDaoImpl.java
ibacher: Just an observation, but while the `sessionFactory` itself is a private field, there is a public getter, so you should be able to use `getSessionFactory()` factory here to get a hold of it (of course, this depends on your IDE having the appropriate extension for Lombok or else I think it's kind of hidden).

ibacher (approved):

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMedicationRequestDaoImpl.java
mogoodrich: duh, good point @ibacher .... (still think I will stick the current approach I've gone with, though)

mseaton (approved):
Looks fine to me, though I still find the principle of excluding these hard to square a bit.  My guess is that at some point we are going to need these to be returned somehow - maybe with a custom extension or as a different resource.  For example, if a client wanted to know _why_ a particular order was stopped, is the current way to get this off of the order reason property of the discontinue order?  Anyway, I'm approving, but I'm not sure we've seen the end of this.  @ibacher ?

ibacher:
> For example, if a client wanted to know why a particular order was stopped, is the current way to get this off of the order reason property of the discontinue order?

At least from a FHIR perspective, this would be something like the original order having a status of "stopped" and then having a `statusReason` that explains why it was stopped. That said, the feedback I got is that these D/C orders should be represented as FHIR Tasks... which is fine but doesn't really solve the challenges around expressing that something is a request to discontinue a previous request.

PR #450, opened 2022-12-02
FM2-476: Add support for searching MedicationRequests based on status

mogoodrich:
FM2-536: Expand Encounter "_has" Medication Request to support filtering by status

This PR implements FM2-476 and FM2-536, but *only* for the "active" status at this point.


api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMedicationRequestDaoImpl.java
```diff
@@ -91,6 +95,9 @@ protected void setupSearchParams(Criteria criteria, SearchParameterMap theParams
 					entry.getValue().forEach(d -> handleMedicationReference("d", (ReferenceAndListParam) d.getParam())
 					        .ifPresent(c -> criteria.createAlias("drug", "d").add(c)));
 					break;
+				case FhirConstants.STATUS_SEARCH_HANDLER:
```
mogoodrich, line 98: The "status_search_handler" is defined for the Task resource... I wasn't really sure if it made sense to repurpose it here, or if it makes more sense to create a new search handler... I'd need to understand better the concept behind a search handler.  Thoughts @ibacher ?

api/src/main/java/org/openmrs/module/fhir2/api/dao/FhirDaoUtils.java
```diff
@@ -0,0 +1,24 @@
...
+import org.hibernate.criterion.Restrictions;
+
+public final class FhirDaoUtils {
+	
+	public static Criterion createActiveOrderCriterion(String path) {
+		if (StringUtils.isNotBlank(path)) {
+			path = path + ".";
+		}
+		
+		// ACTIVE = date activated null or less than or equal to current datetime, date stopped null or in the future, auto expire date null or in the future
```
mogoodrich, line 16: I based this on how activated was defined in Order.java, but happy to debate whether that's the right approach (or if my adaptaion is wrong)

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseEncounterDao.java
```diff
@@ -111,6 +112,15 @@ protected void handleHasAndListParam(Criteria criteria, HasAndListParam hasAndLi
...
 								}
+							} else if (MedicationRequest.SP_STATUS.equals(paramName)) {
+								// only supports ACTIVE at this time
+								if (paramValue != null) {
+									if (MedicationRequest.MedicationRequestStatus.ACTIVE.toString()
+									        .equals(paramValue.toUpperCase())) {
+										criteria.add(FhirDaoUtils.createActiveOrderCriterion("orders"));
+									}
+								}
+								handled = true;
```
mogoodrich, line 123: This may be bad practice, but note that I set "handled" if the paramName is matched, regardless of whether or not the paramValue is matched...  this allows me to add a query param like "&_has:MedicationRequest:encounter:status=&"  and interpret that as "any status" without logging a warning... as I'm writing this, I'm guessing this is likely bad practice , thoughts @ibacher ?

api/src/main/java/org/openmrs/module/fhir2/api/dao/FhirDaoUtils.java
```diff
@@ -0,0 +1,24 @@
...
+		return Restrictions.and(
+		    Restrictions.or(Restrictions.isNull(path + "dateActivated"),
+		        Restrictions.le(path + "dateActivated", new Date())),
+		    Restrictions.or(Restrictions.isNull(path + "dateStopped"), Restrictions.gt(path + "dateStopped", new Date())),
+		    Restrictions.or(Restrictions.isNull(path + "autoExpireDate"),
+		        Restrictions.gt(path + "autoExpireDate", new Date())));
```
mseaton, line 22: I'd probably extract out "new Date()), and pass the date into the method.  That would allow you to both a) use the same date for all of the checks, and b) support both "is active now" and "is active at timetime xyz" with the same utility method in the future.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseEncounterDao.java
mseaton: Seems fine to me.

mseaton (approved):
Few comments/suggestions, but mostly defer to @ibacher

api/src/main/java/org/openmrs/module/fhir2/api/dao/FhirDaoUtils.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.util.Date;
+
+import org.apache.commons.lang3.StringUtils;
+import org.hibernate.criterion.Criterion;
+import org.hibernate.criterion.Restrictions;
+
+public final class FhirDaoUtils {
+	
+	public static Criterion createActiveOrderCriterion(String path) {
```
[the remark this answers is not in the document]
ibacher: I'm not opposed to a utility class, but generally speaking, we've put these kinds of helpers in `BaseDao` (or elsewhere in the DAO hierarchy where it's made sense, e.g., `BaseEncounterDao`).

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMedicationRequestDaoImpl.java
ibacher: I think it's fine to re-use it... If the string changes from `status`, presumably we'd use a different constant name.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMedicationRequestDaoImpl.java
```diff
@@ -99,6 +106,24 @@ protected void setupSearchParams(Criteria criteria, SearchParameterMap theParams
 		excludeDiscontinueOrders(criteria);
 	}
 	
+	private Optional<Criterion> handleStatus(TokenAndListParam tokenAndListParam) {
+		return handleAndListParam(tokenAndListParam, token -> {
+			if (token.getValue() != null) {
+				try {
+					// currently only handles "ACTIVE"
+					if (Medication.MedicationStatus.ACTIVE.toString().equals(token.getValue().toUpperCase())) {
+						return Optional.of(FhirDaoUtils.createActiveOrderCriterion(""));
```
ibacher, line 115: Personally, I think I'd prefer having a variation of this that takes no arguments, even if it just does this in the background.

ibacher (commented):
A couple of quick notes... and I agree with @mseaton's points. Otherwise, this LGTM.

api/src/main/java/org/openmrs/module/fhir2/api/dao/FhirDaoUtils.java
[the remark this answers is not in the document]
mogoodrich: I can move this to BaseDao (it's used both for Encounters and Medication Requests, so it doesn't really belong in either of those DAOs).

api/src/main/java/org/openmrs/module/fhir2/api/dao/FhirDaoUtils.java
mogoodrich: Sure,  I can go ahead and do this... I guess my main doubt is that although it's trivial to extract, it's not trivial to set up a bunch of test cases around it, so it will remain untested except for the "today" case for now.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirMedicationRequestDaoImpl.java
mogoodrich: Fair, will fix

mogoodrich (commented):
Reworked the utility method based on feedback @ibacher  @mseaton

ibacher (approved):

mogoodrich:
Awesome, thanks for quick review @ibacher  and @mseaton , will merge this in...

PR #454, opened 2023-01-26
FM2-551: Rework ValueSet "Compose" Field

mogoodrich:
## Issue I worked on

see https://issues.openmrs.org/browse/FM2-551

  No? -> execute above command


mogoodrich:
Some notes.  First, here's an example of the new format returned:

`{
   "resourceType":"ValueSet",
   "id":"378e63b1-6c75-46ed-95e3-797b48ddc9f8",
   "title":"Farm Animal Set",
   "status":"active",
   "date":"2005-01-01T00:00:00-05:00",
   "compose":{
      "include":[
         {
            "concept":[
               {
                  "code":"bbbf56f8-706e-41ef-95cd-57271567223a",
                  "display":"Cow"
               },
               {
                  "code":"b150225f-2086-43af-a13a-5ff16464a6f7",
                  "display":"Pig"
               },
               {
                  "code":"f5608410-6610-4314-84de-6d23a7fdfa99",
                  "display":"Sheep"
               }
            ]
         },
         {
            "system":"https://www.eaap.org/",
            "concept":[
               {
                  "code":"VACA",
                  "display":"Cow"
               },
               {
                  "code":"CERDO",
                  "display":"Pig"
               },
               {
                  "code":"OVEJA",
                  "display":"Sheep"
               }
            ]
         },
         {
            "system":"http://www.nfacc.ca/",
            "concept":[
               {
                  "code":"MOO",
                  "display":"Cow"
               },
               {
                  "code":"OINK",
                  "display":"Pig"
               },
               {
                  "code":"BAA",
                  "display":"Sheep"
               }
            ]
         }
      ]
   }
}`

Some questions/concerns/comments:

* There was some code that was designed to handle case when a member of a value set was value set itself (ie, sets within sets).  Concepts sets within concept sets is a not-uncommon use case in OpenMRS, but I was not sure how to handle it in the new model so I stripped it out for now.
* For all reference sources, I set the "display" property to the getDisplayString for the concept... not sure if there's something better or another pattern we use in other cases?
* If an (OpenMRS) concept source is not mapped to FHIR Concept Source (via the fhir_concept_source table) it's mappings will *not* show up in the result set.  This was the way the original code worked and I think the theory is that we need a source url in order to properly represent a source in FHIR?

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ValueSetTranslatorImpl.java
```diff
@@ -72,45 +70,52 @@ public ValueSet toFhirResource(@Nonnull Concept concept) {
...
+						ConceptReferenceTerm crt = conceptMapping.getConceptReferenceTerm();
 						String sourceUrl = conceptSourceCache.computeIfAbsent(crt.getConceptSource(),
 						    this::conceptSourceToURL);
-						if (sourceUrl == null) {
-							conceptReferenceComponent.setCode(conceptSetMember.getUuid());
-						} else {
-							conceptSetComponent.setSystem(sourceUrl);
+						// only add sources that we have urls for (provided by Fhir Concept Source table)
+						if (sourceUrl != null) {
+							// first see if we already have an icnlude for this source, and initialize one if not
```
mseaton, line 97: typo on "icnlude"

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ValueSetTranslatorImpl.java
```diff
@@ -72,45 +70,52 @@ public ValueSet toFhirResource(@Nonnull Concept concept) {
...
+							ValueSet.ConceptSetComponent conceptSetComponent = sets.stream().filter(set -> {
+								return sourceUrl.equals(set.getSystem());
+							}).findFirst().orElse(null);
```
mseaton, line 100: If we have to do above, I wonder if it would be more efficient to make this "sets" variable into a Map<String, ValueSet.ConcetSetComponent> rather than a List, and then at the bottom do "compose.setInclude(sets.values())"

mseaton (approved):
LGTM @mogoodrich .  I only glanced at the unit test code.  A few comments for your review that you can take or leave.

mogoodrich:
@ibacher  just a ping for you to take a look at this when you get a chance, thanks!

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ValueSetTranslatorImpl.java
```diff
@@ -72,45 +70,52 @@ public ValueSet toFhirResource(@Nonnull Concept concept) {
...
 			if (conceptSetMember != null) {
-				if (conceptSetMember.getSet()) {
-					ValueSet.ConceptSetComponent conceptSetComponent = new ValueSet.ConceptSetComponent();
-					conceptSetComponent
-					        .addValueSet(valueSetReferenceTranslator.toFhirResource(conceptSetMember).getReference());
-					compose.addInclude(conceptSetComponent);
-					continue;
-				}
+				// first, add the uuid and display to the concept uuid set
+				ValueSet.ConceptReferenceComponent uuidConceptReferenceComponent = new ValueSet.ConceptReferenceComponent();
```
ibacher, line 84: So a point that occurs to me here: in general, in FHIR, we wouldn't include a ValueSet as part of a ValueSet directly. FHIR distinguishes between [intensional and extensional ValueSets](https://www.hl7.org/fhir/valueset.html#int-ext), i.e., those value sets that are defined as "include the terms from Concept Source A and Value Set B" and those that are just an enumeration of all the concepts contained in the set. The implementation here was supposed to just represent extensional value sets (the ones where we enumerate all the concepts in the set).

Because FHIR ValueSets are not (in FHIR terms) concepts, the ValueSet itself would never appear in the extension (or "expansion"—FHIR's term for the operation that converts an intensional ValueSet into it's extensional representation). Thus, in the original implementation, we skip any set members that are themselves sets, instead just including their set members directly.

Obviously, in OpenMRS-world, things are different, i.e., I suppose it's possible for a concept that defines a concept set to also have some meaning other than just defining a set of additionally included concepts (i.e., we might use it to define a concept hierarchy).

I'm not really sure what the right thing to do here is, but I'm a little hesitant to directly include those concept sets here. Maybe what we could do is use the `ValueSet.compose.include.valueSet` parameter to reference the included concept set?

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ValueSetTranslatorImpl.java
```diff
@@ -72,45 +70,52 @@ public ValueSet toFhirResource(@Nonnull Concept concept) {
...
+							}).findFirst().orElse(null);
+							if (conceptSetComponent == null) {
+								conceptSetComponent = new ValueSet.ConceptSetComponent();
+								conceptSetComponent.setSystem(sourceUrl);
+								sets.add(conceptSetComponent);
+							}
+							// set the code and the display string
+							ValueSet.ConceptReferenceComponent conceptReferenceComponent = new ValueSet.ConceptReferenceComponent();
 							conceptReferenceComponent.setCode(crt.getCode());
+							conceptReferenceComponent.setDisplay(conceptSetMember.getDisplayString());
```
ibacher, line 109: For display values, we should use the `getMetadataTranslation()` utility I think, or a version of it. Basically, it does the same thing as the REST module's `getLocalization()` function for metadata, allowing us to use the customised values, if any.

ibacher (commented):
Thanks @mogoodrich! This basically LGTM, though I agree with Mike's comment about investigating using a Map.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ValueSetTranslatorImpl.java
mogoodrich: @ibacher ah, so maybe I misunderstood what the original code was supposed to be doing... are you saying that it's supposed to flatten it out... ie, if it finds a member that is in itself a set, retrieve all the set members and just add them to the top-level list (and then proceeding recursively if there are set members that are sets and in turn have sets?) while skipping the set concept?  I'm fine with that.

When I saw the getReference() call in the old code, it looks like it was building some sort of reference, and I didn't know how to handle that in my new model.  Maybe it was already doing what you suggested in your last paragraph?  If so, I can add some tests around it to confirm.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ValueSetTranslatorImpl.java
mogoodrich: Thanks @ibacher  @mseaton  makes sense.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ValueSetTranslatorImpl.java
mogoodrich: Oh, looking closer, I do think it was doing what you suggested in the last paragraph... let me see if I can restore that and add some more tests and update the PR

ibacher: Yes, I think that's what it was doing previously, but I don't think we had any tests to cover it.

api/src/main/java/org/openmrs/module/fhir2/api/util/FhirUtils.java
```diff
@@ -144,28 +145,46 @@ public static Optional<OpenmrsEncounterType> getOpenmrsEncounterType(Encounter e
...
+		String localization = getLocalization(className, uuid);
+		
+		if (localization == null) {
+			return metadata.getName();
+		} else {
+			return localization;
 		}
+	}
+	
+	public static String getMetadataTranslation(Concept concept) {
```
mogoodrich, line 161: Concept is not OpenmrsMetadata, and has a slightly different format, so I had to extract this out

mogoodrich:
@ibacher  @mseaton  updated the PR based on code review... new sample output here.... Ian, in particular, can you confirm that the format for the nested value set looks correct?

```
{
   "resourceType":"ValueSet",
   "id":"378e63b1-6c75-46ed-95e3-797b48ddc9f8",
   "title":"Farm Animal Set",
   "status":"active",
   "date":"2005-01-01T00:00:00-05:00",
   "compose":{
      "include":[
         {
            "concept":[
               {
                  "code":"bbbf56f8-706e-41ef-95cd-57271567223a",
                  "display":"Cow"
               },
               {
                  "code":"b150225f-2086-43af-a13a-5ff16464a6f7",
                  "display":"Pig"
               },
               {
                  "code":"f5608410-6610-4314-84de-6d23a7fdfa99",
                  "display":"Sheep"
               }
            ],
            "valueSet":[
               "ValueSet/a4e4d0be-8edb-4021-86a9-f7ed7d8d45e3"
            ]
         },
         {
            "system":"https://www.eaap.org/",
            "concept":[
               {
                  "code":"VACA",
                  "display":"Cow"
               },
               {
                  "code":"CERDO",
                  "display":"Pig"
               },
               {
                  "code":"OVEJA",
                  "display":"Sheep"
               }
            ]
         },
         {
            "system":"http://www.nfacc.ca/",
            "concept":[
               {
                  "code":"MOO",
                  "display":"Cow"
               },
               {
                  "code":"OINK",
                  "display":"Pig"
               },
               {
                  "code":"BAA",
                  "display":"Sheep"
               }
            ]
         }
      ]
   }
}
```

ibacher (approved):
Thanks @mogoodrich! This looks great!

mogoodrich:
Thanks @ibacher !  Looks like the tests are still running, will merge once that it done.

PR #457, opened 2023-02-07
FM2-555: Update Medication Dispensing to explicitly set null values a…

mogoodrich:
…s null

## Issue I worked on

  No? -> execute above command

### DIFF
--- a/api-2.6/src/main/java/org/openmrs/module/fhir2/api/translators/impl/MedicationDispenseTranslatorImpl_2_6.java
+++ b/api-2.6/src/main/java/org/openmrs/module/fhir2/api/translators/impl/MedicationDispenseTranslatorImpl_2_6.java
@@ -145,23 +145,37 @@ public class MedicationDispenseTranslatorImpl_2_6 implements MedicationDispenseT
 		openmrsObject.setPatient(patientReferenceTranslator.toOpenmrsType(fhirObject.getSubject()));
 		if (fhirObject.hasContext()) {
 			openmrsObject.setEncounter(encounterReferenceTranslator.toOpenmrsType(fhirObject.getContext()));
+		} else {
+			openmrsObject.setEncounter(null);
 		}
 		if (fhirObject.hasAuthorizingPrescription()) {
 			Reference prescription = fhirObject.getAuthorizingPrescriptionFirstRep();
 			openmrsObject.setDrugOrder(medicationRequestReferenceTranslator.toOpenmrsType(prescription));
+		} else {
+			openmrsObject.setDrugOrder(null);
 		}
 		if (fhirObject.hasStatus()) {
 			openmrsObject.setStatus(medicationDispenseStatusTranslator.toOpenmrsType(fhirObject.getStatus()));
-		}
+		} // status mandatory, so we don't reset it, even if null
+		
 		if (fhirObject.hasStatusReasonCodeableConcept()) {
 			openmrsObject.setStatusReason(conceptTranslator.toOpenmrsType(fhirObject.getStatusReasonCodeableConcept()));
+		} else {
+			openmrsObject.setStatusReason(null);
 		}
+		
 		if (fhirObject.hasLocation()) {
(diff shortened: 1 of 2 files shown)


api-2.6/src/main/java/org/openmrs/module/fhir2/api/translators/impl/MedicationDispenseTranslatorImpl_2_6.java
```diff
@@ -196,22 +210,38 @@ public MedicationDispense toOpenmrsType(@Nonnull MedicationDispense openmrsObjec
...
 		}
 		
-		if (fhirObject.getMeta() != null) {
-			openmrsObject.setDateCreated(fhirObject.getMeta().getLastUpdated());
+		if (fhirObject.getMeta() != null && fhirObject.getMeta().getLastUpdated() != null) {
+			if (openmrsObject.getDateCreated() == null) {
+				openmrsObject.setDateCreated(fhirObject.getMeta().getLastUpdated());
+			} else {
+				openmrsObject.setDateChanged(fhirObject.getMeta().getLastUpdated());
+			}
```
mogoodrich, line 244: This is tangential to the ticket, but seemed like the correct behavior... though, perhaps it begs the question, should we really be updating these OpenMRS "audit" fields at all?

ibacher (approved):
LGTM. Thanks @mogoodrich!

api-2.6/src/main/java/org/openmrs/module/fhir2/api/translators/impl/MedicationDispenseTranslatorImpl_2_6.java
mogoodrich: Merging this in, but still curious about thoughts as to whether we should up updated OpenMRS audit fields here... @ibacher  @mseaton

mogoodrich:
Thanks @ibacher , merging!

mseaton (commented):
Fine with me, since @ibacher has approved.  Would seem like we should consider having the translators just return null though, if they don't already, and then don't put the "has" check around them.  Then it would just set to null if not found, or set to the appropriate value if found...

mogoodrich:
that's a fair point, @mseaton ... and it may already do that, let me see.

api-2.6/src/main/java/org/openmrs/module/fhir2/api/translators/impl/MedicationDispenseTranslatorImpl_2_6.java
ibacher: Oooh... I missed this. I actually don't like the idea of manually updating the audit fields at all... IMO, we can use them to display values when it makes sense, but I don't love allowing anyone to actually edit them.

ibacher:
In general, the translators have been written so that if the supplied value is `null`, they return `null` already. That said, the `has...` check can be more complicated than just a null check, i.e., they usually also call a `<field>.isEmpty()` check which checks for what they regard as an empty element. I don't think it necessarily makes sense for, e.g., the ReferenceTranslator to need to know how to deal with an empty element (since it translates a Reference as a string).

Long story short, I actually think that having these `has...()` checks is the right thing to do because it puts the responsibility for setting the right value on the translator for a particular resource or data element (which should understand the resource or data element) rather than pushing the responsibility to a lower-level translator that may be used for multiple very similar but not identical fields.

api-2.6/src/main/java/org/openmrs/module/fhir2/api/translators/impl/MedicationDispenseTranslatorImpl_2_6.java
mogoodrich: thoughts @mseaton?

mseaton: Fine with me.

PR #460, opened 2023-02-24
FM2-559 : Task Resource should support multiple Task.input and Task.output value Types

mozzy11:
## Issue I worked on

see https://issues.openmrs.org/browse/FM2-559

  No? -> execute above command


api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImpl.java
```diff
@@ -173,52 +173,102 @@ private void setOpenmrsTaskFields(FhirTask openmrsTask, Task fhirTask) {
...
-		FhirReference outputReference = referenceTranslator.toOpenmrsType((Reference) taskOutputComponent.getValue());
-		Concept type = conceptTranslator.toOpenmrsType(taskOutputComponent.getType());
+	private FhirTaskOutput translateToOutput(Task.TaskOutputComponent taskOutputComponent) {
 		FhirTaskOutput output = new FhirTaskOutput();
-		
-		output.setValueReference(outputReference);
+		Concept type = conceptTranslator.toOpenmrsType(taskOutputComponent.getType());
 		output.setType(type);
 		
+		if (taskOutputComponent.getValue() instanceof Reference) {
```
pmanko, line 192: Are there any data types that might be in an input or output but we are not yet supporting? 

I think the possible types for input and output are quite large: https://build.fhir.org/datatypes.html#open, but obviously they don't all apply to the lab order workflows we are dealing with. I'm just wondering if we want to add any other support in this PR, or wait until there is a use case for it.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImpl.java
```diff
@@ -173,52 +173,102 @@ private void setOpenmrsTaskFields(FhirTask openmrsTask, Task fhirTask) {
...
-		Concept type = conceptTranslator.toOpenmrsType(parameterComponent.getType());
+	private FhirTaskInput translateToInput(Task.ParameterComponent parameterComponent) {
 		FhirTaskInput input = new FhirTaskInput();
-		
+		Concept type = conceptTranslator.toOpenmrsType(parameterComponent.getType());
 		input.setType(type);
-		input.setValueText(parameterComponent.getValue().toString());
+		
+		if (parameterComponent.getValue() instanceof Reference) {
+			FhirReference ref = referenceTranslator.toOpenmrsType((Reference) parameterComponent.getValue());
```
pmanko, line 217: Also, would it make sense to make a separate translator to deal with input and output values/codings and the business logic of it?

pmanko (approved):
Looks good at first glance, I just had a question about other possible data types to support. Thanks!

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImpl.java
mozzy11: Yeah , they are quite many .
we can always add more if need arises. i just wanted to support all [those supported ](https://github.com/openmrs/openmrs-module-fhir2/blob/master/api/src/main/java/org/openmrs/module/fhir2/model/FhirTaskParam.java#L42-L52)by the TaskParam Class

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImpl.java
mozzy11: yeah , moved the translation Logic to separate translators

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/TaskInputTranslatorImpl.java
```diff
@@ -0,0 +1,83 @@
...
+		}
+	}
+	
+	@Override
+	public FhirTaskInput toOpenmrsType(ParameterComponent parameterComponent) {
+		FhirTaskInput input = new FhirTaskInput();
+		Concept type = conceptTranslator.toOpenmrsType(parameterComponent.getType());
+		input.setType(type);
+		
+		if (parameterComponent.getValue() instanceof Reference) {
```
ibacher, line 66: A few remarks here:

1. It's probably better to use `Type value = parameterComponent.getValue();` and use that in the `if`s
2. Each of these `if`s (aside from the first) should be an `else if` because we only need one.
3. What happens when `value` is none of these types? There seems to be an inconsistency with what we do here and what we do in `toFhirResource()`. (Also, could we add a test for at least one non-supported value type?)

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r3/TaskFhirResourceIntegrationTest.java
```diff
@@ -169,6 +176,45 @@ public void shouldCreateNewTaskAsJson() throws Exception {
 		assertThat(newTask.getIntent(), equalTo(task.getIntent()));
 		assertThat(task.getAuthoredOn(), equalTo(task.getAuthoredOn()));
 		assertThat(task.getLastModified(), equalTo(task.getLastModified()));
+		assertThat(newTask.getOutput(), hasSize(4));
+		assertThat(newTask.getInput(), hasSize(4));
+		for (Task.TaskOutputComponent output : newTask.getOutput()) {
```
ibacher, line 181: Why write the verifications like this? The order is semi-important, right, or at least we should be able to predict what it is here, so we can just do:

```java
assertThat(newTask.getOutput().get(0), /* ... */)
```

Which is much more straight-forward to understand than these loops.

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r3/TaskFhirResourceIntegrationTest.java
```diff
@@ -169,6 +176,45 @@ public void shouldCreateNewTaskAsJson() throws Exception {
...
+			if (output.getValue() instanceof Reference) {
+				assertThat(((Reference) output.getValue()).getReference(),
+				    equalTo("DiagnosticReport/9b6f11dd-55d2-4ff6-8ec2-73f6ad1b759e"));
+				
+			} else if (output.getValue() instanceof DateTimeType) {
+				DateFormat sdf = new SimpleDateFormat("yyyy-MM-dd");
```
ibacher, line 187: There are some examples of how we can test dates. Please look at them... I don't love converting them to strings, etc., because this can make the time-stamps locale specific.

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r3/TaskFhirResourceIntegrationTest.java
```diff
@@ -169,6 +176,45 @@ public void shouldCreateNewTaskAsJson() throws Exception {
...
+				assertThat(((DecimalType) input.getValue()).getValueAsNumber().doubleValue(), equalTo(37.38));
+				
+			} else if (input.getValue() instanceof StringType) {
+				assertThat(input.getValue().toString(), equalTo("Test code"));
+				
+			}
+		}
 	}
 	
 	@Test
```
ibacher, line 220: We should also test that the API is able to create these from the JSON and XML successfully (and it would be good to use the XML format as well as the JSON format).

ibacher (changes requested):
Thanks @mozzy11!! This is basically all good, with a few small tweaks!

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r3/TaskFhirResourceIntegrationTest.java
mozzy11: done

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r3/TaskFhirResourceIntegrationTest.java
mozzy11: makes sense

ibacher (approved):

PR #472, opened 2023-04-28
FM2-571: Support "PATCH" operation

mogoodrich:
## Issue I worked on

  No? -> execute above command


mogoodrich:
Moving this this out of the WIP state, ready for review.

api/pom.xml
```diff
@@ -56,10 +56,13 @@
...
 			<artifactId>guava</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>com.github.java-json-tools</groupId>
+			<artifactId>json-patch</artifactId>
+		</dependency>
 		<dependency>
 			<groupId>com.fasterxml.jackson.core</groupId>
 			<artifactId>jackson-databind</artifactId>
-			<scope>test</scope>
```
mogoodrich, line 62: So seeing this now, it looks like *I* was the one that moved this out of test scope, because I'm using Jackson in the JsonPatchUtils file... does this seem okay @ibacher ?

mogoodrich: (Although this looks like it already was included in the omod pom.xml as a normal dependency, so I think it would have been packaged with the omod irregardless (and necessary for the 2.1 and below Core versions, like you said?)

api/src/test/java/org/openmrs/module/fhir2/providers/r4/MedicationRequestFhirResourceProviderTest.java
```diff
@@ -317,59 +312,6 @@ public void searchMedicationRequest_shouldNotAddRelatedMedicationsForEmptyInclud
 		assertThat(resources.get(0).getIdElement().getIdPart(), equalTo(MEDICATION_REQUEST_UUID));
 	}
 	
-	@Test
-	public void createMedicationRequest_shouldCreateMedicationRequest() {
```
ibacher, line 321: Is there a reason we're removing these tests?

mogoodrich: I removed the create method, which wasn't working, until we have a chance to fix it, see: https://github.com/openmrs/openmrs-module-fhir2/pull/472/files#diff-7c16ac594c8eeabfe06059fce2d86b382b900c02754483dc17c9f5387536170bL72

I guess we could just comment out these tests and pull them back in when we implement, but I don't find these tests to be particularly useful.  @ibacher

ibacher:
@mogoodrich So I added a commit that basically works around the limitation by temporarily "pretending" that the Merge Patch is a JSON patch. Once the code is actually executed, the Content-Type header should be correct, so maybe we can use that to differentiate between "proper" JSON patches and JSON merge patches, but for now, this at least accepts the merge-patch+json content type.

api/src/test/java/org/openmrs/module/fhir2/providers/r4/MedicationRequestFhirResourceProviderTest.java
ibacher: No that's fine. I didn't realise they weren't working, but now that you mention it, of course that's the case. I much prefer deleting to commenting out.

mogoodrich:
Awesome, thanks @ibacher !  I will test it out with your commit and merge if everything looks good, thanks!

api/pom.xml
ibacher: Yeah... I think it's perfectly fine not to have this `test` scope. This is probably a result of the time I tried to cut down the size of the OMOD as much as possible (embedding the HAPI code and—even worse—the FHIR Jars makes this module relatively large).

api/src/test/java/org/openmrs/module/fhir2/providers/r4/MedicationRequestFhirResourceProviderTest.java
ibacher: Yeah... At one point I wanted to support MedicationRequests, but then I saw all the stuff Mike went through to implement it in HTMLFormEntry and decided we could wait until there was a real need.

mogoodrich:
This works great, thanks @ibacher , I am going to merge.

PR #473, opened 2023-05-02
FM2-562: Fix pagination to address slow running queries

rkorytkowski:
## Issue I worked on

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
@@ -9,17 +9,13 @@
  */
 package org.openmrs.module.fhir2.api.dao.impl;
 
-import static org.hibernate.criterion.Projections.property;
 import static org.hibernate.criterion.Restrictions.eq;
-import static org.openmrs.module.fhir2.api.util.LastnOperationUtils.getTopNRankedIds;
 
 import javax.annotation.Nonnull;
 
 import java.util.ArrayList;
 import java.util.Date;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
@@ -35,6 +31,7 @@ import ca.uhn.fhir.rest.param.TokenParam;
 import org.apache.commons.lang3.StringUtils;
 import org.hibernate.Criteria;
 import org.hibernate.criterion.Criterion;
+import org.hibernate.criterion.Order;
 import org.hibernate.criterion.Projections;
 import org.hibernate.criterion.Subqueries;
 import org.hl7.fhir.r4.model.Observation;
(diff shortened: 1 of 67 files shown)


ibacher (approved):
Thanks @rkorytkowski! This looks great! My only real comment is that, for w/e reason, the OpenMRS coding standard is to never use `*` imports.

rkorytkowski:
This is an initial pass. There's still one failing tests that I need to fix. I have trouble understanding the logic behind the method that is being tested so I'd appreciate some explanation.

The fix is to use DB pagination instead of paginating on the java side. Alongside I changed the code to run only 1 query and not two queries to get results in most cases as there's no need to first get IDs and then actual results. I also removed distinct filtering on results in places where it is not needed and made the distinct work on the DB side. The count also uses proper select count(*) and not size() on the results.

rkorytkowski:
> Thanks @rkorytkowski! This looks great! My only real comment is that, for w/e reason, the OpenMRS coding standard is to never use `*` imports.

I'll fix that. Apparently my IDE is not configured properly.

rkorytkowski:
@ibacher Thanks! Fixed the imports and the failing test. There's one thing with the new implementation for lastn obs feature that the count of results may be off in a rare condition. I commented in code. It feels such rare that I would not sacrifice performance to support it. I'm also not sure though if we expose that count anywhere...

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
```diff
@@ -59,26 +55,92 @@ public class FhirObservationDaoImpl extends BaseFhirDao<Obs> implements FhirObse
...
-			@SuppressWarnings("unchecked")
-			List<LastnResult<Integer>> results = ((List<Object[]>) criteria.list()).stream().map(objects -> {
-				Map<String, Object> attributes = new HashMap<>();
-				attributes.put("concept", objects[1]);
-				return new LastnResult<>((Integer) objects[0], (Date) objects[2], attributes);
-			}).collect(Collectors.toList());
+			List<Obs> results = new ArrayList<>();
+			int firstResult = 0;
+			int maxGroupCount = getMaxParameter(theParams);
+			final int BATCH_SIZE = 3;
```
ibacher, line 69: I'm not sure I quite understand why we're using the constant 3 here?

ibacher:
@rkorytkowski The `count` does end up exposed as the `total` field in the search results. That said, I don't think that having that exactly right is super critical...

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirObservationDaoImpl.java
rkorytkowski: I was testing the algorithm... I forgot to revert it back to 100. Good eye ;)

rkorytkowski:
> @rkorytkowski The `count` does end up exposed as the `total` field in the search results. That said, I don't think that having that exactly right is super critical...

Actually, I misunderstood the algorithm a bit, which is why the integration tests were failing and it may happen more commonly. I made the adjustments to the count so it's always accurate.

I also noticed that getting counts was called twice during a request so decided to cache it for the lifespan of the request.

ibacher:
> I also noticed that getting counts was called twice during a request so decided to cache it for the lifespan of the request.

Ooof... I hadn't that before. Thanks that should be good. I'm happy for you to merge this in whenever you're comfortable doing so, though it would be nice to hear back from the OHRI team whether this helps.

PR #494, opened 2023-06-21
FM2-572: Improve performance for total count in results

rkorytkowski:
## Issue I worked on

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java
@@ -346,4 +346,8 @@ public class FhirConstants {
 	public static final String RX_NORM_SYSTEM_URI = "http://www.nlm.nih.gov/research/umls/rxnorm";
 	
 	public static final String UCUM_SYSTEM_URI = "http://unitsofmeasure.org";
+	
+	public static final String EXACT_TOTAL_SEARCH_PARAMETER = "_exactTotal";
+	
+	public static final String COUNT_QUERY_CACHE = "countQueryCache";
 }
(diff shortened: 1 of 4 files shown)


rkorytkowski:
I'll update openmrs-core to enable query cache by default for this feature to work. 

I can't update JIRA for now as there seems to be login issue... I'll update as soon as it starts working again for me.

ibacher (approved):

ibacher:
We still need to surface this to the actual API via the `providers`, but this looks like a good implementation of this. I assume things don't break when using older verisons of core, the caching is just not enabled? (We support back to core 2.0.5 at present).

rkorytkowski:
> We still need to surface this to the actual API via the `providers`

Could you please explain?

rkorytkowski:
> I assume things don't break when using older verisons of core, the caching is just not enabled? (We support back to core 2.0.5 at present).

It's completely ignored if caching is not enabled globally via hibernate properties.

ibacher:
Yeah, so the actual API for this is provided by the HAPI library using their `IResourceProvider` interface. But to make the parameter actually something user's can toggle with the API, the argument needs to be part of the method marked with the `@Search` parameter in those classes, like [this one](https://github.com/openmrs/openmrs-module-fhir2/blob/356a3b36c2b4b10a9366e7b23fba33659f2b161b/api/src/main/java/org/openmrs/module/fhir2/providers/r4/ObservationFhirResourceProvider.java#L93). These providers are all in the `org.openmrs.module.fhir2.providers` package in the various `api` subprojects, i.e., the bulk of them are [here](https://github.com/openmrs/openmrs-module-fhir2/tree/master/api/src/main/java/org/openmrs/module/fhir2/providers).

Doesn't need to be part of this PR, just future work we'll need to do to add, e.g.,

`@OptionalParameter(name = EXACT_TOTAL_SEARCH_PARAMETER) exactTotalSearch`

To the search methods.

PR #542, opened 2024-06-25
(fix) Set encounter and default role for encounter provider entity

icrc-psousa:

This PR resolves the SQLIntegrityConstraintViolationException when creating an encounter via the FHIR API with a participant. The root cause was the EncounterProvider entity being flushed to the database before the Encounter was created, resulting in a null encounter_id. This update ensures the EncounterProvider entity is properly linked to the Encounter before being persisted.
## Issue I worked on

see https://issues.openmrs.org/browse/FM2-639

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImpl.java
@@ -149,7 +149,7 @@ public class EncounterTranslatorImpl extends BaseEncounterTranslator implements 
 		return existingEncounter;
 	}
 
-	public EncounterRole getDefaultEncounterRole() {
+	private EncounterRole getDefaultEncounterRole() {
 		EncounterRole role = encounterService.getEncounterRoleByName("Unknown");
 		if (role == null) {
 			throw new IllegalStateException("Missing encounter role named 'Unknown'");
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImpl.java
@@ -28,7 +28,7 @@ import org.openmrs.EncounterProvider;
 import org.openmrs.EncounterRole;
 import org.openmrs.EncounterType;
 import org.openmrs.Visit;
-import org.openmrs.api.context.Context;
+import org.openmrs.api.EncounterService;
 import org.openmrs.module.fhir2.FhirConstants;
 import org.openmrs.module.fhir2.api.translators.EncounterLocationTranslator;
 import org.openmrs.module.fhir2.api.translators.EncounterParticipantTranslator;
@@ -61,6 +61,9 @@ public class EncounterTranslatorImpl extends BaseEncounterTranslator implements 
 	
 	@Autowired
 	private EncounterPeriodTranslator<org.openmrs.Encounter> encounterPeriodTranslator;
+
+	@Autowired
+	private EncounterService encounterService;
 	
 	@Override
 	public Encounter toFhirResource(@Nonnull org.openmrs.Encounter openmrsEncounter) {
@@ -129,7 +132,6 @@ public class EncounterTranslatorImpl extends BaseEncounterTranslator implements 
 		existingProviders.addAll(encounter
 		        .getParticipant().stream().map(encounterParticipantComponent -> participantTranslator
 		                .toOpenmrsType(new EncounterProvider(), encounterParticipantComponent))
-				.filter(Objects::nonNull)
 		        .collect(Collectors.toCollection(LinkedHashSet::new)));
 
 		for (EncounterProvider ep : existingEncounter.getEncounterProviders()) {
@@ -148,7 +150,7 @@ public class EncounterTranslatorImpl extends BaseEncounterTranslator implements 
 	}
 
 	public EncounterRole getDefaultEncounterRole() {
-		EncounterRole role = Context.getEncounterService().getEncounterRoleByName("Unknown");
+		EncounterRole role = encounterService.getEncounterRoleByName("Unknown");
 		if (role == null) {
 			throw new IllegalStateException("Missing encounter role named 'Unknown'");
 		}
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImpl.java
@@ -17,6 +17,7 @@ import javax.annotation.Nonnull;
 
 import java.util.Collections;
 import java.util.LinkedHashSet;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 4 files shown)


dkayiwa:
@icrc-psousa did you see the above build failure?

icrc-psousa:
> @icrc-psousa did you see the above build failure?



> @icrc-psousa did you see the above build failure?

Issue was now addressed. Thanks for the heads-up @dkayiwa 
I've also added some more info and linked an issue. I'm not sure who can give more insights on this. Feel free to add reviewers.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImpl.java
```diff
@@ -127,6 +133,11 @@ public org.openmrs.Encounter toOpenmrsType(@Nonnull org.openmrs.Encounter existi
 		        .getParticipant().stream().map(encounterParticipantComponent -> participantTranslator
 		                .toOpenmrsType(new EncounterProvider(), encounterParticipantComponent))
 		        .collect(Collectors.toCollection(LinkedHashSet::new)));
+
+		for (EncounterProvider ep : existingEncounter.getEncounterProviders()) {
+			ep.setEncounter(existingEncounter);
+			ep.setEncounterRole(getDefaultEncounterRole());
+		}
```
ibacher, line 140: Instead of doing this, could we include an additional `.map()` step in the above that does the same?

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImpl.java
```diff
@@ -137,4 +148,12 @@ public org.openmrs.Encounter toOpenmrsType(@Nonnull org.openmrs.Encounter existi
...
 		return existingEncounter;
 	}
+
+	private EncounterRole getDefaultEncounterRole() {
+		EncounterRole role = encounterService.getEncounterRoleByName("Unknown");
+		if (role == null) {
+			throw new IllegalStateException("Missing encounter role named 'Unknown'");
+		}
+		return role;
+	}
```
ibacher, line 158: I don't love hard-coding things like this. At the very least, this needs to be `protected` rather than `private` so it's overridable. Also, I would prefer this to be at least a GP or something the implementation can control in some form or fashion. Finally, ideally, setting the encounter role would be done in the participantTranslator and not here.

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImplTest.java
```diff
@@ -250,11 +247,33 @@ public void toOpenMrsType_shouldTranslateParticipantToEncounterProvider() {
...
+		when(encounterService.getEncounterRoleByName("Unknown")).thenReturn(new EncounterRole());
+		when(participantTranslator.toOpenmrsType(ArgumentMatchers.any(), ArgumentMatchers.any())).thenReturn(new EncounterProvider());
 		org.openmrs.Encounter result = encounterTranslator.toOpenmrsType(fhirEncounter);
 		
 		assertThat(result.getEncounterProviders(), not(empty()));
 		assertThat(result.getEncounterProviders().size(), equalTo(1));
 	}
+
+	@Test(expected = IllegalStateException.class)
+	public void toFhirResource_shouldThrowExceptionWhenUknownRoleIsNull() {
```
ibacher, line 259:
```suggestion
	public void toFhirResource_shouldThrowExceptionWhenUnknownRoleIsNull() {
```

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImplTest.java
```diff
@@ -250,11 +247,33 @@ public void toOpenMrsType_shouldTranslateParticipantToEncounterProvider() {
...
+		fhirEncounter.setParticipant(participantComponents);
+		EncounterProvider encounterProvider = new EncounterProvider();
+		Provider provider = new Provider();
+		provider.setUuid(PRACTITIONER_UUID);
+		encounterProvider.setProvider(provider);
+		Patient patient = new Patient();
+		patient.setUuid(PATIENT_UUID);
+		when(patientReferenceTranslator.toOpenmrsType(patientRef)).thenReturn(patient);
+		when(participantTranslator.toOpenmrsType(ArgumentMatchers.any(), ArgumentMatchers.any())).thenReturn(new EncounterProvider());
+		encounterTranslator.toOpenmrsType(fhirEncounter);
```
ibacher, line 275:
```suggestion

		encounterTranslator.toOpenmrsType(fhirEncounter);
```

ibacher: Prefer using `assertThrows()` with a lambda over `expected` in the test annotation.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImpl.java
```diff
@@ -17,15 +17,18 @@
 
 import java.util.Collections;
 import java.util.LinkedHashSet;
+import java.util.Objects;
```
ibacher, line 20: Seems unnecessary?

ibacher (changes requested):
Thanks @icrc-psousa! A few suggestions on how to better integrate this with the code base here.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/EncounterTranslatorImpl.java
icrc-psousa: Thanks for the feedback Ian, I've now moved this to participant translator.
Would it make sense trying to fetch role from GP, something like this:

```
	protected EncounterRole getDefaultEncounterRole() {
		String defaultEncounterRoleName = globalPropertyService.getGlobalProperty("fhir2.defaultEncounterRoleName");

		String roleName = (defaultEncounterRole != null && !defaultEncounterRole.isEmpty())
				? defaultEncounterRoleName
				: "Unknown";

		EncounterRole role = encounterService.getEncounterRoleByName(roleName);

		if (role == null) {
			throw new IllegalStateException("Missing encounter role named 'Unknown'");
		}

		return role;
	}
```
Without setting the encounter role an ex would occur when saving the encounter:
SQLIntegrityConstraintViolationException: Column 'encounter_role_id' cannot be null

ibacher: I understand that we need to fill in the field, it's just the module should be as agnostic about the system's metadata as possible. That's exactly the sort of solution I was imagining.

icrc-psousa: Modified PR to fetch role from GP if present [here](https://github.com/openmrs/openmrs-module-fhir2/pull/542/files#diff-63c16fdd2f4cbcc348687f85367b9cb611f008db42e516063744c7617c69c334R70).
@ibacher please have a look and let me know if you have any change suggestion. Thanks!

ibacher (approved):

PR #546, opened 2024-07-23
FM2-650: Replace the existing name instead of creating a new name

icrc-jofrancisco:
### Problem
When updating a patient's name using the FHIR endpoint, a new person_name entry is created in the database instead of voiding the existing one and creating a new entry. This results in multiple active names for a single patient, which is incorrect and leads to data inconsistencies. Additionally, OpenMRS does not consider more than one name and always picks the first one, so users cannot see any changes in the **UI.**

### Solution
This fix addresses the problem by:

- Replacing the current name with the new entry, matching their ids (which is how the rest API currently works).
If the person's name doesn't contain an ` id`for some reason (in a normal case it will), it will “ replace ” the preferred name for the person.

By implementing this approach, we ensure the UI reflects the latest changes accurately.
Thnaks,
## Issue I worked on

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/PatientTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/PatientTranslatorImpl.java
@@ -148,13 +148,12 @@ public class PatientTranslatorImpl implements PatientTranslator {
 		}
 		
 		for (HumanName name : patient.getName()) {
-			PersonName existingName;
+			PersonName existingName = null;
 			if (name.hasId()) {
 				existingName = currentPatient.getNames().stream().filter(n -> n.getUuid().equals(name.getId())).findFirst()
 				        .orElse(null);
-			} else {
-				existingName = currentPatient.getPersonName();
 			}
+			
 			PersonName pn = nameTranslator.toOpenmrsType(existingName != null ? existingName : new PersonName(), name);
 			currentPatient.addName(pn);
 		}
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/PatientTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/PatientTranslatorImpl.java
@@ -15,7 +15,6 @@ import static org.openmrs.module.fhir2.api.translators.impl.FhirTranslatorUtils.
 
 import javax.annotation.Nonnull;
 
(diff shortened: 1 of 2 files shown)


ibacher:
Functionally, it's incorrect to assume a patient should only have one name. We can, however, try voiding any names that are not in the list. (The UI defaults to the preferred name for a client, but that doesn't mean additional names are not important).

icrc-jofrancisco:
> Functionally, it's incorrect to assume a patient should only have one name. We can, however, try voiding any names that are not in the list. (The UI defaults to the preferred name for a client, but that doesn't mean additional names are not important).

Hey @ibacher , thanks for the feddback.

You're right. I'm trying to implement a logic to update the name instead of setting the others to voided.
But right now I can't figure out which name should be updated, because we only get a list of all names...

I'm still investigating another possible solution, which is why I put the PR in draft. 
Anyway, thanks for the feedback!

delphinepas:
@denniskigen @ibacher would it be possible for you to review this PR ? Thank you cc @icrc-jofrancisco @icrc-fdeniger @gracepotma

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/PatientTranslatorImpl.java
```diff
@@ -148,7 +148,15 @@ public org.openmrs.Patient toOpenmrsType(@Nonnull org.openmrs.Patient currentPat
 		}
 		
 		for (HumanName name : patient.getName()) {
-			currentPatient.addName(nameTranslator.toOpenmrsType(name));
+			PersonName existingName;
+			if (name.hasId()) {
+				existingName = currentPatient.getNames().stream().filter(n -> n.getUuid().equals(name.getId())).findFirst()
+				        .orElse(null);
+			} else {
+				existingName = currentPatient.getPersonName();
```
ibacher, line 156: This behaviour still strikes me as incorrectly assuming that a patient has only a single name...

ibacher (changes requested):

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/PatientTranslatorImpl.java
ibacher: I'm on board with matching existing names based on UUIDs, but I'm not sold that this is a good idea.

ibacher (approved):

PR #554, opened 2024-12-11
FM2-651: Allow DAOs access to all API load-able by the OpenMRS Class loader.

Ruhanga:
## Issue I worked on

see https://openmrs.atlassian.net/browse/FM2-651

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseFhirDao.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseFhirDao.java
@@ -55,6 +55,7 @@ import org.openmrs.module.fhir2.FhirConstants;
 import org.openmrs.module.fhir2.api.dao.FhirDao;
 import org.openmrs.module.fhir2.api.search.param.PropParam;
 import org.openmrs.module.fhir2.api.search.param.SearchParameterMap;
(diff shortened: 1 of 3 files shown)


ibacher:
So, a couple of things here:

1. For FHIR module PRs we usually want to have tests so we don't regress on the issue
2. Doing this in the DAO layer feels off to me; it would be preferable to do it somewhere centrally, maybe in the servlet itself, e.g., overwriting the `service()` method to something like `Thread.currentThread().setContextClassLoader(OpenmrsClassLoader.getInstance()); super()`.
3. Have you tried this and does this actually fix the issue? The reason I ask is that the `getSearchResultsCount()` method is called before `getSearchResults()` and I'd expect we'd need to the same fix for `createOrUpdate()` and `delete()` as well.

Ruhanga:
> Have you tried this and does this actually fix the issue?

Yes.

> Doing this in the DAO layer feels off to me; it would be preferable to do it somewhere centrally, maybe in the servlet itself, e.g., overwriting the service() method to something like Thread.currentThread().setContextClassLoader(OpenmrsClassLoader.getInstance()); super().

I initially wanted to ensure that the `ContextClassLoader` was set correctly before the DAO layer was called. However, after reconsidering, I realized this wasn’t strictly necessary. As a result, I’ve implemented the suggested approach, which works just as well.

PR #556, opened 2025-01-09
FM2-653

mogoodrich:
## Issue I worked on

https://openmrs.atlassian.net/browse/FM2-653

  No? -> execute above command


api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseDao.java
```diff
@@ -558,8 +558,11 @@ protected Optional<Criterion> handleGender(@Nonnull String propertyName, TokenAn
 		});
 	}
 	
-	protected Optional<Criterion> handleLocationReference(@Nonnull String locationAlias,
-	        ReferenceAndListParam locationReference) {
+	protected Optional<Criterion> handleLocationReference(ReferenceAndListParam locationReference) {
```
mogoodrich, line 561: I needed to support creating a location reference criterion without an alias, see below...

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirLocationDaoImpl.java
```diff
@@ -38,6 +41,8 @@
 @Setter(AccessLevel.PACKAGE)
 public class FhirLocationDaoImpl extends BaseFhirDao<Location> implements FhirLocationDao {
 	
+	private static final int SUPPORTED_LOCATION_HIERARCHY_SEARCH_DEPTH = 9;
```
mogoodrich, line 44: Basically, I use a succession of joins to go back up the ancestor tree.  A bit annoying/hacky, but from a practical matter, seems unlikely there would ever be a hierarchy of more than 9 levels.  We could make this a configurable parameter if we wanted.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirLocationDaoImpl.java
```diff
@@ -84,6 +93,25 @@ public List<LocationAttribute> getActiveAttributesByLocationAndAttributeTypeUuid
 		        .list();
 	}
 	
+	private void handleAncestorLocation(Criteria criteria, ReferenceAndListParam ancestor) {
+		
+		List<Criterion> elementOrAncestorEqualsReferenceCriteria = new ArrayList<>();
+		Optional<Criterion> elementEqualsReferenceCriterion = handleLocationReference(ancestor); // note: "below" is inclusive of the element itself
```
mogoodrich, line 99: see: https://hl7.org/fhir/R4/search.html#recursive

api/src/main/java/org/openmrs/module/fhir2/providers/r3/LocationFhirResourceProvider.java
```diff
@@ -113,6 +113,9 @@ public IBundleProvider searchLocations(@OptionalParam(name = Location.SP_NAME) S
 	        @OptionalParam(name = Location.SP_PARTOF, chainWhitelist = { "", Location.SP_NAME, Location.SP_ADDRESS_CITY,
 	                Location.SP_ADDRESS_STATE, Location.SP_ADDRESS_COUNTRY,
 	                Location.SP_ADDRESS_POSTALCODE }, targetTypes = Location.class) ReferenceAndListParam parent,
+	        @OptionalParam(name = "below", chainWhitelist = { "", Location.SP_NAME, Location.SP_ADDRESS_CITY,
```
mogoodrich, line 116: So it sounds like @ibacher  this extra parameter should not be here, and this new functionality should just be part of the "SP_PARTOF" parameter above?  I was having a problem with that, see my follow on comment and screenshot in the test case below.

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r3/LocationFhirResourceProviderIntegrationTest.java
```diff
@@ -483,6 +486,31 @@ public void shouldSearchForExistingLocationsAsJson() throws Exception {
 		assertThat(entries, containsInRelativeOrder(hasResource(hasProperty("name", equalTo("Test location 6"))),
 		    hasResource(hasProperty("name", equalTo("Test location 8")))));
 		assertThat(entries, everyItem(hasResource(validResource())));
+		
+		// search by ancestors
+		response = get("/Location?below=" + LOCATION_ANCESTOR_TEST_UUID).accept(FhirMediaTypes.JSON).go();
```
mogoodrich, line 491: So when I changed this to "Location?partof:below=" it was getting picked up by the ResourceParam, but the "below" was being set to the ResourceType.  This seems incorrect?  Isn't the type of this resource a location?

image

api/src/main/java/org/openmrs/module/fhir2/providers/r3/LocationFhirResourceProvider.java
mogoodrich: The tl;dr is that if I change this to name="partof:below" any parameter passed in that way gets assigned to *both* the ReferenceAndListParam ancestor and ReferenceAndListParam parent.  

And if I remove the "below" parameter, I'm not quite sure how to parse the "parent" ReferenceAndListParam to tell if this should be a standard "partof" vs a "partof:below"

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirLocationDaoImpl.java
```diff
@@ -112,8 +119,56 @@ private void handleTag(Criteria criteria, TokenAndListParam tags) {
 		}
 	}
 	
-	private void handleParentLocation(Criteria criteria, ReferenceAndListParam parent) {
-		handleLocationReference("loc", parent).ifPresent(loc -> criteria.createAlias("parentLocation", "loc").add(loc));
+	private void handleLocationReference(Criteria criteria, ReferenceAndListParam locationAndReferences) {
+		
+		if (locationAndReferences == null) {
```
mogoodrich, line 124: All this through 148 is basically a series null / size checks to make sure that ReferenceAnd list only contains a single reference.  We probably should change this parameter to just we a

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirLocationDaoImpl.java
```diff
@@ -112,8 +119,56 @@ private void handleTag(Criteria criteria, TokenAndListParam tags) {
...
+		ReferenceParam locationReference = locationReferences.get(0);
+		
+		// **NOTE: this is a *bug* in the current HAPI FHIR implementation, "below" should be the "queryParameterQualifier", not the resource type; likely need update this when/fix the HAPI FHIR implementation is fixed**
+		// this is to support queries of the type "Location?partof=below:uuid"
+		if ("below".equalsIgnoreCase(locationReference.getResourceType())) {
+			List<Criterion> belowReferenceCriteria = new ArrayList<>();
+			
+			// we need to add a join to the parentLocation for each level of hierarchy we want to search, and add "equals" criterion for each level
+			int depth = 1;
+			while (depth <= SUPPORTED_LOCATION_HIERARCHY_SEARCH_DEPTH) {
```
ibacher, line 157: Should this be a for loop?

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirLocationDaoImpl.java
ibacher: Maybe we turn this into a GP? That way an implementation can adjust it to the actual depth of their location hierarchy.

ibacher (approved):
Thanks @mogoodrich! I think this basically looks as good as we can get it!

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirLocationDaoImpl.java
mogoodrich: Fair enough, I will do that (and since it's configurable, set a more normal default of like 5)

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirLocationDaoImpl.java
mogoodrich: You mean aesthetically, or is there a bug here I'm missing?  :)  For some reason I tend to prefer while loops.

ibacher: Just aesthetics. It looks like it works.

mogoodrich:
Merging, thanks @ibacher !

PR #558, opened 2025-03-02
FM2-126  Add support for mapping OpenMRS attribute types to FHIR

parthfloyd:
# FM2-456: Add Person Attribute Extensions Support to Patient Translator

Added support for handling person attributes as FHIR extensions in the Patient resource. Person attributes in OpenMRS can now be properly translated to and from FHIR extensions (Currently supporting String, Boolean, Concept & Location).

The implementation:
1. Adds a dependency on `PersonAttributeTranslatorImpl` for translation
2. Creates `getPersonAttributeExtensions` method to convert attributes to extensions
3. Updates `toFhirResource` to add extensions to the Patient resource
4. Updates `toOpenmrsType` to convert extensions back to OpenMRS person attributes

The implementation expects person attribute extensions to follow this nested structure:
```json
{
  "extension": [
    {
      "url": "http://fhir.openmrs.org/person-attribute",
      "extension": [
        {
          "url": "http://fhir.openmrs.org/person-attribute#Birthplace",
          "valueString": "Lisbon"
        }
      ]
    }
  ]
}
```

Where:
- Outer extension with URL `http://fhir.openmrs.org/person-attribute` serves as a container
- Inner extensions represent individual attributes with type-specific values

## Issue I worked on

see https://openmrs.atlassian.net/browse/FM2-126

- [x] My IDE is configured to follow the [**[code style](https://wiki.openmrs.org/display/docs/Java+Conventions)**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

cc: @icrc-jofrancisco , @icrc-fdeniger ,  @ibacher


api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/PersonAttributeTranslatorImpl.java
```diff
@@ -0,0 +1,198 @@
...
+		
+		Extension wrapperExtension = createWrapperExtension();
+		Extension valueExtension = createValueExtension(personAttribute);
+		wrapperExtension.setExtension(Collections.singletonList(valueExtension));
+		
+		return valueExtension;
+	}
+	
+	@Override
+	public PersonAttribute toOpenmrsType(@Nonnull Extension extension) {
```
ibacher, line 75: We presumably need to be able to handle updating attributes as well? This seems to only handle adding them.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/PersonAttributeTranslatorImpl.java
```diff
@@ -0,0 +1,198 @@
...
+				case JAVA_BOOLEAN_FORMAT:
+					extension.setValue(new BooleanType(value));
+					break;
+				case OPENMRS_LOCATION_FORMAT:
+					handleLocationFormat(extension, value);
+					break;
+				case OPENMRS_CONCEPT_FORMAT:
+					handleConceptFormat(extension, value);
+					break;
+				default:
```
ibacher, line 122: Default behaviour should be to fail silently to process things we don't understand, not throwing exceptions.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/PersonAttributeTranslatorImpl.java
```diff
@@ -0,0 +1,198 @@
...
+					throw new UnsupportedOperationException("Unsupported attribute format: " + format);
+			}
+		}
+		catch (Exception e) {
+			throw new IllegalStateException("Error processing attribute value for format: " + format, e);
+		}
+	}
+	
+	private void handleLocationFormat(Extension extension, String locationId) {
```
ibacher, line 131: This seems to rely on exposing the primary key identifier for the location. These values must _always_ be converted to and from UUIDs when sent out.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/PersonAttributeTranslatorImpl.java
```diff
@@ -0,0 +1,198 @@
...
+			return null;
+		}
+		
+		PersonAttribute personAttribute = new PersonAttribute();
+		Extension valueExtension = extension.getExtensionFirstRep();
+		Type extensionValue = valueExtension.getValue();
+		String attributeTypeName = extractAttributeTypeName(valueExtension.getUrl());
+		
+		personAttribute.setAttributeType(getPersonAttributeType(attributeTypeName));
+		setPersonAttributeValue(personAttribute, extensionValue);
```
ibacher, line 86: I don't think this will work correctly, especially for things like references and concepts. At the very least, it requires that the import format is different from what we're outputting, which isn't the way this should work at all.

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/PersonAttributeTranslatorImplTest.java
```diff
@@ -0,0 +1,283 @@
...
+import org.openmrs.PersonAttribute;
+import org.openmrs.PersonAttributeType;
+import org.openmrs.api.ConceptService;
+import org.openmrs.api.LocationService;
+import org.openmrs.api.PersonService;
+import org.openmrs.module.fhir2.FhirConstants;
+import org.openmrs.module.fhir2.api.translators.ConceptTranslator;
+
+@RunWith(MockitoJUnitRunner.class)
+public class PersonAttributeTranslatorImplTest {
```
ibacher, line 43: In addition to translator tests, we should also have at least one integration test for this behaviour.

parthfloyd:
> I don't really understand why we need the outer extension. What value does that add?

@ibacher I realise now it is confusing & i messed up purpose of the wrapper function. Overall I want the extension to provide both the `personAttribute.value` & `personAttribute.type`

For this can we make the extension like:
(For a person attribute with  value 'Lisbon' and PersonAttributeType name: Birthplace,  

## Option 1 (without wrapper extension):
```
{
  "extension": [
    {
      "url": "http://fhir.openmrs.org/person-attribute#Birthplace",
      "valueString": "Lisbon"
    }
  ]
}

```
## Option 2:
```
{
  "extension": [
    {
      "url": "http://fhir.openmrs.org/person-attribute",
      "valueString": "Lisbon"
      "extension": [
        {
          "url": "http://fhir.openmrs.org/person-attribute-type",
          "valueString": "Birthplace"
        }
      ]
    }
  ]
}
```

## Option 3:
```
{
  "extension": [
    {
      "url": "http://fhir.openmrs.org/person-attribute",
      "valueString": "Lisbon"
      "extension": [
        {
          "url": "http://fhir.openmrs.org/person-attribute-type#Birthplace"
        }
      ]
    }
  ]
}
```

ibacher:
Option 1 is most consistent with how we've mapped things out in the past. I don't really see a huge value in complicating things more than that.

That said, for constructing the system, please follow the patterns we already use, which you can find in the [constants](https://github.com/openmrs/openmrs-module-fhir2/blob/master/api/src/main/java/org/openmrs/module/fhir2/FhirConstants.java). Basically, I think this should be something like: "http://fhir.openmrs.org/ext/person-attribute#Birthplace"

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/PersonAttributeTranslatorImpl.java
parthfloyd: Hi @ibacher Can you elaborate on this? 

For clarification for Person Attribute with Location type as value the extension will be:
```
{
  "extension": [
    {
      "url": "http://fhir.openmrs.org/person-attribute#location",
      "valueReference": {
           "reference" : "Location/<Location_UUID>"
       }
    }
  ]
}
```

Similarly for CodeableConcept:
```
{
  "extension": [
    {
      "url": "http://fhir.openmrs.org/person-attribute#some coding",
      "valueCodeableConcept": {
          "coding": [
            {
              "system": "http://snomed.info/sct",
              "code": "123456007",
              "display": "Hypertension"
            }
          ]
      }
    }
  ]
}
```

Meanwhile for the PersonAttribute table it only saves the `value` as either Location.id or CodingConcept's id.

ibacher: Yeah, I think I misread the code here.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/PersonAttributeTranslatorImpl.java
```diff
@@ -0,0 +1,198 @@
...
+	
+	private void setPersonAttributeValue(PersonAttribute personAttribute, Type extensionValue) {
+		try {
+			if (extensionValue instanceof BooleanType) {
+				personAttribute.setValue(((BooleanType) extensionValue).getValueAsString());
+			} else if (extensionValue instanceof StringType) {
+				personAttribute.setValue(((StringType) extensionValue).getValue());
+			} else if (extensionValue instanceof Reference) {
+				String reference = ((Reference) extensionValue).getReference();
+				personAttribute.setValue(reference.substring(reference.lastIndexOf("/") + 1));
```
ibacher: Use `FhirUtils#referenceToId()` or `ReferenceHandlingTranslator#getReferenceId()`. There are quite a few edge cases where this won't work.

parthfloyd:
Thanks @ibacher I've incorporated the changes along with the integration test.
Please feel free to let me know if any changes are required!

ibacher:
@parthfloyd Look at the test failures. It turns out this code is generating invalid URLs, since person attribute names can contain spaces, but URL hash fragments can't. If you can fix that, we're probably ok to go.

ibacher:
Yeah, that approach makes sense to me. Not ideal, but this will be correct. We'll have to add some error handling for people submitting attribute extensions without a type.

ibacher (approved):

ibacher:
@parthfloyd So, unfortunately, I missed this, but there's a [rule](https://hl7.org/fhir/R4/extensibility-definitions.html) on extensions such that they "Must have either extensions or value[x], not both". These means the little scheme here won't work, so, I guess we'll need something like:

```json
{
   "url": "http://fhir.openmrs.org/person-attribute",
   "extension": [{
     "url": "http://fhir.openmrs.org/person-attribute-type",
     "value": "Birthplace"
   }, {
     "url": "http://fhir.openmrs.org/person-attribute-value",
     "value": "Lisbon"
   }]
}
```

I'm going to try a few ideas to come up with something better, but it may be substantially later in the week before I've got anything ready.

parthfloyd:
> Must have either extensions or value[x], not both

Thanks Ian for looking into this. The Option you proposed seems appropriate & robust!

parthfloyd:
@ibacher After brainstorming on it, even I found your suggestion most optimal.
I've currently opened a PR with the suggested format. Let me know if any changes are required here: #564

PR #559, opened 2025-03-18
O3-4544: Improve FHIR Location endpoint performance

pwargulak:

This is a performance improvement / a PoC of possibilities of performance improvements around FHIR API. This change implements a simple idea of "less db queries, faster application".
The `ToFhirTranslator` has been extended with a new method - `toFhirResources(Collection)` - which intention is to handle multiple OpenMRS data elements together. All existing implementations of `ToFhirTranslator` use default implementation of `toFhirResources(Collection)` that simple delegates to existing `toFhirResource(Data)`, with an exception of `LocationTranslator`. 

The general logic of getting FHIR Location **before** the change:
1. Read OpenMRS Locations according to specified query.
2. For each OpenMRS Location
    1. Read Location Contact Details.
    3. Read other properties.

The general logic of getting FHIR Location **after** the change:
1. Read OpenMRS Locations according to specified query.
2. Read Contact Details referenced by Locations from 1st point. 
3. For each OpenMRS Location
    1. Look-up Location Contact Details from preloaded collection.
    2. Read other properties.

The `Contact Details` has been selected as the property that loading had the most time cost - **67% of time spend on creation of FHIR Location was spent on Contact Details**.

The change provides **an overall speedup of 70-85%**.

See https://talk.openmrs.org/t/how-to-help-with-o3-performance/45157

## Issue I worked on
see https://openmrs.atlassian.net/browse/O3-4544

### DIFF
--- a/api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirLocationDaoImplTest.java
+++ b/api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirLocationDaoImplTest.java
@@ -9,12 +9,14 @@
  */
 package org.openmrs.module.fhir2.api.dao.impl;
 
+import static java.util.Collections.singleton;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.Matchers.nullValue;
 
 import java.util.List;
+import java.util.Map;
 
 import org.junit.Before;
 import org.junit.Test;
@@ -77,6 +79,17 @@ public class FhirLocationDaoImplTest extends BaseFhirContextSensitiveTest {
 		    LOCATION_ATTRIBUTE_TYPE_UUID);
 		
 		assertThat(attributeList, notNullValue());
+	}
+	
+	@Test
+	public void getActiveAttributesByLocationsAndAttributeTypeUuid_shouldReturnLocationAttribute() {
+		Location location = new Location();
+		location.setUuid(LOCATION_UUID);
+		
+		Map<Location, List<LocationAttribute>> attributes = fhirLocationDao
+		        .getActiveAttributesByLocationsAndAttributeTypeUuid(singleton(location), LOCATION_ATTRIBUTE_TYPE_UUID);
+		
+		assertThat(attributes, notNullValue());
 	}
 	
 	@Test


api/src/main/java/org/openmrs/module/fhir2/api/translators/LocationTranslator.java
```diff
@@ -24,6 +27,15 @@ public interface LocationTranslator extends OpenmrsFhirUpdatableTranslator<org.o
...
 	Location toFhirResource(@Nonnull org.openmrs.Location openmrsLocation);
 	
+	/**
+	 * Maps a collection of {@link org.openmrs.Location}s to a {@link org.hl7.fhir.r4.model.Location}
+	 *
+	 * @param openmrsLocations the collection of locations to translate
+	 * @return the mapping of OpenMRS location to corresponding FHIR location resource
+	 */
+	@Override
+	Map<org.openmrs.Location, Location> toFhirResources(Collection<org.openmrs.Location> openmrsLocations);
```
ibacher, line 37: I'm not sure about this return type. Is there some reason this can't just return a `List<Location>` or `Collection<Location>` or similar? While there's clear value to a map inside the implementation, I have a harder time grasping the use-case for this map.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/LocationTranslatorImpl.java
```diff
@@ -218,4 +165,83 @@ public org.openmrs.Location getOpenmrsParentLocation(Reference location) {
...
+		        .getGlobalProperty(FhirConstants.LOCATION_CONTACT_POINT_ATTRIBUTE_TYPE);
+		
+		if (locationContactPointAttributeType == null || locationContactPointAttributeType.isEmpty()) {
+			return locations.stream().collect(Collectors.toMap(Function.identity(), ignored -> Collections.emptyList()));
+		}
+		
+		final Map<org.openmrs.Location, List<LocationAttribute>> contactPointAttributes = fhirLocationDao
+		        .getActiveAttributesByLocationsAndAttributeTypeUuid(locations, locationContactPointAttributeType);
+		
+		return contactPointAttributes.entrySet().stream().collect(Collectors.toMap(Map.Entry::getKey,
```
ibacher, line 181: Not that this shouldn't work, but it seems like we should have a test to cover this method, as we currently don't.

ibacher (changes requested):
Thanks @pwargulak! I like the overall idea here. I think we need to cover the actual transformation to ContactPoints in a test and change the return type from the interface, but otherwise, I think this looks great!

api/src/main/java/org/openmrs/module/fhir2/api/translators/LocationTranslator.java
pwargulak: I've started with a List - modified the `org.openmrs.module.fhir2.api.translators.ToFhirTranslator` interface.

But it wouldn't work with `org.openmrs.module.fhir2.api.translators.LocationTypeTranslator#toFhirResource` that returns a `List<CodeableConcept>`, the `CodeableConcept` have to back-reference to Location, so the caller of `toFhirResources(Collection<Location>)` had no way to group them.

I was considering wrapping `List<CodeableConcept>` into some object with Location/Location.id, but I went with "don't overthink, just make it work" solution.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/LocationTranslatorImpl.java
pwargulak: Do you mean `FhirLocationDao#getActiveAttributesByLocationsAndAttributeTypeUuid(...)`?
Or `LocationTranslatorImpl#getLocationContactDetails(...)`?

I've added small unit tests, the codecov tool seems to be satisfied, or you mean the unit test could be better? :)

pwargulak:
@ibacher Responded in the review threads.

ibacher (approved):

PR #566, opened 2025-05-30
O3-4765: Improved FHIR Get Lab Results endpoint performance

druchniewicz:
This is a performance improvement for Get Lab Results endpoint which is mainly related to fetching observations/concepts. Main idea of this solution is to reduce number of database queries which gives us a significant increase in performance. 

The general logic of fetching Observations before the changes looked like:

1. Get results (obses) from database.
2. For each obs convert to FHIR obs, which consists of:

- convert Obs to FHIR Observation, which consists of: 

- for each concept mapping search for concept source url -> separate db query performed

First introduced improvement was fetchin fhir concept sources once before loop, and inside loop just read these values. This increased the efficiency significantly, around 50%. But still fetching all concept sources was performed for each obs so idea is to use existing service method oFhirResources(Collection data) - and in Obs implementation of this method collect all mappings between Concept and CodeableConcept. For this, new method has been created Map<T, U> toFhirResourcesMap(Collection<T> data).

To sum up, main things that have been changed are loading mappings between Concept and ConceptCodeable and FhirSources once and read these values from preloaded collections - we read from memory instead of database each time. 

The above mentioned changes allowed to obtain an **increase in efficiency of around 70%.**

See https://talk.openmrs.org/t/how-to-help-with-o3-performance/45157
More detailed results can be found [here](https://docs.google.com/document/d/1b5Zr_yp1CX63UfKlwy4dDasry9FXUYHpMkovC17zuC0/edit?usp=drive_link)
and also in attached file in [ticket](https://openmrs.atlassian.net/browse/O3-4765)

## Issue I worked on
See https://openmrs.atlassian.net/browse/O3-4765

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/ConceptTranslator.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/ConceptTranslator.java
@@ -11,6 +11,9 @@ package org.openmrs.module.fhir2.api.translators;
 
 import javax.annotation.Nonnull;
 
+import java.util.Collection;
+import java.util.List;
+
 import org.hl7.fhir.r4.model.CodeableConcept;
 import org.openmrs.Concept;
 
@@ -24,6 +27,15 @@ public interface ConceptTranslator extends OpenmrsFhirTranslator<Concept, Codeab
 	 */
 	@Override
 	CodeableConcept toFhirResource(@Nonnull Concept concept);
+	
+	/**
+	 * Maps a collection of {@link org.openmrs.Concept}s to a
+	 * {@link org.hl7.fhir.r4.model.CodeableConcept}
+	 *
+	 * @param openmrsConcepts the collection of OpenMRS concepts to translate
+	 * @return the mapping of OpenMRS concept to corresponding FHIR concept resource
+	 */
+	List<CodeableConcept> toFhirResources(Collection<Concept> openmrsConcepts);
 	
 	/**
 	 * Maps a FHIR resource to an OpenMRS data element
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/OpenmrsFhirTranslator.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/OpenmrsFhirTranslator.java
@@ -9,8 +9,12 @@
  */
 package org.openmrs.module.fhir2.api.translators;
 
+import static java.util.stream.Collectors.toMap;
+
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 /**
@@ -29,5 +33,15 @@ public interface OpenmrsFhirTranslator<T, U> extends ToFhirTranslator<T, U>, ToO
 	 */
 	default List<U> toFhirResources(Collection<T> data) {
 		return data.stream().distinct().map(this::toFhirResource).collect(Collectors.toList());
+	}
+	
+	/**
+	 * Maps OpenMRS data elements to FHIR resource map.
+	 *
+	 * @param data the collection of OpenMRS data elements to translate
+	 * @return the mapping of OpenMRS data element to corresponding FHIR resource map
+	 */
+	default Map<T, U> toFhirResourcesMap(Collection<T> data) {
+		return data.stream().distinct().collect(toMap(Function.identity(), this::toFhirResource));
 	}
 }
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConceptTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConceptTranslatorImpl.java
@@ -9,15 +9,21 @@
  */
 package org.openmrs.module.fhir2.api.translators.impl;
 
+import static java.util.stream.Collectors.toMap;
 import static lombok.AccessLevel.PROTECTED;
 
 import javax.annotation.Nonnull;
 
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
+import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
@@ -28,9 +34,12 @@ import org.openmrs.ConceptMap;
 import org.openmrs.ConceptMapType;
 import org.openmrs.ConceptReferenceTerm;
 import org.openmrs.ConceptSource;
+import org.openmrs.Duration;
+import org.openmrs.module.fhir2.FhirConstants;
 import org.openmrs.module.fhir2.api.FhirConceptService;
 import org.openmrs.module.fhir2.api.FhirConceptSourceService;
 import org.openmrs.module.fhir2.api.translators.ConceptTranslator;
+import org.openmrs.module.fhir2.model.FhirConceptSource;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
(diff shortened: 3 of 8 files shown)


api/src/main/java/org/openmrs/module/fhir2/api/translators/ConceptTranslator.java
```diff
@@ -25,6 +28,15 @@ public interface ConceptTranslator extends OpenmrsFhirTranslator<Concept, Codeab
...
 	CodeableConcept toFhirResource(@Nonnull Concept concept);
 	
+	/**
+	 * Maps a collection of {@link org.openmrs.Concept}s to a
+	 * {@link org.hl7.fhir.r4.model.CodeableConcept}
+	 *
+	 * @param openmrsConcepts the collection of OpenMRS concepts to translate
+	 * @return the mapping of OpenMRS concept to corresponding FHIR concept resource
+	 */
+	List<CodeableConcept> toFhirResources(Collection<Concept> openmrsConcepts);
```
dkayiwa, line 38: Can we add a since annotation to this? https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#New-Public-Methods-and-Classes

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConceptTranslatorImpl.java
```diff
@@ -48,45 +57,22 @@ public class ConceptTranslatorImpl implements ConceptTranslator {
 	
 	@Override
 	public CodeableConcept toFhirResource(@Nonnull Concept concept) {
-		if (concept == null) {
```
dkayiwa, line 51: What happens when concept == null?

api/src/main/java/org/openmrs/module/fhir2/api/translators/ConceptTranslator.java
druchniewicz: I added this annotation to newly added interface methods

api/src/main/java/org/openmrs/module/fhir2/api/translators/OpenmrsFhirTranslator.java
```diff
@@ -30,4 +34,15 @@ public interface OpenmrsFhirTranslator<T, U> extends ToFhirTranslator<T, U>, ToO
...
 	}
+	
+	/**
+	 * Maps OpenMRS data elements to FHIR resource map.
+	 *
+	 * @param data the collection of OpenMRS data elements to translate
+	 * @return the mapping of OpenMRS data element to corresponding FHIR resource map
+	 * @since 2.6.0
+	 */
+	default Map<T, U> toFhirResourcesMap(Collection<T> data) {
```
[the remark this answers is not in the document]
pwargulak: I might chip in here.
It doesn't need to be in the OpenmrsFhirTranslator, doesn't need to be generic, but this change and the location endpoint change rely on the same idea of translating collections of OMRS Objects as a whole. 
This method suggests a solution for building a cache of translated OMRS Objects that can be used when translating other OMRS Objects. In this case, we have FHIRConcept by Concept used when translating Obs. 

So, the generic use is here only as a suggestion for future changes.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConceptTranslatorImpl.java
```diff
@@ -126,6 +111,53 @@ public Concept toOpenmrsType(@Nonnull CodeableConcept concept) {
 		return null;
 	}
 	
+	private CodeableConcept toFhirConcept(Concept concept, ConceptTranslatorContext context) {
```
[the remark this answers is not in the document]
rkorytkowski: Is this context created for a single http request? I think it would make more sense to cache this in a service with @Cacheable annotation as concepts rarely change and we could use it across many http requests... It's even higher performance gain and simpler implementation.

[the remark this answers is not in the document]
druchniewicz: yes, those contexts like `ConceptTranslatorContext` and `ObservationTranslatorContext` have been created to populate data collections once and then use it in further actions (this did a trick and improves performance significantly). I can try your suggestion with Cacheable annotation. I can create an any new interface or just use some existing, create method like `getConceptTranslatorContext()`, or getObservationTranslatorContext(), annotate it with Cacheable and call where needed. First call of these methods should cache data, second and each subsequent calls should read data from cached ones.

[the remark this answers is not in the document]
pwargulak: @druchniewicz Have you measured the difference? And how the cache invalidation was handled (it doesn't change often, but may sometimes change)?

api/src/main/java/org/openmrs/module/fhir2/api/translators/ConceptTranslator.java
```diff
@@ -33,4 +59,13 @@ public interface ConceptTranslator extends OpenmrsFhirTranslator<Concept, Codeab
...
 	@Override
 	Concept toOpenmrsType(@Nonnull CodeableConcept concept);
+	
+	/**
+	 * Creates {@link ConceptTranslatorContext}. Used for cache data to improve performance in
+	 * subsequent request calls.
+	 *
+	 * @return concept translator context
+	 */
+	@Cacheable
```
rkorytkowski, line 69: Have you checked that it's actually cached? I think the annotation needs to go on the implementation not the interface to be picked up.

Also please see https://openmrs.atlassian.net/wiki/spaces/Archives/pages/25504850/Caching
It's very likely you need to add cache config files for ehcache and infinispan.

Please configure it to auto-expire (configured for inifinispan "entity" by default), because you cannot control evict if ConceptSource changes. We'll add a way in core to enable that for modules in the future.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConceptTranslatorImpl.java
```diff
@@ -48,45 +57,22 @@ public class ConceptTranslatorImpl implements ConceptTranslator {
...
-					}
-				}
-			}
-		}
+		return openmrsConcepts.stream().map(concept -> toFhirResource(concept, context)).collect(Collectors.toList());
+	}
+	
+	@Override
+	public Map<Concept, CodeableConcept> toFhirResourcesMap(Collection<Concept> openmrsConcepts) {
+		ConceptTranslatorContext context = getConceptTranslatorContext();
```
rkorytkowski, line 72: I think you need to call it via self.getConceptTranslatorContext() and add 
```
@Setter(value = PROTECTED, onMethod_ = @Autowired)
private ConceptTranslator self;
```
For the method to be called via proxy. It's explained at https://www.baeldung.com/spring-invoke-cacheable-other-method-same-bean

rkorytkowski: Please see my comment above about not introducing this toFhirResourcesMap at all.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ObservationTranslatorImpl.java
```diff
@@ -153,40 +208,14 @@ public Observation toFhirResource(@Nonnull Obs observation) {
...
-	@Override
-	public Obs toOpenmrsType(Obs existingObs, Observation observation, Supplier<Obs> groupedObsFactory) {
-		notNull(existingObs, "The existing Obs object should not be null");
-		notNull(observation, "The Observation object should not be null");
-		
-		observationStatusTranslator.toOpenmrsType(existingObs, observation.getStatus());
-		
-		existingObs.setEncounter(encounterReferenceTranslator.toOpenmrsType(observation.getEncounter()));
-		existingObs.setPerson(patientReferenceTranslator.toOpenmrsType(observation.getSubject()));
+	private class ObservationTranslatorContext {
```
rkorytkowski, line 211: It might not be necessary if you just annotate 
```
@Cacheable
CodeableConcept toFhirResource(@Nonnull Concept concept) {}
```
Please see my comment above.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/ConceptTranslatorImpl.java
```diff
@@ -164,4 +206,21 @@ private void addConceptsToMap(Map<String, Concept> mapTypeToConceptMap, List<Con
 			}
 		});
 	}
+	
+	private String getSourceUrl(ConceptSource conceptSource, ConceptTranslatorContext context) {
```
rkorytkowski, line 210: It's a good candidate for `@Cacheable` as well.

api/src/main/java/org/openmrs/module/fhir2/api/translators/ConceptTranslator.java
druchniewicz: @rkorytkowski to sum up caching configuration in fhir2 module, do you have any example in another module? (different than omrs-core)?
Looking into your suggestions:
1. Method implementations that I want to cache should be annotated with @ Cacheable
2. I can define custom cache names e.g. @ Cacheable (value = "conceptToCodeableConcept")
3. I see in openmrs-core and documention as well that local cache type is used so...
4. I need to define `infinispan-api-local.xml` in fhir2 module, right? Is this [file](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/resources/infinispan-api-local.xml) defined in omrs-core is used by default in all modules? What exactly should be set in expiration property? If auto-expire is set by default then I'm guessing it can be skipped, but what does exactly auto-expires mean? After how long will the cache data disappear?
5. Or instead point 4 I should create infinispan-api-local.xml but with <distributed-cache> entries as in this [example](https://infinispan.org/docs/13.0.x/titles/configuring/configuring.html#multiple_caches) ?
6. Apart from that should I also create cache-api.xml where I define my cacheNames: configuration: "entity"?
7. Should I also extend ehcache configuration by extending apiCacheConfig.properties ?

@rkorytkowski could you explain what steps in what order should be introduced to apply cache mechanism in fhir2 module in correct way? What is required to implement in fhir2 module/what is inherited from omrs-core module? It would probably save me a lot of time to understand, apply and test changes. Thanks!

rkorytkowski: I'm not aware of any modules using Cacheable. It's just in core. I guess it wasn't advertised enough... Modules used their own caches which is bad practice given we have little control over them and want to support both local and distributed caching.

Added steps in https://openmrs.atlassian.net/wiki/spaces/Archives/pages/25504850/Caching#Applying-Cacheable-in-modules

Re 3. Since openmrs-core 2.8.x we have local and distributed cache using infinispan. Pre openmrs-core 2.8.x we have only local cache using ehcache.

Hope everything else is addressed in the wiki. Please let me know if you have any further questions.

rkorytkowski: And thanks for your questions! I'd like to continue improving the documentation to make it easier for module devs to use it.

rkorytkowski:
Your [commit](https://github.com/SolDevelo/openmrs-module-fhir2/commit/8a0c895abce5e34d50187eaba5ba39a091ee5373) looks great to me. Much cleaner solution.

Re 5. Are you seeing the warnings throughout running the test or just at the beginning. If it's just at the beginning, I'd say that it's expected, because cache updates may not be visible between threads right away. I'm assuming you are running openmrs 2.7.x with ehcache. I'd check without the key. Maybe it's not resolved properly. You can also try monitoring with jconsole https://www.ehcache.org/generated/2.10.0/html/ehc-all/#page/Ehcache_Documentation_Set%2Fco-jmx_jconsole_example.html%23 Just temporarily autowire CacheManager in e.g. conceptTranslatorImpl and register as in the example and fire up jconsole.

druchniewicz:
thanks @rkorytkowski I'm off this week so I will take a look on it next week

rkorytkowski:
@druchniewicz thanks! The screenshot shows just **hibernateCache**. Do you see **apiCache** with **fhir2GetFhirConceptSources** and ** fhir2ConceptToCodeableConcept** caches as well?

rkorytkowski:
Try injecting instead:
```
@Autowired @Qualifier("apiCacheManager")
private CacheManager cacheManager;
```

rkorytkowski:
It's 50% with Cacheable not working due to not using it in an openmrs service, right? Now that we have it in an openmrs service it should be faster.
The issues are:
https://openmrs.atlassian.net/browse/TRUNK-6360
https://openmrs.atlassian.net/browse/TRUNK-6363

rkorytkowski:
Addressed in https://github.com/openmrs/openmrs-module-fhir2/pull/569

PR #567, opened 2025-06-18
FM2-126: Fix handling of concepts in PersonAttributes

icrc-jofrancisco:

Updated concept retrieval to use the correct `ConceptService.get(Integer id)` method by explicitly parsing the string ID to an integer.
Previously, the code was passing a stringified numeric ID to `ConceptService.get(String uuid)`, which led to `null` being returned in real scenarios because no UUID matched the numeric string. This change ensures we call the correct method overload for integer-based IDs.

Thanks, 
CC: @ibacher @parthfloyd 

### DIFF
--- a/integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r3/PatientFhirResourceProviderIntegrationTest.java
+++ b/integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r3/PatientFhirResourceProviderIntegrationTest.java
@@ -197,7 +197,7 @@ public class PatientFhirResourceProviderIntegrationTest extends BaseFhirR3Integr
 		        .filter(ext -> ext.getUrl().equals(FhirConstants.OPENMRS_FHIR_EXT_PERSON_ATTRIBUTE))
 		        .collect(Collectors.toList());
 		
-		assertThat(personAttributeExtensions, hasSize(3));
+		assertThat(personAttributeExtensions, hasSize(4));
 	}
 	
 	@Test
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/PersonAttributeTranslatorImpl.java
```diff
@@ -200,16 +200,14 @@ protected Reference buildLocationReference(String locationId) {
 	}
 	
 	protected CodeableConcept buildCodeableConcept(String conceptId) {
-		if (conceptId == null) {
-			return null;
```
ibacher, line 204: I'd prefer if we still had the null check here, I think.

icrc-jofrancisco: Thanks for the feedback @ibacher . I added the validation and kept the layout similar to what we have above for the location type.

gracepotma:
@pmanko since Ian is extremely swamped right now, could you kindly take a look at this PR? 

Big-picture context here: ICRC is using Google Health's Android FHIR SDK to build a secure offline app for offline/community-based patient visits. To get it to exchange data with their O3 systems, they've been making improvements to the FHIR2 module, like this one. :) So this is based in field-level needs.

ibacher:
@icrc-jofrancisco Could we add an integration test to cover this feature please?

ibacher:
@icrc-jofrancisco Thanks for this and I apologize it's taken this long to review it. One last thing: do you know the ticket of the initial implementation here? I'd like to tag this commit to it, just we have a record of why we're making these changes.

PR #577, opened 2025-08-24
FHIR R4: normalize CodeableConcept codings and HL7 escapes; add tests

mmdavenport:
## What I changed
- Added `FhirNormalizationUtils` to trim `CodeableConcept` codings and decode HL7 v2 escapes (\F\ → |, \S\ → ^, \T\ → &, \R\ → ~, \E\ → \).
- Added unit test `FhirNormalizationUtilsTest`.

## Why
- Prevents stray whitespace and HL7 escape sequences from leaking into FHIR R4 resources.

## Tests
- `FhirNormalizationUtilsTest` passes locally.

## Notes
- Java 17 used.

### DIFF
--- a/.mvn/jvm.config
+++ b/.mvn/jvm.config
@@ -1,0 +1,4 @@
+--add-opens=java.base/java.util=ALL-UNNAMED
+--add-opens=java.base/java.lang=ALL-UNNAMED
+--add-opens=java.base/java.lang.reflect=ALL-UNNAMED
+--add-opens=java.base/java.io=ALL-UNNAMED
--- a/.mvn/jvm.config
+++ b/.mvn/jvm.config
@@ -1,4 +1,0 @@
---add-opens=java.base/java.util=ALL-UNNAMED
---add-opens=java.base/java.lang=ALL-UNNAMED
---add-opens=java.base/java.lang.reflect=ALL-UNNAMED
---add-opens=java.base/java.io=ALL-UNNAMED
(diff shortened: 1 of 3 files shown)


.mvn/jvm.config
```diff
@@ -0,0 +1,4 @@
+--add-opens=java.base/java.util=ALL-UNNAMED
+--add-opens=java.base/java.lang=ALL-UNNAMED
```
sherrif10, line 2: Do we really need  java configurations. ?

mmdavenport: Removing it. That was from my local build setup and isn't needed for this change. Thanks for catching it.

api/src/main/java/org/openmrs/module/fhir2/api/util/FhirNormalizationUtils.java
```diff
@@ -0,0 +1,55 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
```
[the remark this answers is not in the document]
mmdavenport: Fixed in the latest commit - removed the duplicate header block so only the standard MPL header remains. Thanks for catching it.

PR #605, opened 2026-05-15
FM2-680: Always set priorPrescription when previousOrder is set

wikumChamith:
## Description of what I changed

  - Collapse the action-discriminated branch in MedicationRequestTranslatorImpl to a single previousOrder != null check that always sets
  priorPrescription.
  - Drop the basedOn branch previously used for RENEW.
  - Fixes O3 renewals (placed as action = NEW with previous_order_id set) returning null priorPrescription, which broke continuity-of-care and refill
   analytics.
  - priorPrescription is the FHIR R4 field for prescription chains; basedOn is for plan/proposal fulfillment (wrong semantics).
  - Update RENEW test to assert priorPrescription; add new test for action = NEW with previousOrder set.

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-680

## Checklist: I completed these to help reviewers :)

- [x] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

  No? Unsure? -> [configure your IDE](https://wiki.openmrs.org/display/docs/How-To+Setup+And+Use+Your+IDE), format the code and add the changes with `git add . && git commit --amend`

- [ ] I have **added tests** to cover my changes. (If you refactored
  existing code that was well tested you do not have to add tests)

  No? -> write tests and add them to this commit `git add . && git commit --amend`

- [x] I ran `mvn clean package` right before creating this pull request and
  added all formatting changes to my commit.

  No? -> execute above command

- [x] All new and existing **tests passed**.

  No? -> figure out why and add the fix to your commit. It is your responsibility to make sure your code works.

- [x] My pull request is **based on the latest changes** of the master branch.

  No? Unsure? -> execute command `git pull --rebase upstream master`

PR #606, opened 2026-05-26
FM2-681:  Migrate fhir2 to OpenMRS platform 3.0

wikumChamith:
- Switch javax.servlet → jakarta.servlet across api / omod / integration-tests; drop the explicit servlet-api dep (jakarta now comes via
  openmrs-core).
- Bump HAPI FHIR 5.7.9 → 7.6.0; inline the Thymeleaf narrative bridge that HAPI 7 split out, collapse OpenmrsNarrativeTemplateManifest to a static
  facade over HAPI's NarrativeTemplateManifest, delete the unused forked OpenmrsNarrativeTemplate.
- Add hapi-fhir-caching-caffeine so HAPI 7's caching SPI initialises at runtime; add junit-vintage-engine 6.1.0 so surefire's JUnit Platform
  provider discovers the still-on-JUnit-4 tests.
- Migrate every context-sensitive test (api + integration-tests, ~88 files) from JUnit 4 annotations to Jupiter so Spring 6's TestContext actually
  wires applicationContext; convert the lone `@Test(expected=…)` to assertThrows.
- Hibernate-6 entity fixes: `@OneToOne` → `@ManyToOne` on FhirDiagnosticReport.{code, subject, encounter}, FhirTask.{forReference,
  encounterReference, ownerReference, locationReference}, and FhirTaskParam.valueReference (which is inherited by FhirTaskInput / FhirTaskOutput) —
  drops the unintended UNIQUE(FK) constraint H2 was enforcing. `columnDefinition = "varchar(...)"` pinned on remaining `@Enumerated(STRING)` columns.
   FhirTask and FhirDiagnosticReport collections switched to final LinkedHashSet with clear+addAll setters so positional iteration is stable and
  `Set.of(...)` inputs don't trip a UnsupportedOperationException on merge.
- Search-query fixes: `cb.in(path).value(list)` → `path.in(list)` and typed subquery selects in FhirSearchQueryHelper / FhirPatientDaoImpl (clears
  ~55 `Cannot compare …` errors); join on BaseAttribute.owner instead of the convenience accessors LocationAttribute.location /
  ProviderAttribute.provider; type the ConceptMap.concept join instead of `Expression.as(Concept.class)`.
- DAO merge-vs-saveOrUpdate fix centralised in `BaseFhirDao.createOrUpdate` and `delete` (return the managed merge result, not the detached input);
   same correction applied to FhirConceptSourceDaoImpl, FhirContactPointMapDaoImpl, FhirAllergyIntoleranceDaoImpl, and
  FhirPatientIdentifierSystemDaoImpl. Redundant cascade-duplicating reaction loop dropped from FhirAllergyIntoleranceDaoImpl.
- MedicationRequestTranslatorImpl: any non-null DrugOrder.previousOrder now maps to MedicationRequest.priorPrescription regardless of action,
  covering O3's `action=NEW + previousOrder` renewal shape; the unused `RENEW → basedOn` branch is removed.
- DelegatingWebApplicationContext implements Spring 6's new `findAllAnnotationsOnBean(...)` and `getBeanProvider(ParameterizedTypeReference)`.
- Test fix-ups: align PersonSearchQueryTest NAME_MATCHER to the DB's case-insensitive collation and element-wise given/middle compare; bump 
  expected count in one ObservationSearchQueryTest case for the 2.8.0 dataset shift; set `condition-clinical` system URI in 
  ConditionFhirR3ResourceProviderTest so HAPI 7's R4→R3 convertor keeps the status; stub `getHeaderNames()` for HAPI 7's RestfulServer; set the 
  DrugIngredient back-ref in FhirMedicationDaoImplTest; delete two AllergyIntolerance NULL-enum tests that asserted a HAPI 5 quirk.
- Packaging: exclude slf4j-api from openmrs-api test-jar transitive and declare it provided so it isn't bundled into the module lib/ (the bundled 
  copy was tripping a LinkageError against the webapp's slf4j at startup).
- require_version lifted to `3.0.0 - 9.*`; the module is jakarta-only and won't classload on a 2.x server.


omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenmrsNarrativeTemplateManifest.java
```diff
@@ -7,289 +7,28 @@
...
-		} else {
-			throw new IOException("Invalid resource name: '" + name + "' (must start with classpath: or file: or openmrs:)");
-		}
-	}
-	
-	private static <T> List<INarrativeTemplate> getFromMap(EnumSet<TemplateTypeEnum> styles, T key,
-	        Map<T, List<OpenmrsNarrativeTemplate>> map) {
-		return map.getOrDefault(key, Collections.emptyList()).stream().filter(t -> styles.contains(t.getTemplateType()))
-		        .collect(Collectors.toList());
+		return NarrativeTemplateManifest.forManifestFileLocation(propertyFilePaths);
```
dkayiwa: **[blocking] This drops support for the `openmrs:` resource prefix.**

HAPI's `NarrativeTemplateManifest.loadResource()` only understands `classpath:` and `file:`, but the old `loadResource()` here also resolved `openmrs:<path>` against the OpenMRS application data directory. `NarrativeUtils.getValidatedPropertiesFilePath()` (untouched by this PR) still validates and forwards `openmrs:` paths for the `fhir2.narrativesOverridePropertyFile` GP, so any deployment using that prefix will now break — and only at the *first narrative generation*, since `OpenmrsThymeleafNarrativeGenerator.initialize()` is lazy (`InternalErrorException` at request time, not at startup).

Suggested minimal fix: before delegating to HAPI, rewrite `openmrs:rel/path` → `"file:" + new File(OpenmrsUtil.getApplicationDataDirectory(), "rel/path")`, and restore one of the deleted `NarrativeGeneratorTest` prefix tests to cover it. (If dropping the prefix is intentional, `NarrativeUtils` and the docs need updating too, and it should be called out as a breaking change.)

pom.xml
```diff
@@ -875,15 +892,15 @@
...
+        <maven.compiler.source>21</maven.compiler.source>
+        <maven.compiler.target>21</maven.compiler.target>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.build.outputEncoding>UTF-8</project.build.outputEncoding>
         <lombokVersion>1.18.38</lombokVersion>
-        <openmrsPlatformVersion>2.8.0</openmrsPlatformVersion>
-        <openmrsPlatformToolsVersion>2.8.0</openmrsPlatformToolsVersion>
-        <hapifhirVersion>5.7.9</hapifhirVersion>
-        <ucumVersion>1.0.3</ucumVersion>
+        <openmrsPlatformVersion>3.0.0-SNAPSHOT</openmrsPlatformVersion>
```
dkayiwa, line 890: **[note]** Expected at this stage of the platform cycle, but flagging: we can't cut a release while pinned to `3.0.0-SNAPSHOT`, and snapshot drift can break this build at any time. Worth a tracking ticket to pin to the platform 3.0.0 release before fhir2 5.0.0 ships.

dkayiwa (commented):
Reviewed the full diff. Overall this is a careful, well-diagnosed migration — the entity-mapping fixes (`@OneToOne` → `@ManyToOne`, `IDENTITY` ids matching the liquibase DDL), the `merge()` centralization in `BaseFhirDao`, and the Hibernate 6 criteria fixes (`path.in(...)`, typed subqueries, `BaseAttribute.owner` joins) are all correct, and the JUnit 4 → Jupiter migration is consistent (CI green on Java 25).

One functional regression to fix before merge — see the inline comment on `OpenmrsNarrativeTemplateManifest` — the rest are minor suggestions/questions.

A few items not tied to a single line:

- The PR description bullet about `MedicationRequestTranslatorImpl`/`priorPrescription` already landed on master via #605 (FM2-680) — worth trimming from the description so it isn't double-counted in the changelog.
- The two deleted AllergyIntolerance NULL-enum tests: if HAPI 7 now rejects `*.NULL` outright, consider replacing them with `assertThrows` instead of deleting, so the new behavior is pinned — the translators still carry now-(presumably-)dead `case NULL:` branches.
- The codecov patch-coverage failure (66%) is mostly the inlined Thymeleaf bridge in `OpenmrsThymeleafNarrativeGenerator` — a small narrative-generation round-trip test would cover the tag/attribute processors and the template-resolution path.
- Pre-existing (not this PR): `config.xml` still declares `conditionalResources` for `fhir2-api-2.5/2.6/2.7` jars the reactor no longer builds — candidate for a follow-up cleanup.

pom.xml
```diff
@@ -490,6 +495,18 @@
...
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.7.36</version>
```
dkayiwa, line 502: **[suggestion]** The bundling fix makes sense. Consider a version property here, and aligning it with the slf4j version core 3.0 actually ships (slf4j 2.x?) — compiling against 1.7.36 while the webapp runs a different major is mostly fine for this API surface, but matching the runtime's version avoids surprises.

api/pom.xml
```diff
@@ -74,6 +70,16 @@
 			<version>${project.parent.version}</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.junit.vintage</groupId>
+			<artifactId>junit-vintage-engine</artifactId>
+			<version>6.1.0</version>
```
dkayiwa: **[suggestion]** `6.1.0` is hardcoded here and again in `omod/pom.xml` and `integration-tests/pom.xml` — worth centralizing in the parent `dependencyManagement`. Also worth double-checking it matches the JUnit Platform version managed by openmrs-core's BOM: vintage 6.1.0 pulls junit-platform 6.x transitively, and mixed platform versions on a test classpath is a classic source of silent test-discovery failures (it resolves fine today per CI, but it's fragile).

integration-tests/pom.xml
```diff
@@ -41,8 +41,14 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>javax.servlet</groupId>
-            <artifactId>javax.servlet-api</artifactId>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
```
dkayiwa: **[question]** Is vintage still needed in this module? Every JUnit-4 test under `integration-tests` was migrated to Jupiter in this PR, so unlike `api`/`omod` (which still have JUnit-4 unit tests) this dependency looks unused.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirAllergyIntoleranceDaoImpl.java
```diff
@@ -45,13 +44,7 @@ public class FhirAllergyIntoleranceDaoImpl extends BaseFhirDao<Allergy> implemen
...
 	@Override
 	public Allergy createOrUpdate(@Nonnull Allergy allergy) {
-		Allergy savedAllergy = super.createOrUpdate(allergy);
-		
-		for (AllergyReaction reaction : allergy.getReactions()) {
-			getSessionFactory().getCurrentSession().saveOrUpdate(reaction);
-		}
-		
-		return savedAllergy;
+		return (Allergy) getSessionFactory().getCurrentSession().merge(allergy);
```
dkayiwa: **[suggestion]** This override is now byte-for-byte equivalent to `BaseFhirDao.createOrUpdate()` and can be deleted. Bonus: the base method carries `@Transactional` and this override doesn't (Spring doesn't inherit transaction attributes from overridden methods), so removing it actually restores the annotation's effect for direct DAO calls. The `(Allergy)` cast is also unnecessary — Hibernate 6's `Session.merge` is generic.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
```diff
@@ -70,14 +70,7 @@ protected <U> void setupSearchParams(@Nonnull OpenmrsFhirCriteriaContext<FhirTas
...
 	public FhirTask createOrUpdate(@Nonnull FhirTask task) throws DAOException {
-		// TODO: Refactor - and figure out why CascadeType.ALL does not take care of this.
-		if (task.getOwnerReference() != null && task.getOwnerReference().getReference() != null) {
-			getSessionFactory().getCurrentSession().saveOrUpdate(task.getOwnerReference());
-		}
-		
-		getSessionFactory().getCurrentSession().saveOrUpdate(task);
-		
-		return task;
+		return getSessionFactory().getCurrentSession().merge(task);
```
dkayiwa: **[suggestion]** Same as `FhirAllergyIntoleranceDaoImpl` — this override now duplicates `BaseFhirDao.createOrUpdate()` exactly (nice that the cascade/mapping fixes made the special-casing unnecessary!) and can be deleted, which also restores the base method's `@Transactional`.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseFhirDao.java
```diff
@@ -100,14 +100,14 @@ protected BaseFhirDao() {
 	
 	@Override
 	@Transactional
+	@SuppressWarnings("unchecked")
 	public T createOrUpdate(@Nonnull T object) {
-		getSessionFactory().getCurrentSession().saveOrUpdate(object);
-		
-		return object;
+		return (T) getSessionFactory().getCurrentSession().merge(object);
```
dkayiwa: **[nit]** Hibernate 6's `Session.merge` is generic (`<T> T merge(T)`), so the `(T)` cast and `@SuppressWarnings("unchecked")` here (and in `delete()`, `FhirConceptSourceDaoImpl`, `FhirContactPointMapDaoImpl`, `FhirPatientIdentifierSystemDaoImpl`) are unnecessary — `FhirTaskDaoImpl.createOrUpdate` already returns it uncast.

api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java
```diff
@@ -90,28 +89,35 @@ public enum TaskIntent {
 	 */
 	@OneToMany(cascade = CascadeType.ALL)
 	@JoinTable(name = "fhir_task_based_on_reference", joinColumns = @JoinColumn(name = "task_id"), inverseJoinColumns = @JoinColumn(name = "reference_id"))
-	private Set<FhirReference> basedOnReferences;
+	private final Set<FhirReference> basedOnReferences = new LinkedHashSet<>();
+	
+	public void setBasedOnReferences(Set<FhirReference> basedOnReferences) {
```
dkayiwa, line 101: **[suggestion]** The clear/addAll pattern is the right call for Hibernate 6 (keeps the `PersistentSet` intact and tolerates `Set.of(...)` inputs), but it has a self-assignment hazard: `task.setBasedOnReferences(task.getBasedOnReferences())` clears the set and then adds from the now-empty set, silently wiping it. A one-line guard makes the pattern safe:

```java
if (this.basedOnReferences == basedOnReferences) {
    return;
}
```

Applies to all five setters in this class and the three in `FhirDiagnosticReport`.

api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java
```diff
@@ -90,28 +89,35 @@ public enum TaskIntent {
...
+			this.basedOnReferences.addAll(basedOnReferences);
+		}
+	}
 	
 	/**
 	 * The entity who benefits from the performance of the service specified in the task (e.g., the
 	 * patient).
 	 */
-	@OneToOne(cascade = CascadeType.ALL)
+	@ManyToOne(cascade = CascadeType.ALL)
```
dkayiwa, line 109: **[thought]** The `@OneToOne` → `@ManyToOne` switch is clearly right (it drops the unintended UNIQUE FK). One carried-over wrinkle: now that sharing a `FhirReference` row between tasks is legal, `CascadeType.ALL` includes `REMOVE`, so purging one task would try to delete a reference another task may still point at. References aren't shared in practice today, so not blocking — but `{PERSIST, MERGE}` would be the safer cascade set for these four `@ManyToOne` reference fields.

api/src/main/java/org/openmrs/module/fhir2/model/FhirContactPointMap.java
```diff
@@ -49,11 +48,11 @@ public class FhirContactPointMap extends BaseOpenmrsData {
 	private Integer attributeTypeId;
 	
 	@Enumerated(EnumType.STRING)
-	@Column(length = 50)
+	@Column(length = 50, columnDefinition = "varchar(50)")
```
dkayiwa, line 51: **[nit]** `columnDefinition = "varchar(...)"` is dialect-specific DDL (fine for MySQL/MariaDB/Postgres/H2). Hibernate 6's `@JdbcTypeCode(SqlTypes.VARCHAR)` achieves the same “don't map `@Enumerated(STRING)` to a native ENUM column” effect dialect-neutrally, if you'd rather avoid hardcoded DDL fragments. Same applies to the pinned enum columns in `FhirTask` and `FhirDiagnosticReport`.

api/src/test/java/org/openmrs/module/fhir2/TestFhirSpringConfiguration.java
```diff
@@ -9,11 +9,25 @@
...
 
 @Configuration
 @ImportResource({ "classpath:applicationContext-service.xml", "classpath*:moduleApplicationContext.xml" })
 public class TestFhirSpringConfiguration {
 	
+	@Autowired
+	private DataSource dataSource;
+	
+	@PostConstruct
+	public void ensureShedlockTable() {
```
dkayiwa, line 30: **[nit]** Worth a one-line comment saying *why* this is needed (core 3.0's ShedLock-based scheduler expecting a table the test snapshot doesn't ship yet?) and referencing the core ticket, so it's obvious when this workaround can be deleted — same style as the existing “Needed until TRUNK-6299” notes.

api/src/test/java/org/openmrs/module/fhir2/api/search/ObservationSearchQueryTest.java
```diff
@@ -1728,7 +1728,7 @@ public void searchForLastnEncountersObs_shouldHandleNoPatientReferenceRequest()
 		assertThat(results, notNullValue());
 		List<IBaseResource> resultList = get(results);
 		
-		assertThat(resultList.size(), equalTo(2));
+		assertThat(resultList.size(), equalTo(3));
```
dkayiwa, line 1731: **[question]** Can you confirm this is an extra obs in core 3.0's standard test dataset rather than the lastn query returning an extra/duplicate row under Hibernate 6? This assertion is exactly where a join-semantics regression would hide, so it'd be good to pin down the cause before relaxing it.

dkayiwa:
Can we also address the merge conflicts?

api/src/main/java/org/openmrs/module/fhir2/model/FhirDiagnosticReport.java
```diff
@@ -48,23 +47,23 @@ public class FhirDiagnosticReport extends BaseOpenmrsData {
 	
 	@EqualsAndHashCode.Include
 	@Id
-	@GeneratedValue(strategy = GenerationType.AUTO)
+	@GeneratedValue(strategy = GenerationType.IDENTITY)
 	@Column(name = "diagnostic_report_id")
 	private Integer id;
 	
-	@Column(nullable = false)
+	@Column(nullable = false, columnDefinition = "varchar(255)")
```
dkayiwa: **[bug, low-impact]** This declares `varchar(255)`, but the module's liquibase creates `fhir_diagnostic_report.status` as **`varchar(50)`** (liquibase.xml:423, never widened by a later changeset) — the annotation contradicts the actual DDL. For contrast, the other pinned enum columns match their schema: `FhirTask.status`/`intent` = varchar(255) (liquibase:133/141) and `FhirContactPointMap.system`/`use` = varchar(50) (liquibase:1047-1048); this is the only mismatch. Impact is low today (no `hbm2ddl` validation in OpenMRS, and all `DiagnosticReportStatus` constants fit in 50 chars), but the wrong metadata would mis-validate/mis-generate DDL if ever used.

```suggestion
	@Column(nullable = false, columnDefinition = "varchar(50)")
```

omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenmrsThymeleafNarrativeGenerator.java
```diff
@@ -16,87 +16,241 @@
...
+import org.thymeleaf.templatemode.TemplateMode;
+import org.thymeleaf.templateresolver.DefaultTemplateResolver;
+import org.thymeleaf.templateresource.ITemplateResource;
+import org.thymeleaf.templateresource.StringTemplateResource;
 
 /**
  * Class for carrying out the task of narrative generation
  */
-public class OpenmrsThymeleafNarrativeGenerator extends ThymeleafNarrativeGenerator {
+public class OpenmrsThymeleafNarrativeGenerator extends BaseNarrativeGenerator {
```
dkayiwa, line 71: **[suggestion]** The header comment says the bridge was inlined because HAPI 7 removed `narrative2.ThymeleafNarrativeGenerator` — true, but its replacement survived: hapi-fhir-base 7.6.0 still ships **`ca.uhn.fhir.narrative.BaseThymeleafNarrativeGenerator`** (public abstract, `extends narrative2.BaseNarrativeGenerator`), containing this exact logic as `NarrativeTemplateResolver` / `NarrativeTagProcessor` / `NarrativeAttributeProcessor` (verified via `javap` on the 7.6.0 jar). The hooks line up: `getManifest()` is protected-abstract, `setMessageResolver(...)` is public, `applyTemplate`/`getStyle` are protected non-final — so this class could extend it and keep only the constructors, the `OpenmrsMessageResolver` wiring, and the lazy property-file manifest, deleting ~145 inlined lines (incl. the three inner classes) that will otherwise silently diverge from upstream narrative fixes.

Two things to weigh:
- HAPI's version is a slight **superset**: its `applyTemplate` also registers a `NarrativeGeneratorDialect` and exposes a `narrativeUtil` template variable (FhirPath support) — worth a sanity run of `NarrativeGeneratorTest`, but it's additive.
- It would *not* fix the per-call `TemplateEngine` rebuild — HAPI's own `getTemplateEngine()` also constructs a fresh engine per call (no cached field, verified in bytecode). So that cost is upstream-inherited either way; if you do keep the local copy, memoizing the engine in a field would actually improve on upstream (`applyTemplate` runs 1+N times per resource with nested `<th:narrative>` tags, rebuilding dialect + ~25 processors each time, and the `AlwaysValidCacheEntryValidity` template cache dies with each throwaway engine).

omod/src/main/java/org/openmrs/module/fhir2/narrative/OpenmrsNarrativeTemplateManifest.java
```diff
@@ -7,289 +7,28 @@
...
 	}
 	
-	private static <T> Map<T, List<OpenmrsNarrativeTemplate>> makeImmutable(
-	        Map<T, List<OpenmrsNarrativeTemplate>> styleToResourceTypeToTemplate) {
-		styleToResourceTypeToTemplate.replaceAll((key, value) -> Collections.unmodifiableList(value));
-		return Collections.unmodifiableMap(styleToResourceTypeToTemplate);
+	public static NarrativeTemplateManifest forManifestFileLocation(String... propertyFilePaths) throws IOException {
```
dkayiwa, line 47: **[suggestion]** As it stands this facade is a pure pass-through: both overloads return HAPI's `NarrativeTemplateManifest` (so callers depend on the HAPI type anyway, and HAPI exposes these exact `Collection`/varargs overloads itself) — the isolation the javadoc promises is illusory, and the class could be deleted with its 3 callers using HAPI directly. **But** the cleaner resolution ties into the `openmrs:` regression flagged above: give this facade a real job by resolving `openmrs:`-prefixed paths against the application data directory (rewriting to `file:`) before delegating to HAPI. Either outcome is fine — a facade with behavior, or no facade — just not a no-op layer.

dkayiwa (commented):
Second, deeper pass over the full diff (multi-angle sweep, each finding adversarially verified against the working tree, the HAPI 7.6.0 jars, and core 3.0.0-SNAPSHOT). **Nothing new rises to blocking** — the 10 findings below are correctness-metadata/maintainability tier.

Also worth recording what was checked and came back **clean**, since it de-risks the scariest parts of this migration:
- every production caller of `createOrUpdate`/`delete` consumes the `merge()` return value (no caller keeps using the detached argument);
- the `clear+addAll` collection setters are never fed from their own getter in translator paths;
- the `owner` joins match the platform-3.0 attribute mappings (`PersonAttribute.person` correctly left alone);
- `DelegatingWebApplicationContext` forwards every Spring 6 method to the right delegate;
- the JUnit migration has no silent-lifecycle traps (no JUnit-4 `@After`/`@Ignore` left in Jupiter classes; all 39 integration-test `setup()` overrides re-annotated with `@BeforeEach`);
- the inlined Thymeleaf bridge is functionally faithful to upstream (all narrative tests pass).

omod/pom.xml
```diff
@@ -86,14 +86,20 @@
...
 		</dependency>
+		<dependency>
+			<groupId>org.junit.vintage</groupId>
+			<artifactId>junit-vintage-engine</artifactId>
+			<version>6.1.0</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>ca.uhn.hapi.fhir</groupId>
+			<artifactId>hapi-fhir-caching-caffeine</artifactId>
```
dkayiwa: **[nit]** `hapi-fhir-caching-caffeine` is declared three times (api, omod, integration-tests). The api declaration (compile scope) is enough: I verified by building the `.omod` with this omod declaration removed — `lib/` still contains `hapi-fhir-caching-caffeine-7.6.0.jar`, `hapi-fhir-caching-api-7.6.0.jar` and `caffeine-2.9.3.jar` via the transitive fhir2-api dependency (the OpenMRS packaging bundles transitives, same as attoparser/unbescape land today without direct declarations). integration-tests likewise gets it test-scoped through fhir2-omod → fhir2-api. Suggest keeping only the api declaration so future version/exclusion tweaks have one home.

api/src/main/java/org/openmrs/module/fhir2/model/FhirDiagnosticReport.java
```diff
@@ -73,11 +72,25 @@ public class FhirDiagnosticReport extends BaseOpenmrsData {
 	
 	@OneToMany
 	@JoinTable(name = "fhir_diagnostic_report_performers", joinColumns = @JoinColumn(name = "diagnostic_report_id"), inverseJoinColumns = @JoinColumn(name = "provider_id"))
-	private Set<Provider> performers = new HashSet<>();
+	private final Set<Provider> performers = new LinkedHashSet<>();
+	
+	public void setPerformers(Set<Provider> performers) {
```
dkayiwa, line 79: **[suggestion]** This `clear(); if (x != null) addAll(x);` body is stamped out **seven times** — three setters here (`performers`/`results`/`orders`) and four in `FhirTask` (`basedOnReferences`/`input`/`output`/`partOfReferences`), token-identical modulo field name. Worth a single static helper, e.g. in the existing `api/util` package:

```java
public static <T> void replaceContents(Collection<T> target, Collection<T> source) {
    if (target == source) {
        return;
    }
    target.clear();
    if (source != null) {
        target.addAll(source);
    }
}
```

(a static util rather than a base-class method, since the two entities extend different core bases). Bonus: the self-assignment guard from the earlier `FhirTask` comment then gets fixed in one place instead of seven.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirContactPointMapDaoImpl.java
```diff
@@ -119,11 +119,9 @@ public FhirContactPointMap saveFhirContactPointMap(@Nonnull FhirContactPointMap
 			existingContactPointMap.setSystem(contactPointMap.getSystem());
 			existingContactPointMap.setUse(contactPointMap.getUse());
 			existingContactPointMap.setRank(contactPointMap.getRank());
-			sessionFactory.getCurrentSession().merge(existingContactPointMap);
-			return existingContactPointMap;
+			return (FhirContactPointMap) sessionFactory.getCurrentSession().merge(existingContactPointMap);
```
dkayiwa: **[nit]** Now that both branches end in the same `merge(...)` (previously `merge` vs `saveOrUpdate`), the if/else can collapse to picking the target and merging once:

```java
FhirContactPointMap target = contactPointMap;
if (existingContactPointMap != null) {
    existingContactPointMap.setSystem(contactPointMap.getSystem());
    existingContactPointMap.setUse(contactPointMap.getUse());
    existingContactPointMap.setRank(contactPointMap.getRank());
    target = existingContactPointMap;
}
return (FhirContactPointMap) sessionFactory.getCurrentSession().merge(target);
```

api/src/main/java/org/openmrs/module/fhir2/model/FhirContactPointMap.java
```diff
@@ -49,11 +48,11 @@ public class FhirContactPointMap extends BaseOpenmrsData {
...
 	private ContactPointSystem system;
 	
 	@Enumerated(EnumType.STRING)
-	@Column(length = 50)
+	@Column(length = 50, columnDefinition = "varchar(50)")
```
dkayiwa, line 55: **[nit]** When `columnDefinition` is present, JPA ignores `length` — so `length = 50` here (and on `system` above) is inert and the width is stated twice. Dropping the redundant attribute avoids the two silently diverging later; the pinned enum columns in `FhirTask`/`FhirDiagnosticReport` already omit it.

api/src/test/java/org/openmrs/module/fhir2/api/search/PersonSearchQueryTest.java
```diff
@@ -106,27 +105,48 @@ public class PersonSearchQueryTest extends BaseFhirContextSensitiveTest {
 	
 	private static final int END_INDEX = 10;
 	
+	// Mirrors the DB ORDER BY on PersonName columns, which under H2's MySQL-mode collation is
+	// case-insensitive -- so lower-case 'o' in "of Cos" sorts between upper-case 'N' and 'P',
+	// not after 'Z' as Java's default String.compareTo would have it.
+	private static int compareIgnoreCase(String a, String b) {
```
dkayiwa, line 111: **[nit]** This helper is exactly `Comparator.nullsLast(String.CASE_INSENSITIVE_ORDER)` (null→last, both-null→0, else `compareToIgnoreCase`) — a
```java
private static final Comparator<String> NAME_PART = Comparator.nullsLast(String.CASE_INSENSITIVE_ORDER);
```
replaces the 6-line method and can't drift on the null-ordering.

Related observation while verifying this (pre-existing, fine as a follow-up): the production sort this comparator mirrors (`BasePersonDao` `SP_NAME` → familyName/familyName2/givenName/middleName/prefix/suffix) has **no deterministic tiebreaker** (e.g. personId), so equal-name rows have DB-dependent order. Adding a stable tiebreaker to the query would give clients stable pagination and let this test assert one canonical order instead of mirroring collation quirks.

api/src/test/java/org/openmrs/module/fhir2/api/search/ObservationSearchQueryTest.java
```diff
@@ -1728,7 +1728,7 @@ public void searchForLastnEncountersObs_shouldHandleNoPatientReferenceRequest()
...
 		assertThat(getDistinctEncounterDatetime(resultList), lessThanOrEqualTo(2));
```
dkayiwa, line 1732: **[suggestion]** Complementing the cause question above: whatever the answer, the assertion only counts rows, so a lastn regression that returns three *wrong* observations still passes — which is also why the magic number could silently flip 2→3. Pinning identity instead (collect the returned obs `getIdElement().getIdPart()`s into a set and compare against the expected most-recent uuids from the dataset) makes the test fail when the wrong rows come back, and the count becomes an implied consequence.

api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImplTest.java
```diff
@@ -38,7 +37,7 @@
 import org.openmrs.module.fhir2.model.FhirTaskOutput;
 import org.springframework.beans.factory.annotation.Autowired;
 
-public class FhirTaskDaoImplTest extends BaseFhirContextSensitiveTest {
+class FhirTaskDaoImplTest extends BaseFhirContextSensitiveTest {
```
dkayiwa, line 48: **[nit]** This migration made this file the lone outlier: package-private class + 16 package-private `void` test methods, while all ~24 sibling DaoImpl/SearchQuery tests kept `public class` / `public void`. Either convention works under Jupiter — but one convention per package, please (and keeping `public` here would also shrink this file's diff to the real changes).

api/src/test/java/org/openmrs/module/fhir2/TestFhirSpringConfiguration.java
dkayiwa: Follow-up — I dug into the *why* so the comment can be precise: core's `SchedulerConfig` (component-scanned from `applicationContext-service.xml`) wires a ShedLock `JdbcTemplateLockProvider` at **context-init** time, but core's jupiter `BaseContextSensitiveNonTransactionalTest` only creates the `shedlock` table inside `initializeInMemoryDatabase()` at **@BeforeEach** time (the test schema comes from hbm2ddl, not core's liquibase, so the `20260212-create-shedlock-table` changeset never runs there). This `@PostConstruct` front-loads core's own workaround to context init — i.e. it's mirroring an established core pattern, not a temporary hack. Suggested comment text: `// core's SchedulerConfig (ShedLock LockProvider) needs this table at context init; core's test base only creates it at @BeforeEach since the test schema comes from hbm2ddl, not liquibase`.

wikumChamith:
Thanks for the review @dkayiwa. However, we'll have to pause this PR for a little while as @ibacher is working on some features in this module, and merging this backport from a 3.0 supporting version would just make those more complicated.

dkayiwa:
@wikumChamith are we ready to resume on this?

wikumChamith:
> @wikumChamith are we ready to resume on this?

@ibacher ??

dkayiwa:
@claude review

wikumChamith:
@dkayiwa I've addressed the issues you pointed out. Note that I also changed the target branch to a new platform-3 branch instead of master.

cc: @ibacher

api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java
```diff
@@ -95,28 +97,32 @@ public enum TaskIntent {
...
 	@JoinColumn(name = "encounter_reference_id", referencedColumnName = "reference_id")
 	private FhirReference encounterReference;
 	
 	/**
 	 * Individual organization or Device currently responsible for task execution.
 	 */
-	@OneToOne(cascade = CascadeType.ALL)
+	@ManyToOne(cascade = CascadeType.ALL)
```
dkayiwa, line 124: Just below this hunk, `focusReference` (which came in from FM2-682 during the rebase) still has `@OneToOne(cascade = CascadeType.ALL)` while this PR converts its four sibling reference fields to `@ManyToOne`. The reason for the switch applies to focus too: Hibernate 6 puts a unique index on a `@OneToOne` join column in the hbm2ddl test schema, so the first test dataset that lets two tasks share a focus reference fails on a constraint production doesn't have (the liquibase FK `task_focus_reference_fk` has no unique). I reproduced it by pointing tasks 101 and 102 in `FhirTaskDaoImplTest_owner_data.xml` at the same reference: 9 errors in `TaskSearchQueryTest`, all `Unique index or primary key violation: ... FHIR_TASK(FOCUS_REFERENCE_ID ...)`. Nothing shares a focus reference today so this isn't blocking, but I'd flip it to `@ManyToOne` in this PR while the sibling fields are changing anyway.

api/src/test/java/org/openmrs/module/fhir2/api/impl/FhirPractitionerServiceImplTest.java
```diff
@@ -67,7 +69,8 @@
 import org.openmrs.module.fhir2.api.search.param.SearchParameterMap;
 import org.openmrs.module.fhir2.api.translators.PractitionerTranslator;
 
-@RunWith(MockitoJUnitRunner.class)
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
```
dkayiwa: The Jupiter conversion adds class-level `Strictness.LENIENT` here and in 21 other files (most translator tests, the Task and Patient web tests, `AuthenticationFilterTest`, `OpenmrsFhirAddressStrategyTest`). It is needed as-is: without it this class fails with per-test `UnnecessaryStubbingException`, since the old JUnit 4 runner counted a shared `@Before` stub as used if any test in the class used it, while `MockitoExtension` checks each test's stubs individually. But LENIENT is the bluntest setting: it also turns off argument-mismatch detection (`PotentialStubbingProblem`) for every mock in these files, the check that catches a stub silently returning null after a signature change. `@MockitoSettings(strictness = Strictness.WARN)` is a one-word drop-in that keeps all of these passing while still reporting both kinds of problem, so I'd use WARN here (or mark just the shared setup stubs `lenient()` in files where it's only one or two). Not something I'd hold the PR for.

api/src/test/java/org/openmrs/module/fhir2/api/search/ObservationSearchQueryTest.java
dkayiwa: It's the dataset: core 3.0's standardTestDataset moves encounter 6's encounter_datetime from 2008-08-19 to 2007-12-01 (the obs rows are unchanged, 10 in both 2.8.0 and 3.0.0-SNAPSHOT), which changes which encounters make up the two-most-recent window for lastn, so a third obs is legitimately in range. Not a Hibernate 6 query change, so nothing to do here.

wikumChamith:
@dkayiwa I've updated the PR.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/BaseFhirDao.java
```diff
@@ -101,9 +100,7 @@ protected BaseFhirDao() {
 	@Override
 	@Transactional
 	public T createOrUpdate(@Nonnull T object) {
-		getSessionFactory().getCurrentSession().saveOrUpdate(object);
-		
-		return object;
+		return getSessionFactory().getCurrentSession().merge(object);
```
dkayiwa, line 103: Running the integration suite (`mvn verify -Pintegration-test`, which CI doesn't execute, see the review body) surfaces two write-path regressions against core 3.0. The six affected test classes pass 195/195 on the base branch; here, 12 of those tests fail. If this merges as-is, POST Practitioner and PUT Medication return 500 on every platform-3 deployment:

- `shouldCreateNewPractitionerAsJson`/`AsXML` (r3 and r4): the request dies with `TransientPropertyValueException: Persistent instance of 'org.openmrs.Provider' references an unsaved transient instance of 'org.openmrs.Person'`. `PractitionerTranslatorProviderImpl` news up a `Person` to hold the incoming names, and platform 2.8 persisted it through `Provider.hbm.xml`'s `cascade="save-update"` back when this method was `saveOrUpdate`. Core 3.0 maps `Provider.person` as a plain `@ManyToOne` with no cascade (I checked the 3.0.0-SNAPSHOT jar), so nothing saves the person anymore and the flush fails. The practitioner create path needs to persist that person explicitly now (or core grows a cascade on `Provider.person`, but the module has to work against current snapshots either way).

- `shouldUpdateExistingMedicationAsJson`/`AsXml` (r3 and r4): `HibernateException: A collection with orphan deletion was no longer referenced by the owning entity instance: org.openmrs.Drug.ingredients` (the XML variants surface a `MergeContext` NPE instead). `MedicationTranslatorImpl.toOpenmrsType` swaps in a brand-new ingredients collection via `existingDrug.setIngredients(...)`. The 2.8 mapping (`<set inverse="true">`, no cascade) tolerated that; core 3.0 declares `Drug.ingredients` with `cascade = ALL, orphanRemoval = true`, and Hibernate 6 refuses a collection-instance swap on a managed entity. Mutating in place should fix it, e.g. `GeneralUtils.replaceContents(existingDrug.getIngredients(), ingredients)` inside the existing `hasIngredient()` branch, the same pattern the module's own entities adopted in this PR. (The api-level `FhirMedicationDaoImplTest` stays green because it never replaces the collection on a managed drug, so its passing isn't coverage for this.)

The failing tests themselves are the acceptance check for both fixes.

api/src/test/java/org/openmrs/module/fhir2/api/search/ObservationSearchQueryTest.java
dkayiwa, line 1731: The integration twins of this test still expect the old dataset: `shouldReturnLastnEncountersObservationsWhenPatientReferenceIsMissingAsJson`/`AsXml` in both the r3 and r4 `ObservationFhirResourceProviderIntegrationTest` assert `total = 2` and now fail with `property 'total' was <3>` under `-Pintegration-test`, for exactly the dataset shift this line accounts for. They need the same 2->3 bump before merge, or the suite fails for whoever does run it (CI currently doesn't, see the review body).

api/src/test/java/org/openmrs/module/fhir2/TestFhirSpringConfiguration.java
```diff
@@ -9,11 +9,27 @@
...
 @ImportResource({ "classpath:applicationContext-service.xml", "classpath*:moduleApplicationContext.xml" })
 public class TestFhirSpringConfiguration {
 	
+	@Autowired
+	private DataSource dataSource;
+	
+	// core's SchedulerConfig (ShedLock LockProvider) needs this table at context init; core's test base
+	// only creates it at @BeforeEach since the test schema comes from hbm2ddl, not liquibase
+	@PostConstruct
+	public void ensureShedlockTable() {
```
dkayiwa, line 30: Worth copying this workaround into the integration-tests module's `IntegrationTestConfiguration` too: it imports the same core context but has no shedlock bootstrap, so a `-Pintegration-test` run logs a `LockException: Table "SHEDLOCK" not found` stack from core's scheduler on every context start. Nothing fails from it today, so fine to leave if you'd rather keep this PR contained, but it buries the real failures in that log.

dkayiwa (commented):
This needs one more round before it merges: the platform bump breaks two production write paths, which the inline comments cover, and the reason nothing flagged them is that the retarget to platform-3 switched CI off for this PR.

`build.yml` triggers on pushes and PRs targeting master only, so every commit since the base change ran SonarCloud static analysis and nothing else (8f6648f was the last commit CI compiled and tested, back when this still targeted master). Merged as-is, the filter leaves the platform-3 branch itself without CI as well: follow-up PRs against it land with no build or test run at all. Since this PR is what establishes the branch, I'd wire it back on here:

```yaml
on:
  push:
    branches: [ master, platform-3 ]
  pull_request:
    branches: [ master, platform-3 ]
```

pull_request workflows are read from the merge ref, so the Build workflow starts running on this PR as soon as that edit is pushed to it (a maintainer pushing the same change directly to platform-3 works too). Two adjacent gaps to be aware of, both fine as follow-ups: the #619 coverage upload exists only on master, so platform-3 keeps no coverage reporting until it's brought over; and the shared build workflow runs plain `clean install`, so the `-Pintegration-test` failsafe suite never executes in CI on any branch. That last one is pre-existing (master behaves the same), but it is exactly where this PR's regressions were hiding, so `maven_args: 'clean install -Pintegration-test'` in build.yml, or at least a ticket for it, would be money well spent.

Unrelated to code: the description still says junit-vintage-engine 6.1.0 is added so the not-yet-migrated JUnit 4 tests keep running, but c17a3ee removed vintage and migrated everything to Jupiter. Worth a description pass before merging, since a squash merge turns it into the platform-3 commit message.

I'd hold the merge for the two write-path regressions, the observation IT totals, and the build.yml filter; the rest is take-or-leave.

wikumChamith:
@dkayiwa I updated the PR.

api/src/test/java/org/openmrs/module/fhir2/api/search/ObservationSearchQueryTest.java
dkayiwa: Collecting into a `Set` loses the duplicate check that `size() == 3` gave: a result of `[A, B, C, A]` collapses to `{A, B, C}` and still passes, and a duplicated row is the kind of join regression this assertion was meant to catch. My earlier comment said set, so that one's on me. Using `.toList()` instead of `.collect(Collectors.toSet())` fixes it (I ran the test with that change and it passes). Not a merge blocker.

wikumChamith: Done!

PR #607, opened 2026-05-28
FM2-682: Add Task.focus support to the FHIR2 Task model

vvkpd:
## JIRA
[FM2-682](https://openmrs.atlassian.net/browse/FM2-682)

## Summary

This PR adds native support for the `Task.focus` element to the OpenMRS FHIR2 module, bringing it into full alignment with the [HL7 FHIR R4 Task specification](https://www.hl7.org/fhir/task.html).

The proposal was discussed and acknowledged by the community at:
https://talk.openmrs.org/t/proposal-to-add-task-focus-support-to-the-fhir2-module-task-model/49497

## Problem

The current `FhirTask` model (v2.5.0) has no native field for `Task.focus`. In FHIR, `Task.focus` is the element that references the specific resource a task is acting upon — distinct from `Task.basedOn` (the authorizing request) and `Task.for` (the beneficiary). Without it, implementers must misuse `Task.input` or `Task.basedOn` to carry this reference, introducing semantic ambiguity in clinical workflows.

## Changes

- **`FhirTask` entity** — adds `focusReference` (`@OneToOne FhirReference`, column `focus_reference_id`)
- **Liquibase migration** — `add_fhir_task_focus_reference_20260507`: idempotent changeSet adding the FK column to `fhir_task`, referencing `fhir_reference`
- **`TaskTranslatorImpl`** — bidirectional mapping: `Task.focus` ↔ `FhirTask.focusReference` in both `toFhirResource()` and `toOpenmrsType()`
- **`TaskSearchParams`** — adds `focusReference` field and `FOCUS_REFERENCE_SEARCH_HANDLER`
- **`FhirTaskDaoImpl`** — handles `FOCUS_REFERENCE_SEARCH_HANDLER` in `setupSearchParams` via a join on `fo` alias; also fixes `handleReference` to match on `targetUuid` (bare UUID) instead of the `reference` column (full `Type/UUID` string) — this is a correctness fix that applies to all four Task reference search params (`basedOn`, `owner`, `for`, `focus`) and affects both R3 and R4 since they share the same DAO
- **`TaskFhirResourceProvider` (R4)** — exposes `@OptionalParam(name = Task.SP_FOCUS)` on the `searchTasks` endpoint
- **Tests** — unit tests for translator (bidirectional), DAO (round-trip + search-by-focus), R4 provider; integration test (create + read with focus, search-by-focus)

R3 provider search params are intentionally unchanged — `Task.focus` search is R4-only scope. However, the `handleReference` fix improves R3 reference search correctness as a side effect.

## Test plan

- [ ] `mvn test -Dtest=TaskTranslatorImplTest` — translator bidirectional mapping
- [ ] `mvn test -Dtest=TaskFhirResourceProviderTest` — provider search-by-focus parameter
- [ ] `mvn test -Dtest=FhirTaskDaoImplTest` — DAO round-trip and search
- [ ] `mvn verify -P integration-test -Dtest=TaskFhirResourceIntegrationTest` — integration: create/read with focus, search-by-focus
- [ ] `mvn clean verify` — full build with all unit tests

[FM2-682]: https://openmrs.atlassian.net/browse/FM2-682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vvkpd:
https://talk.openmrs.org/t/proposal-to-add-task-focus-support-to-the-fhir2-module-task-model/49497

api/src/main/resources/liquibase.xml
```diff
@@ -1095,6 +1095,22 @@
                                  referencedColumnNames="concept_id"/>
     </changeSet>
 
+    <changeSet id="add_fhir_task_focus_reference_20260507" author="cure-dev" dbms="mysql,mariadb,postgresql,h2">
```
dkayiwa: `author="cure-dev"` and `dbms="...,h2"` both diverge from this file's conventions:

- **`author`** - every other changeSet uses a real contributor handle (`pmanko`, `ibacher`, `moses_mutesa`, `r.kalai`, `mherman22`). `cure-dev` looks like a leftover identity from the downstream Bahmni fork these commits originated in. Since `author` is part of a changeSet's unique identity (`id` + `author` + file), please use your own handle (e.g. `vvkpd`).
- **`dbms="mysql,mariadb,postgresql,h2"`** - the direct precedent, `add_task_location_reference_20221121` (line 1014), uses `dbms="mysql,mariadb,postgresql"` (no `h2`), as do all other dbms-scoped changeSets here. The H2 schema used by the context-sensitive tests is generated by Hibernate `hbm2ddl` from the `@Entity` mapping (via `BaseModuleContextSensitiveTest`), not by this changeSet - so `h2` here has no effect on the tests and only diverges from precedent. Suggest matching the location changeSet: `dbms="mysql,mariadb,postgresql"`.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
```diff
@@ -52,6 +52,11 @@ protected <U> void setupSearchParams(@Nonnull OpenmrsFhirCriteriaContext<FhirTas
 					        .forEach(param -> handleReference(criteriaContext, (ReferenceAndListParam) param.getParam(),
 					            "forReference", "f").ifPresent(criteriaContext::addPredicate));
 					break;
+				case FhirConstants.FOCUS_REFERENCE_SEARCH_HANDLER:
+					entry.getValue()
+					        .forEach(param -> handleReference(criteriaContext, (ReferenceAndListParam) param.getParam(),
+					            "focusReference", "focr").ifPresent(criteriaContext::addPredicate));
```
dkayiwa: Minor: the neighbouring reference handlers use short aliases (`bo`, `o`, `f`). `focr` is unique and works, but per the commit history the `fr` -> `focr` rename was to dodge an alias collision in the downstream `BahmniFhirTaskDaoImpl` that doesn't exist upstream. Fine to keep - just noting the rationale is fork-specific; a shorter alias in keeping with the neighbours (e.g. `fo`) would read more naturally here.

api/src/main/java/org/openmrs/module/fhir2/api/search/param/TaskSearchParams.java
```diff
@@ -46,30 +46,45 @@ public class TaskSearchParams extends BaseResourceSearchParams {
...
 		this.taskCode = taskCode;
 		this.status = status;
 		
 	}
 	
+	/**
+	 * Backward-compatible 9-arg overload (without focusReference) so existing callers (R3 provider,
+	 * ServiceRequest translator, etc.) continue to compile unchanged.
+	 */
+	public TaskSearchParams(ReferenceAndListParam basedOnReference, ReferenceAndListParam ownerReference,
```
dkayiwa, line 76: Keeping a backward-compatible overload for the existing positional callers (R3 provider, `ServiceRequestTranslatorImpl`, and several tests) is a reasonable, low-churn choice. Two notes:

- The 9- and 10-arg constructors now differ only by `focusReference` inserted at position 4, so the two adjacent `ReferenceAndListParam` runs are easy to transpose at a call site. Since the R4 provider already moved to `TaskSearchParams.builder()`, steering remaining/new callers to the builder would let this positional overload be retired later.
- For reference: `@EqualsAndHashCode(callSuper = true)` + `@Data` picks up the new `focusReference` field automatically, so equality/caching stays correct - and that's why the existing `new TaskSearchParams(...nulls...)` Mockito matchers still match the builder-produced object (both have `focusReference == null`).

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImpl.java
```diff
@@ -203,6 +207,10 @@ private void setOpenmrsTaskFields(FhirTask openmrsTask, Task fhirTask) {
 			openmrsTask.setOwnerReference(referenceTranslator.toOpenmrsType(fhirTask.getOwner()));
 		}
 		
+		if (fhirTask.hasFocus()) {
```
dkayiwa, line 210: Consistent with the sibling references (`owner`/`for`/`location`), which also only assign inside `has...()`. One thing worth flagging: because there's no `else`, an update that *removes* `Task.focus` won't clear the stored `focusReference` - the old reference persists. That's the pre-existing behavior for every reference on this entity, so not a blocker, but there's no test for focus removal. If "unset focus" is a workflow you need, it'd require an explicit `else { openmrsTask.setFocusReference(null); }` (plus orphan handling, since the mapping has no `orphanRemoval`).

api/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceProviderTest.java
```diff
@@ -285,6 +294,29 @@ public void searchTasks_shouldAddRelatedServiceRequestResourceToResultListWhenIn
...
 	}
 	
+	@Test
+	public void searchTasks_shouldReturnMatchingTasksByFocusReference() {
+		//given
+		List<Task> tasks = new ArrayList<>();
+		tasks.add(task);
+
+		ReferenceAndListParam focusReference = new ReferenceAndListParam()
+		        .addAnd(new ReferenceOrListParam().add(new ReferenceParam("Observation", null, OBSERVATION_UUID)));
```
dkayiwa: Nit: prefer `FhirConstants.OBSERVATION` over the literal `"Observation"` here - the DAO test (`FhirTaskDaoImplTest`) uses the constant, so this keeps the two focus tests consistent.

dkayiwa (commented):
## Review: FM2-682 - Add `Task.focus` support

Thanks for this - the change is well structured and the feature looks correct end to end (entity -> translator -> search param -> DAO -> R4 provider), with good coverage at the translator, DAO, provider, and integration layers. `Task.focus` is 0..1 in R4, which the `@OneToOne FhirReference` model and the single `setFocus`/`hasFocus` translation correctly reflect. The R4 builder migration in the provider is a nice readability win, and HAPI will auto-add `focus` to the R4 Task CapabilityStatement from the new `@OptionalParam` (no conformance test enumerates Task params, so nothing breaks there).

### Worth addressing
- **Liquibase `author`/`dbms`** (see inline) - the fork-specific `cure-dev` author and the non-standard `h2` in `dbms` should be brought in line with the rest of the file.

### Questions / suggestions (non-blocking)
- **`FhirTaskDaoImpl.createOrUpdate`** carries an explicit `saveOrUpdate(ownerReference)` workaround with a `// TODO ... figure out why CascadeType.ALL does not take care of this` comment. `focusReference` uses the same `@OneToOne(cascade = ALL)` mapping but relies purely on cascade. The integration test confirms the *create* path persists focus, so this is probably fine - but since `owner` apparently needed the manual save, it's worth confirming there's no reference-reuse/update path where `focus` would hit the same issue. (Can't comment inline - that method isn't in the diff.)
- **R3 parity** - the PR scopes the search param to R4, but the translator is shared, so R3 *reads* already surface `Task.focus` (STU3 Task has a `focus` element). Exposing the `focus` search param on the R3 provider too would be a one-liner and keep R3/R4 search symmetric; fine to defer if explicitly out of scope.
- **Integration test** - `observationUuid = "5085AAAA..."` is duplicated as a local literal in both new tests (lines 801 and 828); a shared constant would be tidier.
- **Whitespace** - `TaskFhirResourceProviderTest` has a couple of unrelated blank-line whitespace changes (e.g. around `mockRequestDetails`); reverting them keeps the diff focused.

Overall this is solid, focused work with the right tests. LGTM once the Liquibase `author`/`dbms` items are sorted.

api/src/main/resources/liquibase.xml
vvkpd: Fixed — changed `author` to `vvkpd` and removed `h2` from `dbms` to match the `add_task_location_reference_20221121` precedent. Commit: 5c995fdb.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
vvkpd: Good catch — renamed the alias from `focr` to `fo` to keep it consistent with the neighbouring handlers (`bo`, `o`, `f`). The `fr` → `focr` rename was indeed a downstream Bahmni artefact. Commit: 5c995fdb.

api/src/main/java/org/openmrs/module/fhir2/api/search/param/TaskSearchParams.java
vvkpd: Thanks for the explanation on `@EqualsAndHashCode` + `@Data` picking up the new field automatically — good to know the existing null-argument Mockito matchers still match correctly. Agreed that new callers should use the builder; the positional overload can be retired in a follow-up once the remaining callers are migrated. No change needed here.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImpl.java
vvkpd: Acknowledged — the missing `else` branch is consistent with how `owner`/`for`/`location` behave today, so not changing it here. "Unset focus" is not a required workflow at this stage; if it becomes one it would need an explicit `else { openmrsTask.setFocusReference(null); }` plus `orphanRemoval` on the mapping, which is a separate concern.

api/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceProviderTest.java
vvkpd: Fixed — replaced the `"Observation"` literal with `FhirConstants.OBSERVATION` to match the DAO test. Also extracted the duplicated `observationUuid` literal in the integration test to a shared `FOCUS_OBSERVATION_UUID` constant and normalized blank-line whitespace in the new test methods. Commit: 5c995fdb.

dkayiwa:
@vvkpd when you run this, does it pass? `mvn verify -P integration-test -pl integration-tests -Dit.test='TaskFhirResourceIntegrationTest#shouldSearchTasksByFocusReferenceAsJson'`

vvkpd:
@dkayiwa It does now — but it required a fix. The test was originally written to create a task via the FHIR API and then search for it, but the upstream `ReferenceTranslatorImpl` stores the full `"Observation/UUID"` string in the `reference` column, while the DAO's `handleReference` correctly compares against `param.getIdPart()` (bare UUID) — which matches production behavior (verified on our QA environment where the `reference` column stores bare UUIDs).

The fix (commit `09f3daec`): replaced the create-via-API approach in the search test with a pre-seeded `fhir_reference` + `fhir_task` entry in the shared test data fixture, storing the bare UUID consistent with production. The test now passes:

```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
```

test-data/src/main/resources/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImplTest_initial_data.xml
```diff
@@ -84,4 +84,8 @@
 
 	<fhir_task task_id="10" name="Task with for reference and code " status="REQUESTED" task_code="5030" for_reference_id="5" intent="ORDER" creator="1" date_created="2024-04-07 00:00:00.0" date_changed="2024-04-07 00:00:00.0" retired="false" uuid="271a44b4-dbc8-4b83-8d3d-c794339d06e9"/>
 
+	<!-- Task.focus -->
+	<fhir_reference reference_id="7" name="focus reference" target_type="Observation" reference="5085AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" uuid="a1b2c3d4-e5f6-7890-abcd-ef1234567890" creator="1" date_created="2024-01-01 00:00:00.0" retired="false"/>
+	<fhir_task task_id="11" name="Task with focus reference" status="REQUESTED" focus_reference_id="7" intent="ORDER" creator="1" date_created="2024-04-07 00:00:00.0" date_changed="2024-04-07 00:00:00.0" retired="false" uuid="a1b2c3d4-e5f6-7890-abcd-ef1234567891"/>
```
dkayiwa: issue (blocking): This new `REQUESTED` task (with `reference_id=7`) lands in the **shared** `FhirTaskDaoImplTest_initial_data.xml`, which is loaded by four test classes (`TaskSearchQueryTest`, `FhirTaskDaoImplTest`, and the R3 + R4 `TaskFhirResourceIntegrationTest`). It is the 6th non-retired `REQUESTED` task in the fixture, so it breaks every hard-coded `REQUESTED` / `ACCEPTED-or-REQUESTED` count there. If merged as-is, CI fails.

I built `test-data` 4.1.0-SNAPSHOT fresh and ran the suites — 9 failures:

- `api` `TaskSearchQueryTest#searchForTasks_shouldReturnTasksByMultipleStatusOr` (line 413): `Expected: <7> but: was <8>`. The comment you added on that line — "tasks 1,2,3,4,8,9,10" — omits this task (#11), which is the tell.
- `integration` **R4** `TaskFhirResourceIntegrationTest`: `shouldReturnCountForTaskAsJson` / `shouldReturnCountForTaskAsXml` (`total` was 6, expected 5) and `shouldSearchForFilteredAndSortedTasksAsJson` / `shouldSearchForFilteredAndSortedTasksAsXML` (collection size was 6, expected 5).
- `integration` **R3** `TaskFhirResourceIntegrationTest`: the same four — so the R3 provider this PR states is "intentionally unchanged" is broken collaterally.

(`FhirTaskDaoImplTest` itself stays green, and the two new focus tests pass.)

**Recommended fix:** move these two rows into a dedicated dataset (e.g. `FhirTaskDaoImplTest_focus_data.xml`) and `executeDataSet` it only inside `shouldSearchTasksByFocusReferenceAsJson`, mirroring the existing `FhirTaskDaoImplTest_owner_data.xml` / `TASK_DATA_OWNER_XML` pattern. That keeps the shared counts — and the R3 tests — untouched, consistent with the PR's R4-only scope.

**Alternative** (if you'd rather keep it shared): bump all nine assertions — `api` line 413 → `8` (and fix its comment), and the four R4 + four R3 `?status=requested` count/sorted tests → `6`. Note this pulls the R3 integration test file into the PR.

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceIntegrationTest.java
```diff
@@ -790,6 +793,57 @@ public void shouldReturnCountForTaskAsXml() throws Exception {
...
+		Task retrievedTask = readResponse(getResponse);
+
+		assertThat(retrievedTask, notNullValue());
+		assertThat(retrievedTask.hasFocus(), is(true));
+		assertThat(retrievedTask.getFocus().getReference(), equalTo("Observation/" + FOCUS_OBSERVATION_UUID));
+	}
+	
+	@Test
+	public void shouldSearchTasksByFocusReferenceAsJson() throws Exception {
+		//given — a task with focus reference pre-seeded in the test dataset
```
dkayiwa: question (non-blocking): This test seeds a bare-UUID reference instead of creating the task via the API and then searching — which sidesteps a real limitation worth confirming. `ReferenceTranslatorImpl.toOpenmrsType` stores `Reference.getReference()` verbatim, so a client POSTing `focus: { reference: "Observation/<uuid>" }` persists the full `"Observation/<uuid>"` string, while `handleReference` matches on the bare `param.getIdPart()`. So a Task created through the FHIR API with a standard relative reference would **not** be found by `GET /Task?focus=Observation/<uuid>`.

This is not a focus regression — `basedOn` / `owner` / `for` go through the exact same `ReferenceTranslatorImpl` + `handleReference` path — but since the seeded-data test doesn't exercise the create-via-API → search round-trip, is making that path work in scope for this feature, or are we deferring it as the pre-existing module-wide behavior?

dkayiwa (commented):
Re-reviewed at the latest head (`09f3daec`); the items from my earlier round are addressed.

**Disposition: needs work.** The most recent commit ("use pre-seeded test data for the focus search integration test") adds a focus task to the **shared** test fixture, which breaks 9 hard-coded count/size assertions across the `api` suite and the R3 + R4 integration suites — so CI (still running) will fail. All nine failures trace to that one shared-fixture change; see finding #1 (inline on the dataset) for the exact assertions and the recommended fix. Finding #2 is a non-blocking question.

**Needs action before merge:**
1. Keep the seeded focus task from inflating the shared `REQUESTED` counts (inline on `FhirTaskDaoImplTest_initial_data.xml`) — I'd recommend isolating the seed data into a dedicated dataset, à la `TASK_DATA_OWNER_XML`.

test-data/src/main/resources/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImplTest_initial_data.xml
vvkpd: Fixed — removed task_id=11 and reference_id=7 from the shared fixture entirely. The DAO focus tests are fully programmatic: they call `dao.createOrUpdate()` to set the focus reference on an existing task, so no dedicated XML dataset is needed either. REQUESTED counts are back to 5 across all test suites. Commit: aa24af3c.

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceIntegrationTest.java
vvkpd: Confirmed and fixed. The root cause was that `handleReference` was comparing against the `reference` column (which `ReferenceTranslatorImpl` stores as `"Observation/UUID"`) while `param.getIdPart()` returns a bare UUID — so they never matched. Fixed in commit 19353cac by switching the predicate to query `targetUuid` instead. `FhirReference.setReference()` already auto-extracts the UUID from `"Type/UUID"` strings into `targetUuid` via `FhirUtils.referenceToId()`, so a task created via the FHIR API with `focus: { reference: "Observation/UUID" }` now correctly persists `targetUuid = UUID` and the search finds it. The integration test is updated in commit aa24af3c to create via API first and then search, confirming the end-to-end path works.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
```diff
@@ -97,35 +102,35 @@ private <U> Optional<Predicate> handleStatus(OpenmrsFhirCriteriaContext<FhirTask
...
 	        ReferenceAndListParam reference, String property, String alias) {
 		return handleAndListParam(criteriaContext.getCriteriaBuilder(), reference, param -> {
 			if (validReferenceParam(param)) {
 				Join<?, ?> taskAliasJoin = criteriaContext.addJoin(property, alias);
-				
+
 				List<Optional<? extends Predicate>> predicateList = new ArrayList<>();
 				predicateList.add(Optional
-				        .of(criteriaContext.getCriteriaBuilder().equal(taskAliasJoin.get("reference"), param.getIdPart())));
+				        .of(criteriaContext.getCriteriaBuilder().equal(taskAliasJoin.get("targetUuid"), param.getIdPart())));
```
dkayiwa, line 120: question (non-blocking): this `reference` -> `targetUuid` switch is the right fix — for a FHIR-conformant `Type/uuid` reference, `setReference()` populates `targetUuid` with the bare uuid that `param.getIdPart()` compares against, whereas the old `reference` column held the full `"Type/uuid"` string and never matched (so the create-via-API -> search path was effectively broken here, not only for focus).

The thing to confirm is scope: `handleReference` is shared by all four Task reference params, so this also changes `basedOn`/`owner`/`for` matching — and since R3 and R4 share this one `FhirTaskDaoImpl`, it changes R3 Task reference-search behavior too. That's broader than the PR's stated "R3 intentionally unchanged / R4-only scope." My recommendation is to keep it (it's a correctness improvement across all four params), but call out the `basedOn`/`owner`/`for` + R3 change in the PR/commit description so it lands on record rather than silently. Was shipping it in this PR intended?

One narrow edge worth knowing: a reference stored at runtime as a *bare* uuid (no `Type/` prefix) leaves `targetUuid` null — `referenceToId` only extracts when the value matches the `Type/id` pattern — so it would stop matching here, where the old column comparison did. Not a concern for conformant references; just noting it.

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceIntegrationTest.java
```diff
@@ -789,6 +792,64 @@ public void shouldReturnCountForTaskAsXml() throws Exception {
...
+		assertThat(response, isOk());
+		assertThat(response.getContentType(), startsWith(FhirMediaTypes.JSON.toString()));
+		assertThat(response.getContentAsString(), notNullValue());
+
+		Bundle results = readBundleResponse(response);
+
+		assertThat(results, notNullValue());
+		assertThat(results.getType(), equalTo(Bundle.BundleType.SEARCHSET));
+		assertThat(results.hasEntry(), is(true));
+		assertThat(results.getEntry(), hasSize(greaterThanOrEqualTo(1)));
```
dkayiwa: suggestion (non-blocking): `hasSize(greaterThanOrEqualTo(1))` would pass even if the `focus` filter were ignored entirely — with no focus predicate the search returns all (non-retired) tasks, which is still >= 1. So this test catches the "predicate matches nothing -> 0 results" failure (the bug the `targetUuid` fix addressed) but not a "focus silently dropped -> returns everything" regression. The sibling `TaskSearchQueryTest#searchForTasks_shouldReturnTasksByFor` pins it down with `size == 1` plus the specific returned id.

Recommend asserting the result set actually carries the searched focus rather than just being non-empty — e.g. that every returned entry's `Task.focus` reference equals `Observation/<FOCUS_OBSERVATION_UUID>` (robust even though `shouldCreateTaskWithFocusReferenceAndRetrieveItAsJson` seeds the same focus), or that the created task's id is among the results. The DAO test `searchForTasks_shouldReturnTasksByFocusReference` has the same inclusion-only gap (`not(empty())` + `hasItem`).

api/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceProviderTest.java
```diff
@@ -285,6 +294,29 @@ public void searchTasks_shouldAddRelatedServiceRequestResourceToResultListWhenIn
...
+	@Test
+	public void searchTasks_shouldReturnMatchingTasksByFocusReference() {
+		//given
+		List<Task> tasks = new ArrayList<>();
+		tasks.add(task);
+		
+		ReferenceAndListParam focusReference = new ReferenceAndListParam()
+		        .addAnd(new ReferenceOrListParam().add(new ReferenceParam(FhirConstants.OBSERVATION, null, OBSERVATION_UUID)));
+		
+		when(taskService.searchForTasks(any())).thenReturn(new MockIBundleProvider<>(tasks, PREFERRED_PAGE_SIZE, COUNT));
```
dkayiwa, line 308: suggestion (non-blocking): stubbing `searchForTasks(any())` means this test passes whether or not the provider actually forwards `focusReference` into `TaskSearchParams` — it would still pass if `focus` were dropped from the builder. The include-tests just above stub on an explicit `new TaskSearchParams(...)` matcher, which does pin that mapping down.

Recommend either matching on a `TaskSearchParams` built with `focusReference` set, or capturing it with `ArgumentCaptor<TaskSearchParams>` and asserting `captured.getFocusReference()` equals the param passed in. (Tightening #2 would also cover this end-to-end, but the unit test is the faster, more targeted signal.)

dkayiwa (commented):
**Re-review at head `aa24af3c` — merge.**

The round-2 blocker is resolved: removing the focus task from the shared `FhirTaskDaoImplTest_initial_data.xml` puts the fixture back to 5 non-retired `REQUESTED` tasks, and CI is green on this head across all four JDK builds — including the R3/R4 integration suites the earlier fixture change had broken. Nothing below blocks merge; they're test-rigor tightenings plus one scope question.

- **#1 (question, inline on `FhirTaskDaoImpl`)** — the `reference` -> `targetUuid` switch lives in the shared `handleReference`, so it also changes `basedOn`/`owner`/`for` matching, including for the R3 provider (shared DAO) — broader than the stated "R3 intentionally unchanged / R4-only scope." It's the right fix; just confirm intent and document it.
- **#2 (suggestion, inline on the R4 integration + DAO focus-search tests)** — the assertions are inclusion-only (`>= 1` / `hasItem`), so they'd pass even if the focus filter were a no-op returning every task.
- **#3 (suggestion, inline on the R4 provider test)** — `searchForTasks(any())` doesn't verify `focusReference` is actually forwarded into the search params.

Needs action before merge: none.

dkayiwa:
In addition to the above comments, you can also look into the merge conflicts.

vvkpd:
@dkayiwa — all three suggestions from your latest round have been addressed:

1. PR description updated to document that the `handleReference` `reference` → `targetUuid` fix is a shared DAO change affecting all four Task reference params (`basedOn`/`owner`/`for`/`focus`) and both R3 and R4.
2. Focus search assertions tightened: DAO test now asserts `hasSize(1)` instead of `not(empty())`; integration test asserts the created task ID appears in results.
3. Provider test uses `ArgumentCaptor` to verify `focusReference` is actually forwarded into `TaskSearchParams`.

Could you please take another look when you get a chance? Thanks!

dkayiwa (commented):
**merge.** Nothing blocking; one non-blocking test-rigor note left inline on the focus-search integration test.

Needs action before merge: none.

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceIntegrationTest.java
dkayiwa: suggestion (non-blocking): the `hasItem(createdTask)` assertion added here still passes even if focus filtering is a no-op — with no focus predicate the search returns all non-retired tasks, and the just-created task is among them, so `hasSize(>= 1)` + `hasItem` both hold regardless. So this test still verifies wiring (the param reaches the provider and a bundle comes back), not discrimination.

Not blocking: the DAO test's `hasSize(1)` against the multi-task fixture now covers discrimination directly. If you'd want this integration test to guard the focus filter on its own, assert that a task with a different (or absent) focus is *not* in the results, or assert an exact size.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
```diff
@@ -33,7 +33,7 @@
 
 @Component
 public class FhirTaskDaoImpl extends BaseFhirDao<FhirTask> implements FhirTaskDao {
-	
+
```
dkayiwa: nit: this blank line — and ~12 others in this file — had its indentation tab stripped to empty, which diverges from every sibling DAO and from this file on `master`. The OpenMRS Eclipse formatter sets `indent_empty_lines=true`, so `spotless:apply` re-adds them — I ran it on this head and all 13 revert exactly to the `master` state. It won't fail CI (the build binds `spotless:apply`, not `check`), but it's churn the formatter undoes and will keep reappearing. Running `spotless:apply` and committing would drop these and keep the diff to the `Task.focus` change.

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceIntegrationTest.java
vvkpd: Fixed — changed to `hasSize(1)` so the assertion fails if the focus filter is a no-op returning all tasks. Since `FOCUS_OBSERVATION_UUID` is unique and no pre-seeded task carries it, exactly one result is expected.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
vvkpd: Fixed — ran `mvn spotless:apply` on the `api` module and committed the `FhirTaskDaoImpl.java` blank-line tab restorations.

vvkpd: Updated — also ran `spotless:apply` scoped to only the PR-modified files (using `-DspotlessFiles` regex). Two integration test files (R3 + R4 `TaskFhirResourceIntegrationTest`) had additional blank-line tab stripping and are now fixed too. The other 30+ files that a full-repo spotless run touches are left out to keep the diff focused.

PR #610, opened 2026-06-09
FM2-683 | Extend Fhir Task to have Draft and In Progress Task Statuses

PoojaSastry-TW:
JIRA card: https://openmrs.atlassian.net/browse/FM2-683

Description of what I changed:
- Added the existing R4 Task statuses: DRAFT and IN PROGRESS to the TaskStatus enum of the FhirTask class.
- Added relevant tests.

Please review: @ibacher @dkayiwa

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
@@ -21,6 +21,7 @@ import java.util.Optional;
 import ca.uhn.fhir.rest.param.ReferenceAndListParam;
 import ca.uhn.fhir.rest.param.ReferenceParam;
 import ca.uhn.fhir.rest.param.TokenAndListParam;
+import org.hl7.fhir.r4.model.Task;
 import org.openmrs.api.db.DAOException;
 import org.openmrs.module.fhir2.FhirConstants;
 import org.openmrs.module.fhir2.api.dao.FhirTaskDao;
@@ -91,7 +92,7 @@ public class FhirTaskDaoImpl extends BaseFhirDao<FhirTask> implements FhirTaskDa
 			if (token.getValue() != null) {
 				try {
 					return Optional.of(criteriaContext.getCriteriaBuilder().equal(criteriaContext.getRoot().get("status"),
-					    FhirTask.TaskStatus.valueOf(token.getValue().toUpperCase())));
+					    FhirTask.TaskStatus.valueOf(Task.TaskStatus.fromCode(token.getValue()).name())));
 				}
 				catch (IllegalArgumentException e) {
 					return Optional.empty();
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
@@ -21,6 +21,7 @@ import java.util.Optional;
 import ca.uhn.fhir.rest.param.ReferenceAndListParam;
 import ca.uhn.fhir.rest.param.ReferenceParam;
 import ca.uhn.fhir.rest.param.TokenAndListParam;
+import org.hl7.fhir.exceptions.FHIRException;
 import org.hl7.fhir.r4.model.Task;
 import org.openmrs.api.db.DAOException;
 import org.openmrs.module.fhir2.FhirConstants;
@@ -92,9 +93,9 @@ public class FhirTaskDaoImpl extends BaseFhirDao<FhirTask> implements FhirTaskDa
 			if (token.getValue() != null) {
 				try {
 					return Optional.of(criteriaContext.getCriteriaBuilder().equal(criteriaContext.getRoot().get("status"),
-					    FhirTask.TaskStatus.valueOf(Task.TaskStatus.fromCode(token.getValue()).name())));
+					    FhirTask.TaskStatus.valueOf(Task.TaskStatus.fromCode(token.getValue().toLowerCase()).name())));
 				}
-				catch (IllegalArgumentException e) {
+				catch (IllegalArgumentException | FHIRException e) {
 					return Optional.empty();
 				}
 			}
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
@@ -90,7 +90,7 @@ public class FhirTaskDaoImpl extends BaseFhirDao<FhirTask> implements FhirTaskDa
 	        TokenAndListParam tokenAndListParam) {
 		
 		return handleAndListParam(criteriaContext.getCriteriaBuilder(), tokenAndListParam, token -> {
-			if (token.getValue() != null) {
+			if (token.getValue() != null && !token.getValue().isEmpty()) {
 				try {
 					return Optional.of(criteriaContext.getCriteriaBuilder().equal(criteriaContext.getRoot().get("status"),
 					    FhirTask.TaskStatus.valueOf(Task.TaskStatus.fromCode(token.getValue().toLowerCase()).name())));
--- a/api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java
@@ -44,8 +44,10 @@ public class FhirTask extends BaseOpenmrsMetadata {
 	 * Based on <a href="https://www.hl7.org/fhir/task.html">...</a> v4.0.1
 	 */
 	public enum TaskStatus {
+		DRAFT,
 		REQUESTED,
 		READY,
+		INPROGRESS,
 		ONHOLD,
 		CANCELLED,
 		REJECTED,
--- a/api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java
@@ -46,13 +46,16 @@ public class FhirTask extends BaseOpenmrsMetadata {
 	public enum TaskStatus {
 		DRAFT,
 		REQUESTED,
+		RECEIVED,
+		ACCEPTED,
+		REJECTED,
 		READY,
+		CANCELLED,
 		INPROGRESS,
 		ONHOLD,
-		CANCELLED,
-		REJECTED,
-		ACCEPTED,
+		FAILED,
 		COMPLETED,
+		ENTEREDINERROR,
 		UNKNOWN
 	}
 	
--- a/api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImplTest.java
+++ b/api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImplTest.java
@@ -87,16 +87,24 @@ public class TaskTranslatorImplTest {
 	private static final FhirTask.TaskStatus OPENMRS_NEW_TASK_STATUS = FhirTask.TaskStatus.ACCEPTED;
 	
 	private static final Task.TaskStatus FHIR_TASK_STATUS_READY = Task.TaskStatus.READY;
-	
+
 	private static final Task.TaskStatus FHIR_TASK_STATUS_ONHOLD = Task.TaskStatus.ONHOLD;
-	
+
 	private static final Task.TaskStatus FHIR_TASK_STATUS_CANCELLED = Task.TaskStatus.CANCELLED;
-	
+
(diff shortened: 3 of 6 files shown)


api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java
```diff
@@ -44,8 +44,10 @@ public class FhirTask extends BaseOpenmrsMetadata {
 	 * Based on <a href="https://www.hl7.org/fhir/task.html">...</a> v4.0.1
 	 */
 	public enum TaskStatus {
+		DRAFT,
```
dkayiwa, line 47: 👍 for matching the HAPI enum names exactly (`DRAFT`, `INPROGRESS` rather than `IN_PROGRESS`) — both directions of `TaskTranslatorImpl` translate via `valueOf(<name>)`, so the name-mirroring is load-bearing.

One scope question: R4 still has three statuses we don't cover — `received`, `failed` and `entered-in-error`. Any of them arriving on a POST/PUT gets coerced to `UNKNOWN` in `TaskTranslatorImpl#setOpenmrsTaskFields`, and since `org.hl7.fhir.r4.model.Task.TaskStatus` has no `UNKNOWN` constant, every subsequent read of such a task throws `IllegalArgumentException: No enum constant ...Task.TaskStatus.UNKNOWN` (an HTTP 500, and it poisons any search bundle that includes the task). `UNKNOWN` is also the liquibase default for this column. Since this PR establishes the pattern and the javadoc above says this enum is based on the v4.0.1 status list, could we add the remaining three here as well? If you'd rather keep the ticket scope as-is, let's open a follow-up ticket so this doesn't stay half-defused.

dkayiwa (commented):
Thanks @PoojaSastry-TW! The core change is correct and safe: `status` is string-mapped (`@Enumerated(EnumType.STRING)`) so inserting enum constants can't disturb existing rows, the new constants mirror the HAPI R4 names exactly, the R3/DSTU3 `TaskVersionConverter` already maps both statuses, and `TaskTranslatorImplTest` passes locally (61/61).

Inline comments cover three things:
1. Formatter-reverted whitespace churn in the test file (please re-run spotless and re-commit).
2. `GET /Task?status=in-progress` currently fails open in `FhirTaskDaoImpl#handleStatus` — worth fixing here or linking a follow-up before merge, since search is the main way clients will consume this status.
3. A scope question about adding the remaining R4 statuses (`received`, `failed`, `entered-in-error`) to defuse the `UNKNOWN` round-trip crash.

api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java
```diff
@@ -44,8 +44,10 @@ public class FhirTask extends BaseOpenmrsMetadata {
...
 		REQUESTED,
 		READY,
+		INPROGRESS,
```
dkayiwa, line 54: Heads-up: adding the enum constant doesn't yet make `GET /Task?status=in-progress` work. `FhirTaskDaoImpl#handleStatus` resolves the token with `FhirTask.TaskStatus.valueOf(token.getValue().toUpperCase())`, and the wire code for this status is `in-progress` (`Task.TaskStatus.INPROGRESS.toCode()`), which uppercases to `IN-PROGRESS` and throws. The token is then swallowed (`Optional.empty()`), and `BaseDao#handleAndListParam` builds `cb.and()` over an empty predicate array — which is always true — so the search silently returns **all** tasks instead of filtering. Same pre-existing problem for `on-hold`, but `in-progress` is the status queue-style clients will actually poll for, so it really bites here.

The existing search tests never catch this because they pass enum `toString()` values (`"ACCEPTED"` etc.), not wire codes. Suggested fix in `handleStatus`, keeping the existing catch block:

```java
FhirTask.TaskStatus.valueOf(Task.TaskStatus.fromCode(token.getValue()).name())
```

plus a `TaskSearchQueryTest` case that passes the literal `"in-progress"` token. Happy for that to be a follow-up FM2 ticket instead if you want to keep this PR model-only — but then let's link it before merging.

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImplTest.java
```diff
@@ -87,16 +87,24 @@ public class TaskTranslatorImplTest {
 	private static final FhirTask.TaskStatus OPENMRS_NEW_TASK_STATUS = FhirTask.TaskStatus.ACCEPTED;
 	
 	private static final Task.TaskStatus FHIR_TASK_STATUS_READY = Task.TaskStatus.READY;
-	
+
```
dkayiwa: This and the ~18 similar blank-line changes in this file get reverted as soon as anyone builds: spotless (OpenMRS Eclipse formatter) runs at `process-sources` and re-adds the leading tabs — building this branch locally immediately leaves the file modified again in `git status`. Could you run `mvn spotless:apply` and re-commit so the diff only contains the real changes? Otherwise master ends up in non-canonical format and the next PR that touches this file inherits the churn.

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImplTest.java
```diff
@@ -253,26 +261,38 @@ public void toFhirResource_shouldTranslateStatus() {
 		
 		task.setStatus(OPENMRS_TASK_STATUS_CANCELLED);
 		assertThat(taskTranslator.toFhirResource(task).getStatus(), equalTo(FHIR_TASK_STATUS_CANCELLED));
+
+		task.setStatus(OPENMRS_TASK_STATUS_DRAFT);
+		assertThat(taskTranslator.toFhirResource(task).getStatus(), equalTo(FHIR_TASK_STATUS_DRAFT));
+
+		task.setStatus(OPENMRS_TASK_STATUS_INPROGRESS);
+		assertThat(taskTranslator.toFhirResource(task).getStatus(), equalTo(FHIR_TASK_STATUS_INPROGRESS));
```
dkayiwa, line 281: Translator coverage in both directions looks good (and passes locally). Could we also get DAO/search-level coverage? The shared dataset (`test-data/.../FhirTaskDaoImplTest_initial_data.xml`) only has `ACCEPTED`/`REJECTED`/`REQUESTED` rows, so neither new status is exercised through `TaskSearchQueryTest`. An `INPROGRESS` row plus a search test using the wire-format token `"in-progress"` would cover this end-to-end — and would catch the `handleStatus` issue flagged on `FhirTask.java`.

api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java
PoojaSastry-TW: In the new commit, I have added the remaining three R4 statuses as well. 

Also added a comment about the scope extension in the current JIRA card: https://openmrs.atlassian.net/browse/FM2-683.

api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java
PoojaSastry-TW: The wire-code search tests for the three hyphenated statuses: in-progress, on-hold and entered-in-error have been included in the latest commit.

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImplTest.java
PoojaSastry-TW: Addressed in the new commit.

api/src/test/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImplTest.java
PoojaSastry-TW: Ran mvn spotless:apply before pushing the new commit, please check.

dkayiwa (commented):
**Merge after one fix.** The only blocker is a new, reachable HTTP 500 introduced by the wire-code search fix: `GET /Task?status=` (empty status value) throws an uncaught `NullPointerException` in `handleStatus`, where the pre-PR code returned all tasks. Details and a one-line guard are inline on `FhirTaskDaoImpl.java`.

Needs action before merge:
- Guard the empty status value in `handleStatus` (inline `issue (blocking)` + suggestion).

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
```diff
@@ -91,9 +93,9 @@ private <U> Optional<Predicate> handleStatus(OpenmrsFhirCriteriaContext<FhirTask
 			if (token.getValue() != null) {
```
dkayiwa: **issue (blocking):** `GET /Task?status=` (empty status value) returns HTTP 500 with this change, where the pre-PR code returned all tasks.

When the value is empty, line 96 calls `Task.TaskStatus.fromCode("")`, which returns `null` — HAPI returns `null` for null/empty input and only throws `FHIRException` for *non-empty* unknown codes (verified against `org.hl7.fhir.r4.model.Task` 5.6.27). `null.name()` is then an uncaught `NullPointerException`; the `catch (IllegalArgumentException | FHIRException)` doesn't cover it. The previous `valueOf("".toUpperCase())` threw `IllegalArgumentException`, which *was* caught, so an empty status used to fail open.

I confirmed this end-to-end: a request to `/Task?status=` binds to a single token with value `""`, and running `handleStatus` with that token throws `NullPointerException: ...Task$TaskStatus.fromCode(String) is null`. It also contradicts the new `searchForTasks_shouldReturnAllTasksForUnrecognizedStatus` test — an empty value is the one unrecognized input that doesn't fail open.

Guarding the empty value keeps the fail-open contract:

```suggestion
			if (token.getValue() != null && !token.getValue().isEmpty()) {
```

dkayiwa (commented):
**Merge.** The sole blocker from my previous review is resolved: `GET /Task?status=` (empty value) no longer throws an uncaught `NullPointerException` in `FhirTaskDaoImpl#handleStatus` — the `&& !token.getValue().isEmpty()` guard restores the prior fail-open behavior, now locked in by a regression test in both `FhirTaskDaoImplTest` and `TaskSearchQueryTest` (either would error without the guard). I don't see anything new to block on in the merged result.

Needs action before merge: **none.** I've left one non-blocking, tracking-only follow-up on the existing `UNKNOWN` thread — capture the residual `UNKNOWN`-on-read 500 as its own ticket before this card closes.

api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java
dkayiwa: Thanks @PoojaSastry-TW — adding the three statuses resolves the POST/PUT path here: `received`/`failed`/`entered-in-error` now map 1:1 instead of collapsing to `UNKNOWN`. ✅

**suggestion (non-blocking, follow-up):** the residual half is still open. A task that is *already* `UNKNOWN` — the liquibase default for `fhir_task.status` (`api/src/main/resources/liquibase.xml:133`) — still 500s on read, because `TaskTranslatorImpl#setFhirTaskFields` does an unguarded `Task.TaskStatus.valueOf(openmrsTask.getStatus().name())` and HAPI's `org.hl7.fhir.r4.model.Task.TaskStatus` has no `UNKNOWN` constant (it's `...ENTEREDINERROR, NULL`), so `valueOf("UNKNOWN")` throws `IllegalArgumentException`. That's out of scope for this PR and not a merge blocker. But since FM2-683 closes when this merges, could we capture it as its own FM2 ticket rather than a note on this card, so the half-defused bit doesn't get lost?

PoojaSastry-TW:
> **Merge.** The sole blocker from my previous review is resolved: `GET /Task?status=` (empty value) no longer throws an uncaught `NullPointerException` in `FhirTaskDaoImpl#handleStatus` — the `&& !token.getValue().isEmpty()` guard restores the prior fail-open behavior, now locked in by a regression test in both `FhirTaskDaoImplTest` and `TaskSearchQueryTest` (either would error without the guard). I don't see anything new to block on in the merged result.
> 
> Needs action before merge: **none.** I've left one non-blocking, tracking-only follow-up on the existing `UNKNOWN` thread — capture the residual `UNKNOWN`-on-read 500 as its own ticket before this card closes.



> **Merge.** The sole blocker from my previous review is resolved: `GET /Task?status=` (empty value) no longer throws an uncaught `NullPointerException` in `FhirTaskDaoImpl#handleStatus` — the `&& !token.getValue().isEmpty()` guard restores the prior fail-open behavior, now locked in by a regression test in both `FhirTaskDaoImplTest` and `TaskSearchQueryTest` (either would error without the guard). I don't see anything new to block on in the merged result.
> 
> Needs action before merge: **none.** I've left one non-blocking, tracking-only follow-up on the existing `UNKNOWN` thread — capture the residual `UNKNOWN`-on-read 500 as its own ticket before this card closes.

Hi Daniel @dkayiwa,

Here's the newly created ticket for the 500 error: https://openmrs.atlassian.net/browse/FM2-684. Please let me know if this needs any modification. 
Thanks for your support and guidance. 

Pooja

dkayiwa:
Thank you so much @PoojaSastry-TW 👍

PR #615, opened 2026-07-06
TRUNK-6674: Translate drug order duration units for the current SNOMED CT minute code and UCUM mappings

dkayiwa:
## Description

Companion to openmrs/openmrs-core#6246 (TRUNK-6674). SNOMED CT inactivated the legacy minute code `258701004` in its 2021-07-31 release, and dictionaries that track current SNOMED CT releases, such as CIEL, now map their Minutes concepts to the active code `1156209001` (current CIEL carries both as SAME-AS). Core is learning to resolve these; fhir2's `DurationUnitTranslatorImpl` only knew the legacy SNOMED codes, so:

- **Outbound**, a drug order whose duration units concept resolves to the new code translated to `Timing.UnitsOfTime.NULL`, silently dropping the duration unit from the MedicationRequest. In addition, `Duration.getCode(...)` returns whichever SNOMED SAME-AS mapping iterates first, so with current CIEL (both codes present) the outcome depended on mapping-set iteration order.
- **Inbound**, `toOpenmrsType` iterated a `HashMap` and looked up exactly one code per unit, so `MIN` only resolved for dictionaries carrying the legacy mapping.

## Changes

- `toFhirResource` now resolves the concept's SAME-AS mappings against source-scoped code maps: SNOMED CT (the legacy codes plus `1156209001`) and UCUM (`s`, `min`, `h`, `d`, `wk`, `mo`, `a`; the `UnitsOfTime` codes are the UCUM codes). Per review, SNOMED CT mappings take priority, stopping at the first known code, and UCUM is consulted only when no SNOMED CT mapping resolves, so concepts that translated while UCUM was ignored keep translating identically and UCUM strictly fills gaps. Concepts with no known mapping still translate to `NULL`.
- `toOpenmrsType` tries every code that denotes the requested unit: legacy codes first, so dictionaries carrying only legacy mappings resolve exactly as before, then the current SNOMED CT minute code, then the UCUM mapping (the UCUM source is matched by name or HL7 code, since real dictionaries register it by name).
- The new minute code is a string literal with a TODO, because the module compiles against platform 2.8.0, which predates `Duration.SNOMED_CT_MINUTES_CODE_2021`.
- New `DurationUnitTranslatorImplCurrentCodesTest` runs against its own dataset that deliberately carries no legacy mappings, proving the fallbacks in both directions (new-code-only Minutes, UCUM-only Hours). In-memory concepts cover the shape current CIEL actually ships, one concept carrying both minute codes, a concept whose UCUM mapping disagrees with its SNOMED CT mapping (SNOMED CT wins), and the UCUM fallback when a SNOMED CT mapping carries only an unknown code. The existing `DurationUnitTranslatorImplTest` is untouched and keeps proving the legacy path.

## Ticket

Driven by [TRUNK-6674](https://openmrs.atlassian.net/browse/TRUNK-6674). There is no FM2 ticket for this yet; happy to retitle once one is filed.

[TRUNK-6674]: https://openmrs.atlassian.net/browse/TRUNK-6674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/DurationUnitTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/DurationUnitTranslatorImpl.java
@@ -20,6 +20,7 @@ import java.util.Map;
 
 import lombok.Getter;
 import lombok.Setter;
+import lombok.extern.slf4j.Slf4j;
 import org.hl7.fhir.r4.model.Timing;
 import org.openmrs.Concept;
 import org.openmrs.ConceptMap;
@@ -32,6 +33,7 @@ import org.openmrs.module.fhir2.api.translators.DurationUnitTranslator;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
+@Slf4j
 @Component
 public class DurationUnitTranslatorImpl implements DurationUnitTranslator {
 	
@@ -86,17 +88,28 @@ public class DurationUnitTranslatorImpl implements DurationUnitTranslator {
 	
 	@Override
 	public Timing.UnitsOfTime toFhirResource(@Nonnull Concept concept) {
+		// mirrors Duration.getDuration in TRUNK-6674: a concept whose known codes denote different
+		// units is miscurated, so refuse to pick one rather than serialize an arbitrary unit
+		Timing.UnitsOfTime knownUnit = null;
 		for (ConceptMap conceptMapping : concept.getConceptMappings()) {
 			if (!ConceptMapType.SAME_AS_MAP_TYPE_UUID.equals(conceptMapping.getConceptMapType().getUuid())) {
 				continue;
 			}
 			ConceptReferenceTerm term = conceptMapping.getConceptReferenceTerm();
 			Timing.UnitsOfTime unitsOfTime = findUnitsOfTime(term.getConceptSource(), term.getCode());
-			if (unitsOfTime != null) {
-				return unitsOfTime;
+			if (unitsOfTime == null) {
+				continue;
+			}
+			if (knownUnit == null) {
+				knownUnit = unitsOfTime;
+			} else if (knownUnit != unitsOfTime) {
+				log.warn(
+				    "Not translating duration units concept {} because its SAME-AS mappings denote different units, e.g. {} and {}",
+				    concept.getUuid(), knownUnit, unitsOfTime);
+				return Timing.UnitsOfTime.NULL;
 			}
 		}
-		return Timing.UnitsOfTime.NULL;
+		return knownUnit != null ? knownUnit : Timing.UnitsOfTime.NULL;
 	}
 	
 	@Override
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/DurationUnitTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/DurationUnitTranslatorImpl.java
@@ -105,6 +105,9 @@ public class DurationUnitTranslatorImpl implements DurationUnitTranslator {
 			if (entry.getValue() == unitsOfTime) {
 				Concept concept = conceptService.getConceptByMapping(entry.getKey(),
 				    Duration.SNOMED_CT_CONCEPT_SOURCE_HL7_CODE);
+				if (concept == null) {
+					concept = conceptService.getConceptByMapping(entry.getKey(), SNOMED_CT_CONCEPT_SOURCE_NAME);
+				}
 				if (concept != null) {
 					return concept;
 				}
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/DurationUnitTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/DurationUnitTranslatorImpl.java
@@ -49,6 +49,11 @@ public class DurationUnitTranslatorImpl implements DurationUnitTranslator {
 	
 	private static final String UCUM_CONCEPT_SOURCE = "UCUM";
 	
+	// Matched as a fallback for dictionaries that register SNOMED CT by name without setting its HL7
+	// code, mirroring UCUM. TODO: replace with Duration.SNOMED_CT_CONCEPT_SOURCE_NAME once the module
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/DurationUnitTranslatorImpl.java
```diff
@@ -34,36 +39,86 @@ public class DurationUnitTranslatorImpl implements DurationUnitTranslator {
...
 			}
 		}
+		if (UCUM_CODE_MAP.containsKey(unitsOfTime.toCode())) {
+			return conceptService.getConceptByMapping(unitsOfTime.toCode(), UCUM_CONCEPT_SOURCE);
+		}
+		return null;
+	}
+	
+	private static Timing.UnitsOfTime findUnitsOfTime(ConceptSource conceptSource, String code) {
+		if (Duration.SNOMED_CT_CONCEPT_SOURCE_HL7_CODE.equals(conceptSource.getHl7Code())) {
```
mseaton: Not sure why we don't check the source name here, but we do for ucum?

mseaton (commented):
Generally LGTM, one comment

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/DurationUnitTranslatorImpl.java
dkayiwa: Same root cause as the core `Duration` change: SNOMED CT reliably carries the `SCT` HL7 code while UCUM is often registered by name only, so only UCUM had the name check. I'll mirror the core fix here — check the SNOMED source name in addition to the HL7 code, consistent with the UCUM branch.

NethmiRodrigo:
@wikumChamith could you take a look at this?

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/DurationUnitTranslatorImpl.java
dkayiwa: Done. 9cb6921 matched the source by name on the way out, and b9c74a9b completes the round trip: toOpenmrsType now falls back to the SNOMED CT source name when the SCT HL7 code lookup finds nothing, and both directions are covered by tests against a dataset whose SNOMED CT source has no HL7 code.

dkayiwa:
@claude review

ibacher (approved):

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/DurationUnitTranslatorImpl.java
```diff
@@ -34,36 +39,95 @@ public class DurationUnitTranslatorImpl implements DurationUnitTranslator {
...
-		if (unitsOfTime == null) {
-			return Timing.UnitsOfTime.NULL;
+		for (ConceptMap conceptMapping : concept.getConceptMappings()) {
+			if (!ConceptMapType.SAME_AS_MAP_TYPE_UUID.equals(conceptMapping.getConceptMapType().getUuid())) {
+				continue;
+			}
+			ConceptReferenceTerm term = conceptMapping.getConceptReferenceTerm();
+			Timing.UnitsOfTime unitsOfTime = findUnitsOfTime(term.getConceptSource(), term.getCode());
+			if (unitsOfTime != null) {
+				return unitsOfTime;
```
dkayiwa: This loop diverges from how the core PR resolves the same mappings, and I think we should decide that on purpose rather than by accident. Core's `Duration.getDuration` scans every SAME-AS mapping and refuses to resolve when two known codes denote different units (returns null and logs a warning). Here the first known code wins, and since `getConceptMappings()` is a `HashSet`, "first" is arbitrary: a miscurated concept mapped SAME-AS to both a SNOMED minutes code and UCUM `h` would serialize as minutes on one deployment and hours on another, silently. The old code only ever consulted mappings whose source carries the SCT HL7 code, so a cross-source conflict like that could not surface before.

My preference is to mirror core: remember the first known unit, keep scanning, and return `UnitsOfTime.NULL` with a warning when a later known code disagrees. If we do that, it also becomes worth adding a test with a concept carrying both the legacy and the 2021 minute code (the shape current CIEL actually ships); it resolves identically either way, but it pins the main real-world scenario.

Not blocking: on a well-curated dictionary the two behaviors are identical.

ibacher: I'm somewhat agnostic on this one. Mirroring core makes sense for ecosystem consistency. On the other hand, I actually think the (proposed) core behaviour is potentially a mistake, since it enforces a consistently rule we don't otherwise require (i.e., if a concept is mapped SAME-AS a particular SNOMED CT code than it must either have no UCUM SAME-AS mapping or the SAME-AS mapping must resolve to the same value). Since previously we ignored UCUM mappings, this introduces a behaviour change that could potentially invalidate existing valid data.

The option that appeals to me is to iterate through only the SNOMED CT codes, stop on the first match, and then if there's no match, iterate through the UCUM codes. While less algorithmically efficient, this would ensure that everything that works today continues to work with no behaviour change and we support edge-cases where the code is defined only using a UCUM mapping.

dkayiwa: Went with your approach in 788657b6. SNOMED CT SAME-AS mappings are resolved first, stopping at the first known code, and UCUM is consulted only when no SNOMED CT mapping resolves, so everything that translated while UCUM was ignored keeps translating identically and UCUM strictly fills gaps. It also matches the priority toOpenmrsType already applies (SCT lookups first, UCUM as the last fallback). Tests cover both a concept whose UCUM mapping disagrees with its SNOMED CT mapping (SNOMED CT wins) and the fallback when the only SNOMED CT mapping carries a code we don't know.

You've convinced me on the core behaviour too: refusing on conflict there would bite harder than here, since getDuration feeds order validation and previously valid drug orders could start failing to save. I'll rework openmrs/openmrs-core#6246 to this same source-priority resolution.

ibacher (approved):

ibacher:
Thanks @dkayiwa!

PR #618, opened 2026-07-15
Require Orders privileges for FHIR Task resource access

dkayiwa:
### Summary
`FhirTaskDao` is an empty interface, and neither it nor the base `FhirDao` declares any `@Authorized` privilege, so `AuthorizationAdvice` performs no privilege check at all for FHIR Task operations. Any authenticated user, including an account with no clinical privileges, can read, search, create, update and delete every FHIR Task via `/ws/fhir2/R4/Task` (and the R3 path). Every other FHIR2 DAO (Allergy, Observation, ServiceRequest, MedicationRequest, ...) annotates its `get`/`getSearchResults`/`createOrUpdate`/`delete` methods; Task was the one gap.

### Change
Override the inherited `FhirDao` methods on `FhirTaskDao` with `@Authorized`, mirroring the annotated fleet:
- reads (`get`, `get(Collection)`, `getSearchResults`, `getSearchResultsCount`) require `Get Orders`
- `createOrUpdate` requires `Add Orders`/`Edit Orders`
- `delete` requires `Delete Orders`

### Privilege choice (please confirm)
FHIR Task has no dedicated OpenMRS privilege. It represents order-fulfillment and clinical-workflow state and references based-on orders, so I mapped it to the Orders privileges, identical to `FhirServiceRequestDao`. If your Task workflows (for example lab integrations such as OpenELIS) run under a service account, that account will now need the Orders privileges to read or write Task. If you would rather introduce dedicated Task privileges, that can replace the constants here without changing the shape of the fix.

### Tests
Added `get_shouldRequireGetOrdersPrivilege` and `getSearchResults_shouldRequireGetOrdersPrivilege` to `FhirTaskDaoImplTest`, obtaining the AOP-proxied DAO so the annotations are actually exercised. All Task api tests pass: DAO 21, service 7, search 33, r3/r4 resource providers 10/15, translators unchanged.

Addresses GHSA-vm98-x99m-w327.

### DIFF
--- a/api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImplTest.java
+++ b/api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImplTest.java
@@ -18,6 +18,8 @@ import static org.hamcrest.Matchers.hasProperty;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.notNullValue;
+import static org.junit.Assert.assertThrows;
+import static org.openmrs.util.PrivilegeConstants.GET_ORDERS;
 
 import java.time.LocalDateTime;
 import java.time.Month;
@@ -35,15 +37,19 @@ import org.hibernate.SessionFactory;
 import org.junit.Before;
 import org.junit.Test;
 import org.openmrs.Concept;
+import org.openmrs.api.APIAuthenticationException;
 import org.openmrs.api.ConceptService;
+import org.openmrs.api.context.Context;
 import org.openmrs.api.db.hibernate.HibernateConceptDAO;
 import org.openmrs.module.fhir2.BaseFhirContextSensitiveTest;
 import org.openmrs.module.fhir2.FhirConstants;
+import org.openmrs.module.fhir2.api.dao.FhirTaskDao;
 import org.openmrs.module.fhir2.api.search.param.SearchParameterMap;
 import org.openmrs.module.fhir2.model.FhirReference;
 import org.openmrs.module.fhir2.model.FhirTask;
 import org.openmrs.module.fhir2.model.FhirTaskInput;
 import org.openmrs.module.fhir2.model.FhirTaskOutput;
+import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.beans.factory.annotation.Autowired;
 
 public class FhirTaskDaoImplTest extends BaseFhirContextSensitiveTest {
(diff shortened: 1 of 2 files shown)


dkayiwa:
@claude review

api/src/test/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImplTest.java
```diff
@@ -475,4 +484,36 @@ public void searchForTasks_shouldReturnTasksByFocusReference() {
...
+		assertThrows(APIAuthenticationException.class, () -> authorizedDao.getSearchResults(theParams));
+		
+		try {
+			Context.addProxyPrivilege(GET_ORDERS);
+			assertThat(authorizedDao.getSearchResults(theParams), notNullValue());
+		}
+		finally {
+			Context.removeProxyPrivilege(GET_ORDERS);
+		}
+	}
```
dkayiwa, line 518: The write half of the fix has no test: createOrUpdate (Add Orders/Edit Orders) and delete (Delete Orders) are annotated but nothing exercises them. If those annotations are ever dropped or the advice stops resolving them, nothing throws. An unauthenticated caller quietly saves or soft-deletes Tasks and the suite stays green; that is exactly what master does today (I ran these probes there as a control, and delete just soft-deletes with nobody logged in). Enforcement does currently work on all six methods, including calls through the raw FhirDao type that BaseFhirService uses, so nothing blocks here. Two more tests in the same shape would lock in the write path:

```java
@Test
public void createOrUpdate_shouldRequireAddOrEditOrdersPrivilege() {
	FhirTaskDao authorizedDao = daoProvider.getObject();
	Context.logout();
	
	assertThrows(APIAuthenticationException.class, () -> authorizedDao.createOrUpdate(new FhirTask()));
}

@Test
public void delete_shouldRequireDeleteOrdersPrivilege() {
	FhirTaskDao authorizedDao = daoProvider.getObject();
	Context.logout();
	
	assertThrows(APIAuthenticationException.class, () -> authorizedDao.delete(TASK_UUID));
}
```

dkayiwa (commented):
Went over this again end to end; nothing further blocks it from my side. Two notes for reviewers.

On the privilege mapping (the confirm request in the description), one more data point beyond consistency with FhirServiceRequestDao: rendering a ServiceRequest already runs a Task search internally (ServiceRequestTranslatorImpl.determineServiceRequestPerformer calls taskService.searchForTasks to populate the performer), so a user who can read ServiceRequests has to be able to search Tasks regardless. Dedicated Task privileges would have broken ServiceRequest reads for anyone not granted them. If dedicated privileges are ever introduced later, that call site needs to be part of the migration.

Separately, for the record on scope: FhirOrderDao and FhirContactPointMapDao also carry no @Authorized annotations. Neither sits behind a resource provider (FhirOrderDao only resolves order references inside OrderReferenceTranslatorImpl, and contact point maps are internal telecom metadata), so there is no direct exposure like Task had, but if the goal is that every FhirDaoAop bean enforces privileges, those two are the remainder. Follow-up material, not this PR.

PR #620, opened 2026-08-04
FM2-686: Include Order Number and Accession Number reference in Medic…

mogoodrich:
…ationRequest and ServiceRequest

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-686

- [X] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

- [X] I have **added tests** to cover my changes. (If you refactored

- [X] I ran `mvn clean package` right before creating this pull request and

  No? -> execute above command

- [X] All new and existing **tests passed**.

- [X] My pull request is **based on the latest changes** of the master branch.

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/OrderIdentifierTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/OrderIdentifierTranslatorImpl.java
@@ -11,11 +11,7 @@ package org.openmrs.module.fhir2.api.translators.impl;
 
 import javax.annotation.Nonnull;
 
-import java.util.ArrayList;
-import java.util.List;
-
 import ca.uhn.fhir.rest.server.exceptions.InvalidRequestException;
-import org.apache.commons.lang3.StringUtils;
 import org.hl7.fhir.r4.model.CodeableConcept;
 import org.hl7.fhir.r4.model.Coding;
 import org.hl7.fhir.r4.model.Identifier;
@@ -41,28 +37,6 @@ public class OrderIdentifierTranslatorImpl implements OrderIdentifierTranslator 
 		orderIdentifier.setUse(Identifier.IdentifierUse.USUAL);
 		
 		return orderIdentifier;
-	}
-	
-	@Override
-	public List<Identifier> toFhirIdentifiers(@Nonnull Order order) {
-		List<Identifier> identifiers = new ArrayList<>();
-		identifiers.add(toFhirResource(order));
-		
-		if (StringUtils.isNotBlank(order.getAccessionNumber())) {
-			Identifier accessionIdentifier = new Identifier();
-			
-			Coding acsnCoding = new Coding().setSystem(FhirConstants.IDENTIFIER_TYPE_SYSTEM_URI).setCode("ACSN")
-			        .setDisplay("Accession ID");
-			accessionIdentifier.setType(new CodeableConcept().addCoding(acsnCoding));
-			
-			accessionIdentifier.setValue(order.getAccessionNumber());
-			
-			accessionIdentifier.setUse(Identifier.IdentifierUse.USUAL);
-			
-			identifiers.add(accessionIdentifier);
-		}
-		
-		return identifiers;
 	}
 	
 	@Override
--- a/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/OrderIdentifierTranslatorImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/OrderIdentifierTranslatorImpl.java
@@ -11,11 +11,16 @@ package org.openmrs.module.fhir2.api.translators.impl;
 
 import javax.annotation.Nonnull;
 
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/OrderIdentifierTranslatorImpl.java
```diff
@@ -27,7 +28,7 @@ public Identifier toFhirResource(@Nonnull Order order) {
 		
 		Identifier orderIdentifier = new Identifier();
 		
-		Coding placCoding = new Coding().setSystem("http://terminology.hl7.org/CodeSystem/v2-0203").setCode("PLAC")
+		Coding placCoding = new Coding().setSystem(FhirConstants.IDENTIFIER_TYPE_SYSTEM_URI).setCode("PLAC")
```
dkayiwa, line 31: Now that this Identifier is becoming the resource's own business identifier rather than just the payload of a `Reference.identifier`, the missing `system` starts to matter: a consumer reads `identifier[0].value` and gets a bare `ORD-1` with nothing saying whose numbering scheme that is. Every other resource-level identifier this module emits is namespaced (Patient through `PatientIdentifierSystemService`, both Practitioner translators through `OPENMRS_FHIR_EXT_PROVIDER_IDENTIFIER` and `.../user/identifier`, RelatedPerson, Task), and the ServiceRequest example we ship under `test-data/src/main/resources/org/openmrs/module/fhir2/narrative/example` pairs this same v2-0203 PLAC type with one.

A `type` with no `system` is legal, and the spec offers it precisely for the case where the namespace isn't known, so I'm not calling this a blocker. But we do know the namespace here, so I'd add something like `OPENMRS_FHIR_EXT_PREFIX + "/order/identifier"` to `FhirConstants` and set it alongside the type. If you do, please assert the type coding and the system in one of the two new tests as well: nothing in the repo asserts the PLAC coding today, so a typo in either URI would sail through CI.

dkayiwa (commented):
Nothing here blocks merge for me. One thing that has no line to attach it to: the title still mentions Accession Number, but the last two commits took the ACSN identifier back out, and FM2-686 only asks for the order number. This repo squashes with the PR title as the commit subject (cfdab1f is #615's six commits), so as it stands the history would claim accession-number support that isn't in the code. Worth trimming the title before it lands.

mseaton (approved):
LGTM

mogoodrich:
Thanks @mseaton !

@ibacher  I'm going ahead and merging this but happy to discuss.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/OrderIdentifierTranslatorImpl.java
ibacher: I think the suggestion here is, in fact, semantically incorrect and would put us in a worse place interoperability-wise. I agree that in an ideal world, we would have a system associated with an identifier, but the `system` + `value` combination needs to exactly identify one instance and only one instance of the order and that is a guarantee we can't make.

Order numbers are generated and are specific to each OpenMRS "instance", and while there is an associated "Instance ID", we do not enforce that this is filled out and in practice it rarely, if ever is. Consequently, we don't have what we need to construct as system that would correctly label the identifier.

If we hard-code something using, e.g., `OPENMRS_FHIR_EXT_PREFIX + "/order/identifier"`, ORD-1 generated from PIH Liberia would be declaring that it is the "same" as from ORD-1 from PIH-Mirebalais _and_ ORD-1 from Uganda EMR Kawaala HCIV (and every other OpenMRS installation). By leaving things "systemless" we're admitting the true state of things which is that we have an identifier but we do not have a scope for the identifier.

mogoodrich: So yeah, @ibacher , I didn't change anything based on what @dkayiwa  said above but note that the existing code *did* describe a system, ie what was added here: https://openmrs.atlassian.net/browse/FM2-313

This ticket expands this pattern to apply to the identifier on the order itself, ie:

```
 "identifier": [
        {
            "use": "usual",
            "type": {
                "coding": [
                    {
                        "system": "http://terminology.hl7.org/CodeSystem/v2-0203",
                        "code": "PLAC",
                        "display": "Placer Identifier"
                    }
                ]
            },
            "value": "ORD-1"
        }
    ],
```

Is this incorrect, or is something implicit in the defined terminology.hl7.org that this is not unique?

mogoodrich: Aha, this may be my confusion... Claude suggests there would be two systems, like:

```
{
  "identifier": [
    {
      "type": {
        "coding": [
          {
            "system": "http://terminology.hl7.org/CodeSystem/v2-0203",
            "code": "ACSN",
            "display": "Accession ID"
          }
        ]
      },
      "system": "http://your-facility.org/accession",
      "value": "2026-000481"
    }
  ]
}
```
Haven't sanity checked this response, but it makes sense in this case that the second "system" would be the unique namespace, and I agree we should not hard code that.  @ibacher  @dkayiwa

PR #622, opened 2026-08-13
FM2-678, FM2-682, FM2-683: FHIR2 2.5.1 — Backport Task status and focus changes for Bahmni

vvkpd:
## Summary

This PR proposes a **FHIR2 2.5.1 patch release** based on the existing 2.5.0 baseline, cherry-picking only the three required changes for Bahmni, plus one reference search correctness fix surfaced during review:

- **FM2-678**: Add support for Ready, On Hold and Cancelled Task Statuses
- **FM2-682**: Backport Add Task.focus support to the FHIR2 Task model (2.x)
- **FM2-683**: Extend FHIR Task to have Draft and In Progress Task Statuses

## Context

Bahmni is currently on FHIR2 2.5.0. Upgrading from 2.5.0 to 2.9.1 is a significant jump and has already surfaced issues (AllComponentsState/App Framework compatibility). The three changes above are the specific features required by Bahmni to support their Task workflow, and they can be cleanly backported without pulling in dependencies from newer FHIR2 versions.

A 2.5.1 patch release would allow Bahmni to consume these changes with a much smaller regression scope, avoiding a full FHIR2 major upgrade at this point. Without this, the Bahmni team would need to:
1. Wait for OpenMRS to resolve the Platform 2.6.x/App Framework compatibility issue, and
2. Perform extensive regression testing across the product for the 2.5.0 → 2.9.1 jump.

## Changes included

| Ticket | Description |
|--------|-------------|
| FM2-678 | Add support for Ready, On Hold and Cancelled Task Statuses |
| FM2-682 | Backport Task.focus support to the FHIR2 Task model (2.x) |
| FM2-683 | Extend FHIR Task to have Draft and In Progress Task Statuses |
| — | Fix Task reference searches (`owner`, `subject`, `based-on`, `focus`) to match against `targetUuid` instead of `reference`, correcting searches that silently returned 0 entries on 2.5.0 |

## Behavioral notes

**Reference search fix (behavior change):** The `handleReference` method previously compared `fhir_reference.reference` (stored as `Type/uuid`) against `getIdPart()` (bare uuid), so `owner`, `subject`, `based-on`, and `focus` reference searches always returned empty bundles on 2.5.0. This PR corrects the comparison to use `targetUuid`. Existing Bahmni deployments sending `GET /Task?owner=...` will receive results for the first time — this is the intended behaviour.

## Test plan

- [ ] FM2-678: Verify Task resource returns `ready`, `on-hold`, and `cancelled` statuses correctly
- [ ] FM2-682: Verify `Task.focus` field is populated and searchable via FHIR API
- [ ] FM2-683: Verify Task resource returns `draft` and `in-progress` statuses correctly
- [ ] Verify `GET /Task?owner=Practitioner/{uuid}` returns matching tasks (was broken/returning 0 on 2.5.0)
- [ ] No regressions on existing Task FHIR endpoints


api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImpl.java
```diff
@@ -199,6 +203,10 @@ private void setOpenmrsTaskFields(FhirTask openmrsTask, Task fhirTask) {
 			openmrsTask.setOwnerReference(referenceTranslator.toOpenmrsType(fhirTask.getOwner()));
 		}
 		
+		if (fhirTask.hasFocus()) {
```
dkayiwa, line 206: Writing `focus` here activates a mis-mapping in `TaskVersionConverter` that has been harmless until now. Lines 199-201 of that class (not touched by this PR) map DSTU3 `Task.context` onto R4 `Task.focus` rather than `Task.encounter`, and they run after the `focus` mapping just above them, so `context` overwrites whatever focus the client sent. The translator used to discard `focus`, so the wrong value went nowhere. With this block it gets persisted.

Running a DSTU3 Task with both fields set through the converter and this translator on your branch:

```
r3 focus = Observation/obs-uuid-1111, r3 context = Encounter/enc-uuid-2222
  -> r4 focus = Encounter/enc-uuid-2222, r4 encounter = <none>
  -> persisted focusReference = Encounter/enc-uuid-2222, encounterReference = null
```

So if this merges as is, a POST or PUT to `/ws/fhir2/R3/Task` carrying a `context` writes the Encounter reference into the new `fhir_task.focus_reference_id`, drops the client's real focus, and a later R4 read of that task reports `Task.focus` pointing at an Encounter. Nothing throws and no test covers it, so the R3 Task integration tests still pass. It is also the one place where "no regressions on existing Task FHIR endpoints" from your test plan does not hold.

One line in `TaskVersionConverter` fixes it, `tgt.setEncounter(convertReference(src.getContext()))`, and as a side effect R3 `context` would finally persist as the encounter instead of being dropped. That line is wrong on 2.x and master too, so fixing it there and cherry-picking works for me if you would rather keep this PR a straight backport. I just do not think 2.5.1 should ship the focus column with the R3 write path aimed at it.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
```diff
@@ -81,11 +87,11 @@ private Boolean validReferenceParam(ReferenceParam ref) {
 	
 	private Optional<Criterion> handleStatus(TokenAndListParam tokenAndListParam) {
 		return handleAndListParam(tokenAndListParam, token -> {
-			if (token.getValue() != null) {
+			if (token.getValue() != null && !token.getValue().isEmpty()) {
 				try {
-					return Optional.of(eq("status", FhirTask.TaskStatus.valueOf(token.getValue().toUpperCase())));
+					return Optional.of(eq("status", FhirTask.TaskStatus.valueOf(Task.TaskStatus.fromCode(token.getValue().toLowerCase()).name())));
```
dkayiwa: This comes out at 147 columns, past the 125 the OpenMRS formatter is configured for, so `formatter:format` rewraps it on the next build and leaves the tree dirty. Worth committing it already wrapped:

```suggestion
					return Optional.of(eq("status",
					    FhirTask.TaskStatus.valueOf(Task.TaskStatus.fromCode(token.getValue().toLowerCase()).name())));
```

api/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceProviderTest.java
```diff
@@ -56,230 +62,265 @@
...
+
 	private static final int COUNT = 1;
-	
+
 	@Mock
 	private FhirTaskService taskService;
-	
+
+	@Mock
+	private RequestDetails mockRequestDetails;
```
dkayiwa: Two leftovers from adapting the 2.x commit to this branch. This mock is not used by anything in the file: on 2.x the provider has `doUpsert(IdType, Task, RequestDetails, boolean)` and the tests pass the mock into it, but the 2.5.x provider has no such method. The cherry-pick also replaced this file's tab-indented blank lines with plain empty ones, which accounts for 60 of the 103 changed lines here and gets reverted by `formatter:format` on the next build. Dropping both would leave the file matching the 2.x commit it came from.

dkayiwa (commented):
This needs a couple of things before it can be a 2.5.1, and two of them have no line in the diff to sit on.

The poms are still at 2.5.0, which is a published release (`fhir2-api-2.5.0.jar` and `fhir2-omod-2.5.0.jar` are both on mavenrepo.openmrs.org), and nothing here bumps them. Once this merges, a build from 2.5.x produces a `fhir2-2.5.0.omod` whose contents are not the 2.5.0 everyone else has, and there is no 2.5.1-SNAPSHOT to release from. Every other maintenance branch sits on its next snapshot (2.7.x on 2.7.1-SNAPSHOT, 2.9.x on 2.9.1-SNAPSHOT); 2.5.x never got the "prepare for next development iteration" commit after the 2.5.0 release. Since this PR is the 2.5.1 proposal, I think the bump belongs in it.

The other one is that the checks on this PR do not cover the branch. `.github/workflows/main.yml` triggers on push and pull_request against `master` only, so the Maven build and test job never ran here and SonarCloud is all you are seeing. If we are going to cut a release from 2.5.x, adding it to those branch filters seems worth doing.

Those two plus the inline comment on `TaskTranslatorImpl` are what I would want settled before 2.5.1 goes out.

api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/TaskTranslatorImpl.java
vvkpd: Fixed in commit 64240233. Changed `tgt.setFocus(convertReference(src.getContext()))` to `tgt.setEncounter(convertReference(src.getContext()))` in `TaskVersionConverter` so R3 `Task.context` correctly persists as `encounter` and no longer overwrites `focusReference`.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
vvkpd: Fixed in commit 64240233. Rewrapped the line to stay within the 125-column formatter limit.

api/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceProviderTest.java
vvkpd: Fixed in commit 64240233. Removed the unused `RequestDetails` mock and restored the tab-indented blank lines to match the formatter expectations.

vvkpd:
All three items from @dkayiwa's review addressed in commit 64240233:

- **POM version bump**: All module POMs updated from `2.5.0` to `2.5.1-SNAPSHOT`.
- **CI workflow**: Added `2.5.x` to the `push` and `pull_request` branch filters in `.github/workflows/main.yml` so Maven builds and tests will run on PRs targeting this branch.
- **`TaskVersionConverter` inline comment**: Fixed — see reply on that thread.

api/src/main/java/org/openmrs/module/fhir2/providers/util/TaskVersionConverter.java
```diff
@@ -197,7 +197,7 @@ public static org.hl7.fhir.r4.model.Task convertTask(org.hl7.fhir.dstu3.model.Ta
 				tgt.setFor(convertReference(src.getFor()));
 			}
 			if (src.hasContext()) {
-				tgt.setFocus(convertReference(src.getContext()));
+				tgt.setEncounter(convertReference(src.getContext()));
```
dkayiwa, line 200: Worth raising the same one-liner on the other branches while it is fresh. `2.x` and `master` both still have `tgt.setFocus(convertReference(src.getContext()))` on this line, and both of them already carry `focusReference` on `FhirTask`, so on those branches an R3 `POST` or `PUT` that carries a `context` writes the Encounter into `fhir_task.focus_reference_id`, which is the same thing we were trying to keep out of 2.5.1. (`2.9.x` has the identical line but no focus column yet, so nothing lands there.)

The local half is that this line is now the one place 2.5.x deliberately differs from 2.x, and nothing in the tree says so, so the next Task cherry-pick into this branch will quietly put `setFocus` back. There is no `TaskVersionConverterTest` at the moment; a small one asserting that a DSTU3 Task carrying both `focus` and `context` comes out as r4 `focus` and `encounter` respectively would record the divergence and fail if it returns.

Neither of those needs to hold 2.5.1 up, but I would not leave the upstream one sitting.

api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java
```diff
@@ -42,10 +42,18 @@ public class FhirTask extends BaseOpenmrsMetadata {
...
 		ACCEPTED,
+		REJECTED,
+		READY,
+		CANCELLED,
+		INPROGRESS,
+		ONHOLD,
+		FAILED,
 		COMPLETED,
+		ENTEREDINERROR,
 		UNKNOWN
```
dkayiwa, line 57: Do you know whether Bahmni has any `fhir_task` rows sitting at `UNKNOWN`? It is now the only member here with no R4 code behind it, and that quietly changes what the status search does: `Task.TaskStatus.fromCode("unknown")` throws, since there is no `unknown` in the R4 `Task.status` value set, so `handleStatus` drops that token and the whole status criterion along with it. `GET /Task?status=unknown` therefore comes back with every task instead of the ones actually stored as `UNKNOWN`. On 2.5.0 it filtered.

I am asking rather than just flagging it because `UNKNOWN` is both the `fhir_task.status` column default and what 2.5.0 wrote whenever a client sent a status it did not recognise, which is exactly the set of codes these three tickets add. Reading one of those rows was already broken (`Task.TaskStatus.valueOf("UNKNOWN")` throws inside `TaskTranslatorImpl`, so the read blows up, and that part is not new here), and after this they cannot be found either. If such rows exist on their 2.5.0 database, 2.5.1 probably wants to move them to something readable; if not, I would leave it. Either way the widening is inherited from 2.x and master, so any fix belongs there too.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
dkayiwa: `mvn install` still rewrites this block. The formatter puts the `if` condition back on one line and re-wraps the `Optional.of` the way the suggestion above had it, so a build on this branch leaves `FhirTaskDaoImpl.java` modified in the working tree. For comparison, a build on 2.5.x itself only dirties `FhirDaoAop.java` and `FhirAutoProxyCreator.java`, and both of those were already that way before this PR. What it settles on:

```java
			if (token.getValue() != null && !token.getValue().isEmpty()) {
				try {
					return Optional.of(eq("status",
					    FhirTask.TaskStatus.valueOf(Task.TaskStatus.fromCode(token.getValue().toLowerCase()).name())));
```

Committing that text verbatim ends it.

api/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceProviderTest.java
dkayiwa: The mock field is gone, but two bits of it survived. `import ca.uhn.fhir.rest.api.server.RequestDetails;` is still on line 31 with nothing left using it, and the blank line after `private FhirTaskService taskService;` is still a plain empty line rather than the tab-indented one. `impsort` runs with `removeUnused` and the formatter puts the tab back, so `mvn install` leaves this file modified too. Dropping that import and restoring the one tab finishes it.

api/src/main/java/org/openmrs/module/fhir2/providers/util/TaskVersionConverter.java
vvkpd: Good catch on the upstream branches. Will raise the same `tgt.setEncounter(convertReference(src.getContext()))` fix on `openmrs/2.x` and `openmrs/master` in a follow-up PR. Will also add a `TaskVersionConverterTest` asserting that a DSTU3 Task with both `focus` and `context` maps to R4 `focus` and `encounter` respectively, so any future cherry-pick that reverts this line fails visibly.

vvkpd: Done — raised as PR #623 (https://github.com/openmrs/openmrs-module-fhir2/pull/623). Fixes both `openmrs/master` and includes `TaskVersionConverterTest` with two cases: DSTU3 Task carrying both `focus` and `context` mapping to R4 `focus` and `encounter` respectively, and `context`-only mapping to `encounter` without touching `focus`.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
PoojaSastry-TW: Committed the formatter's output verbatim so the tree stays clean after mvn install.

api/src/test/java/org/openmrs/module/fhir2/providers/r4/TaskFhirResourceProviderTest.java
vvkpd: Fixed in commit 1dd83343 — dropped the unused `RequestDetails` import and restored the tab-indented blank line after `taskService`.

api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java
vvkpd: We don't have a definitive answer on whether existing Bahmni deployments have `fhir_task` rows at `UNKNOWN` — the enum member was added long before this PR and may be in use. The `UNKNOWN` behaviour (broken reads, no R4 code mapping) is entirely pre-existing and unchanged by this PR; none of the three tickets touch it. Since any fix would belong on 2.x and master too, we'd prefer to track it as a follow-up and keep it out of the 2.5.1 scope.

api/src/main/java/org/openmrs/module/fhir2/providers/util/TaskVersionConverter.java
dkayiwa: Could you pull the `TaskVersionConverterTest` from #623 into this branch too? 2.5.x is the branch that actually needs it: `2.x` line 200 is still `tgt.setFocus(convertReference(src.getContext()))`, so this branch's line 200 is a deliberate divergence that nothing in the tree records, and the next Task cherry-pick in here is what would quietly restore `setFocus`. The copy sitting on master does not catch that.

It applies unchanged. I dropped the file onto 1dd83343 and both cases compiled and passed as-is, then reverting this line to `setFocus` failed both of them. `formatter:format` and `impsort:sort` leave the file byte-identical, so it should not cost another formatting round here.

Same as before, this does not need to hold 2.5.1 up.

vvkpd: Done — ported `TaskVersionConverterTest` into this branch in commit 203e8068. Both cases pass and `formatter:format` leaves the file unchanged.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
```diff
@@ -102,7 +109,7 @@ private void handleReference(Criteria criteria, ReferenceAndListParam reference,
 				}
 				
 				List<Optional<? extends Criterion>> criterionList = new ArrayList<>();
-				criterionList.add(Optional.of(eq(String.format("%s.reference", alias), param.getIdPart())));
+				criterionList.add(Optional.of(eq(String.format("%s.targetUuid", alias), param.getIdPart())));
```
dkayiwa, line 112: Worth a line in the 2.5.1 release notes: this changes what the existing Task reference searches return, not only the new `focus` one. `owner`, `subject` and `based-on` all run through this method, and in production `fhir_reference.reference` holds `Type/uuid` while `getIdPart()` is the bare uuid, so on 2.5.0 they never matched. POSTing a Task with an owner through the R4 endpoint and then searching for it:

```
GET /Task?owner=Practitioner/f9badd80-ab76-11e2-9e96-0800200c9a66
  2.5.0   -> 0 entries
  this PR -> 1 entry
```

It is a fix rather than a regression, and I would keep it exactly as it is; reverting would put 2.5.x out of step with 2.x for nothing. But the description lists three tickets and the test plan says no regressions on existing Task endpoints, and this is a fourth change, to endpoints Bahmni already calls. Anyone getting an empty bundle from `/Task?owner=...` today starts getting rows.

Coverage is also thinner than it looks. The fixtures store a bare uuid in `reference`, which matches either way, so the only test that pins this down is `searchForTasks_shouldReturnTasksByFocusReference`, where the reference happens to be typed. Reverting this line to `%s.reference` fails that one test and nothing else in the api or Task integration suites.

vvkpd: Good call. Agreed on both points.

Updated the PR description to call this out explicitly: the `targetUuid` fix unblocks `owner`, `subject`, and `based-on` reference searches that silently returned 0 entries on 2.5.0, not just the new `focus` search. Added it as a fourth entry in the changes table, a Behavioral notes section, and a dedicated test plan checkbox.

On coverage — acknowledged. The bare-uuid fixtures masked the gap; `searchForTasks_shouldReturnTasksByFocusReference` is the only test that actually exercises the typed-reference path. Will track adding typed-reference variants for `owner` and `subject` as a follow-up rather than holding up the 2.5.1 merge.

api/src/main/java/org/openmrs/module/fhir2/model/FhirTask.java
```diff
@@ -42,10 +42,18 @@ public class FhirTask extends BaseOpenmrsMetadata {
 	
 	// Based on https://www.hl7.org/fhir/task.html v4.0.1
 	public enum TaskStatus {
+		DRAFT,
```
dkayiwa, line 45: Something to settle on 2.9.x before Bahmni ships on this, rather than here. Once they start writing these statuses, the 2.5.1 to 2.9.1 move that this PR defers stops being available to them: released 2.9.0 and 2.9.1, and the current 2.9.x head, still carry the original five constants, and `status` is mapped `@Enumerated(EnumType.STRING)` there exactly as it is here, so Hibernate cannot hydrate a row whose stored status it has no name for.

I narrowed this enum on your branch down to 2.9.x's five and ran `FhirTaskDaoImplTest` against the dataset this PR ships. 18 of the 19 pass; the one that loads every task does not:

```
searchForTasks_shouldReturnAllTasksForEmptyStatus
java.lang.IllegalArgumentException: No enum constant org.openmrs.module.fhir2.model.FhirTask.TaskStatus.INPROGRESS
	at org.hibernate.type.descriptor.java.EnumJavaTypeDescriptor.fromName(EnumJavaTypeDescriptor.java:84)
	at org.hibernate.loader.Loader.loadFromResultSet(Loader.java:1907)
```

It fails per query rather than per row, so on 2.9.1 the casualty is an unfiltered `GET /Task`, or any owner, subject or code search whose page happens to contain one of these tasks, not only a read of the task itself.

Widening the enum on 2.9.x is code only, `fhir_task.status` is already `varchar(255)` on that branch, so it looks worth doing before anyone moves a 2.5.1 database onto it. Upgrading forwards instead is fine from 4.1.0, which is the earliest release carrying all thirteen: 4.0.0 has FM2-678's three but not FM2-683's, and 3.0.0 and the whole 2.9.x line have neither. Failing the 2.9.x backport I would at least put it in the 2.5.1 release notes. This does not need to hold 2.5.1 up.

pom.xml
```diff
@@ -4,7 +4,7 @@
 
     <groupId>org.openmrs.module</groupId>
     <artifactId>fhir2</artifactId>
-    <version>2.5.0</version>
+    <version>2.5.1-SNAPSHOT</version>
```
dkayiwa, line 7: The scm block at line 35 still says `<tag>2.5.0</tag>`. The "prepare for next development iteration" commit that 2.5.x never got does both halves in one go: on master, e2799a4e moved the version from `4.2.0` to `4.3.0-SNAPSHOT` and the tag from `4.2.0` to `HEAD` in the same commit, and master, 2.x and 2.9.x all sit on `HEAD` today. `release:prepare` rewrites it anyway when 2.5.1 is cut, so nothing depends on this, it just leaves the branch claiming to be built from the 2.5.0 tag until then. Worth flipping to `HEAD` while you are in the file.

PR #626, opened 2026-08-25
FM2-697 | Add. Support for encounter filter and includes in Observation lastn operation

mohan-13:
This PR enhances the `lastn` operation in ObservationResource providers to support filtering of lastest Observation by a list of encounters.

This enhancement is needed to efficiently leverage the `lastn` operation to fetch the latest observations that are captured in a specific visit/episode. This will help clients to query for the latest of observations that are captured during a visit/episode without running extensive filters on the consuming side.

A few common examples:
- Query for the latest vitals signs in a visit
- Query for the latest BMI during an episode

## Issue I worked on

see https://issues.openmrs.org/browse/FM2-697

  No? -> execute above command

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/providers/r4/ObservationFhirResourceProvider.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/providers/r4/ObservationFhirResourceProvider.java
@@ -15,6 +15,7 @@ import static lombok.AccessLevel.PROTECTED;
 import javax.annotation.Nonnull;
 
 import java.util.HashSet;
+import java.util.Set;
 
 import ca.uhn.fhir.model.api.Include;
 import ca.uhn.fhir.model.valueset.BundleTypeEnum;
@@ -160,6 +161,9 @@ public class ObservationFhirResourceProvider implements IResourceProvider {
 		        id, lastUpdated, sort, includes, revIncludes));
 	}
 	
+	// Note: @OperationParam has no chainWhitelist, so the encounter parameter accepts only a plain
+	// reference (e.g. encounter=<uuid>). Chained modifiers such as encounter.type are silently
+	// ignored by HAPI; use the @Search endpoint above for chained encounter filtering.
 	@Operation(name = "lastn", idempotent = true, type = Observation.class, bundleType = BundleTypeEnum.SEARCHSET)
 	public IBundleProvider getLastnObservations(@OperationParam(name = "max") NumberParam max,
 	        @OperationParam(name = Observation.SP_SUBJECT) ReferenceAndListParam subjectParam,
@@ -168,9 +172,9 @@ public class ObservationFhirResourceProvider implements IResourceProvider {
 	        @OperationParam(name = Observation.SP_CODE) TokenAndListParam code,
 	        @OperationParam(name = Observation.SP_ENCOUNTER) ReferenceAndListParam encounterParam,
 	        @IncludeParam(allow = { "Observation:" + Observation.SP_ENCOUNTER, "Observation:" + Observation.SP_PATIENT,
-	                "Observation:" + Observation.SP_HAS_MEMBER }) HashSet<Include> includes,
+	                "Observation:" + Observation.SP_HAS_MEMBER }) Set<Include> includes,
 	        @IncludeParam(reverse = true, allow = { "Observation:" + Observation.SP_HAS_MEMBER,
-	                "DiagnosticReport:" + DiagnosticReport.SP_RESULT }) HashSet<Include> revIncludes) {
+	                "DiagnosticReport:" + DiagnosticReport.SP_RESULT }) Set<Include> revIncludes) {
 		if (patientParam != null) {
 			subjectParam = patientParam;
 		}
--- a/api/src/main/java/org/openmrs/module/fhir2/providers/r4/ObservationFhirResourceProvider.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/providers/r4/ObservationFhirResourceProvider.java
@@ -175,21 +175,13 @@ public class ObservationFhirResourceProvider implements IResourceProvider {
 			subjectParam = patientParam;
 		}
 		
-		if (CollectionUtils.isEmpty(includes)) {
-			includes = null;
-		}
-		
-		if (CollectionUtils.isEmpty(revIncludes)) {
-			revIncludes = null;
-		}
-		
 		ObservationSearchParams searchParams = new ObservationSearchParams();
 		searchParams.setPatient(subjectParam);
 		searchParams.setCategory(category);
 		searchParams.setCode(code);
 		searchParams.setEncounter(encounterParam);
-		searchParams.setIncludes(includes);
-		searchParams.setRevIncludes(revIncludes);
+		searchParams.setIncludes(CollectionUtils.isEmpty(includes) ? null : includes);
+		searchParams.setRevIncludes(CollectionUtils.isEmpty(revIncludes) ? null : revIncludes);
 		
 		return observationService.getLastnObservations(max, searchParams);
 	}
--- a/api/src/main/java/org/openmrs/module/fhir2/providers/r4/ObservationFhirResourceProvider.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/providers/r4/ObservationFhirResourceProvider.java
@@ -165,15 +165,31 @@ public class ObservationFhirResourceProvider implements IResourceProvider {
 	        @OperationParam(name = Observation.SP_SUBJECT) ReferenceAndListParam subjectParam,
 	        @OperationParam(name = Observation.SP_PATIENT) ReferenceAndListParam patientParam,
 	        @OperationParam(name = Observation.SP_CATEGORY) TokenAndListParam category,
-	        @OperationParam(name = Observation.SP_CODE) TokenAndListParam code) {
+	        @OperationParam(name = Observation.SP_CODE) TokenAndListParam code,
+	        @OperationParam(name = Observation.SP_ENCOUNTER) ReferenceAndListParam encounterParam,
+	        @IncludeParam(allow = { "Observation:" + Observation.SP_ENCOUNTER, "Observation:" + Observation.SP_PATIENT,
+	                "Observation:" + Observation.SP_HAS_MEMBER }) HashSet<Include> includes,
+	        @IncludeParam(reverse = true, allow = { "Observation:" + Observation.SP_HAS_MEMBER,
+	                "DiagnosticReport:" + DiagnosticReport.SP_RESULT }) HashSet<Include> revIncludes) {
 		if (patientParam != null) {
 			subjectParam = patientParam;
+		}
+		
+		if (CollectionUtils.isEmpty(includes)) {
+			includes = null;
+		}
+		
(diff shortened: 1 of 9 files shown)


angshu (approved):
lgtm. thanks @mohan-13

api/src/main/java/org/openmrs/module/fhir2/providers/r4/ObservationFhirResourceProvider.java
```diff
@@ -165,15 +165,31 @@ public IBundleProvider getLastnObservations(@OperationParam(name = "max") Number
...
+		
+		if (CollectionUtils.isEmpty(revIncludes)) {
+			revIncludes = null;
+		}
+		
 		ObservationSearchParams searchParams = new ObservationSearchParams();
 		searchParams.setPatient(subjectParam);
 		searchParams.setCategory(category);
 		searchParams.setCode(code);
+		searchParams.setEncounter(encounterParam);
```
dkayiwa, line 186: The one thing I'd want added before this merges is a test that actually runs the filtered query.

The filter itself works. With the standard test data plus `FhirObservationDaoImplTest_initial_data_2.1.xml` loaded, `Observation/$lastn?max=5&patient=5946f880-b197-400b-9caa-a3c661d23041&encounter=eec646cb-c847-45a7-98bc-91c8c4f70add` comes back with total 6, all of them in that encounter, where the same call without `encounter` returns 21, and an unknown uuid returns 0. On master every one of those variants returns 21, so the change does what FM2-697 asks, and R3 behaves the same.

Nothing in the PR would notice if that stopped being true, though. All twelve new tests are omod web tests, and those mock `FhirObservationService` (`verify(observationService).getLastnObservations(isNull(), searchParamsCaptor.capture())`), so what they assert is that HAPI bound the parameter and the provider copied it onto `ObservationSearchParams`. They pass just as happily if the DAO never applies the predicate.

That is worth covering here specifically, because the `$lastn` branch of `FhirObservationDaoImpl` doesn't reuse the shared search path: `getSearchResults` builds its own `CriteriaQuery` with its own ordering and batching loop, and `getSearchResultsCount` builds a separate `multiselect`/`groupBy` query. The existing encounter coverage is all against plain search, so neither of those hand-written queries is currently exercised with an encounter predicate. The project also covers every other `$lastn` parameter at query level (`ObservationSearchQueryTest` has `searchForLastnObs_shouldHandleNormalRequest` / `NoPatientReference` / `NoCategory` / `NoCode` / `MaxIsOne`, and both integration tests have a `shouldReturnLastnObservations*` case per parameter), which leaves this one as the odd parameter out.

One case in either place, with the counts above, would close it. The same gap applies to the `_include`/`_revinclude` support, which also works (`$lastn?_include=Observation:encounter` really does add the three distinct Encounters for the page), so a single integration test could cover both.

api/src/main/java/org/openmrs/module/fhir2/providers/r4/ObservationFhirResourceProvider.java
```diff
@@ -165,15 +165,31 @@ public IBundleProvider getLastnObservations(@OperationParam(name = "max") Number
 	        @OperationParam(name = Observation.SP_SUBJECT) ReferenceAndListParam subjectParam,
 	        @OperationParam(name = Observation.SP_PATIENT) ReferenceAndListParam patientParam,
 	        @OperationParam(name = Observation.SP_CATEGORY) TokenAndListParam category,
-	        @OperationParam(name = Observation.SP_CODE) TokenAndListParam code) {
+	        @OperationParam(name = Observation.SP_CODE) TokenAndListParam code,
+	        @OperationParam(name = Observation.SP_ENCOUNTER) ReferenceAndListParam encounterParam,
```
dkayiwa, line 173: HAPI matches an `@OperationParam` by exact name or `name:modifier` only (`OperationParameter.translateQueryParametersIntoServerArgumentForGet`), and it silently drops query parameters an operation doesn't declare. So `Observation/$lastn?patient=<p>&encounter.type=<t>` returns 200 with every observation for the patient, while the same chain on the `@Search` method above filters properly: 21 entries against 14 on the test data. `@OperationParam` has no `chainWhitelist`, so there's no declarative fix, but a sentence on the operation saying `encounter` takes a plain reference would save the next person the debugging.

Same mechanism, and follow-up material rather than more scope here: `$lastn-encounters` just below now quietly differs from `$lastn`, in that `$lastn-encounters?_include=Observation:encounter` returns no Encounter entries and no error.

Both points apply equally to the R3 provider.

api/src/main/java/org/openmrs/module/fhir2/providers/r4/ObservationFhirResourceProvider.java
mohan-13: Thanks for reviewing @dkayiwa. I have added a test case.

api/src/main/java/org/openmrs/module/fhir2/providers/r4/ObservationFhirResourceProvider.java
mohan-13: Resolved this as well

mohan-13:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=openmrs_openmrs-module-fhir2&pullRequest=626) **Quality Gate failed**
> Failed conditions  [B Maintainability Rating on New Code](https://sonarcloud.io/project/issues?id=openmrs_openmrs-module-fhir2&pullRequest=626&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true) (required ≥ A)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=openmrs_openmrs-module-fhir2&pullRequest=626)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

The FHIR providers needs these parameters to be passed.

api/src/main/java/org/openmrs/module/fhir2/providers/r4/ObservationFhirResourceProvider.java
```diff
@@ -160,12 +161,20 @@ public IBundleProvider searchObservations(
 		        id, lastUpdated, sort, includes, revIncludes));
 	}
 	
+	// Note: @OperationParam has no chainWhitelist, so the encounter parameter accepts only a plain
+	// reference (e.g. encounter=<uuid>). Chained modifiers such as encounter.type are silently
+	// ignored by HAPI; use the @Search endpoint above for chained encounter filtering.
 	@Operation(name = "lastn", idempotent = true, type = Observation.class, bundleType = BundleTypeEnum.SEARCHSET)
 	public IBundleProvider getLastnObservations(@OperationParam(name = "max") NumberParam max,
```
dkayiwa, line 168: On the quality gate: you're right that these parameters have to sit on the signature. HAPI binds each one from its annotation, which is also why the `@Search` method above carries sixteen, and the only two issues the gate fails on are S107 (more than 7 parameters) here and on the R3 twin. Nothing to change for it. SonarCloud isn't one of master's required checks either (the ruleset asks for `build / Build with Java 8`), so the red gate won't hold the merge up.

api/src/main/java/org/openmrs/module/fhir2/providers/r4/ObservationFhirResourceProvider.java
dkayiwa: The `_include`/`_revinclude` half of this is still uncovered. The six new web tests mock `FhirObservationService`, so what they pin down is that HAPI bound `_include` and that the provider copied it onto `ObservationSearchParams`; nothing checks what comes back in the bundle. Not blocking, since everything past the provider is the same code path `@Search` uses and that part is covered at query level.

If you do add a case, `ObservationSearchQueryTest` beside `searchForLastnObs_shouldHandleNormalRequest` is the place for it. The integration-tests module only runs under `-Pintegration-test` and CI runs a plain `mvn clean install`, so nothing in the pipeline executes the two encounter cases you added either; my "either place" above was wrong about that, and mirroring them there would fix it.

mohan-13: Resolved.

api/src/test/java/org/openmrs/module/fhir2/api/search/ObservationSearchQueryTest.java
```diff
@@ -1645,7 +1645,142 @@ public void searchForLastnObs_shouldHandleRequestWhenMaxIsOne() {
...
+	public void searchForLastnObs_shouldHandleEncounterRequest() {
+		ReferenceAndListParam referenceParam = new ReferenceAndListParam();
+		ReferenceParam patient = new ReferenceParam();
+
+		patient.setValue(PATIENT_UUID);
+
+		referenceParam.addValue(new ReferenceOrListParam().add(patient));
+
+		ReferenceAndListParam encounterReference = new ReferenceAndListParam()
+		        .addAnd(new ReferenceOrListParam().add(new ReferenceParam().setValue(ENCOUNTER_UUID)));
```
dkayiwa, line 1659: One more case would be worth adding here. FM2-697 asks for a set of encounters ("we need an Optional Param to specify a set of encounters from which the latest Observations can be fetched"), and the PR description says a list of encounters, but every case added here passes a single uuid, and so do the six web tests and the four integration tests.

The multi-value path does work. `handleEncounterReference` puts the reference through `handleAndListParam`, which ORs one `e.uuid = ?` predicate per token, and I checked it against this dataset: for `PATIENT_UUID` with `max=5`, encounter `6519d653-393b-4118-9c83-a3715b82d4ac` on its own returns 14 observations, `eec646cb-c847-45a7-98bc-91c8c4f70add` on its own returns 6, the two in one or-list return 20 spanning both, and two unknown uuids return 0. Token order makes no difference.

So nothing is broken, but the part of the ticket that motivates the And/List type is the part with no test behind it, and whoever touches this signature next has nothing telling them a plain `ReferenceParam` would not do. Adding a second uuid to the or-list above and expecting 20 would cover it. One snag if you go looking for a ready-made constant: `ENCOUNTER_UUID_TWO` in this class is the same string as `ENCOUNTER_UUID`, so it needs a genuinely new one.

api/src/test/java/org/openmrs/module/fhir2/api/search/ObservationSearchQueryTest.java
```diff
@@ -1645,7 +1645,142 @@ public void searchForLastnObs_shouldHandleRequestWhenMaxIsOne() {
 		assertThat(resultList, everyItem(anyOf(allOf(is(instanceOf(Observation.class))))));
 		assertThat(resultList, isSortedAndWithinMax(1));
 	}
-	
+
```
dkayiwa, line 1648: The blank lines inside these new methods are bare, where the project's formatter wants them indented to the surrounding level, so spotless rewrites 42 lines of this file on the next build. It is the only file in the tree this happens to: `mvn process-sources` on master modifies nothing, and on this branch it modifies only this one. The spotless execution is bound to `apply` rather than `check`, so no CI job notices, and the reformat instead lands in the working copy of whoever builds next. `mvn clean package` and amending would settle it.

PR #627, opened 2026-08-25
FM2-669: Add support for client provided resource Ids - Backport

VaishSiddharth:
## Issue I worked on

  No? -> execute above command


kipchubett (approved):

api/src/main/java/org/openmrs/module/fhir2/providers/r4/BaseUpsertFhirResourceProvider.java
```diff
@@ -0,0 +1,80 @@
...
+	
+	private List<String> getSupportedResources() {
+		if (globalPropsService == null) {
+			globalPropsService = Context.getRegisteredComponents(FhirGlobalPropertyService.class).get(0);
+		}
+		
+		String value = globalPropsService.getGlobalProperty(GP_NAME_SUPPORTED_RESOURCES);
+		if (value != null) {
+			return Arrays.stream(StringUtils.split(value, ',')).filter(v -> v.trim().length() > 0)
+			        .collect(Collectors.toList());
```
dkayiwa, line 74: `StringUtils.split` doesn't trim, so the entries collected here keep any leading space, while `upsert()` above compares them against a bare `resource.fhirType()`. I ran this method's body verbatim against commons-lang3 3.14.0:

```
GP=[Medication,Location]  -> [Medication, Location]   contains("Location")=true
GP=[Medication, Location] -> [Medication,  Location]  contains("Location")=false
GP=[ Location ]           -> [ Location ]             contains("Location")=false
```

So an admin who writes `Medication, Location` into `fhir2.upsert.supported.resources`, which is how the GP description's "Comma-separated list" reads, gets upsert on Medication only: PUTs to `/Location/{new-uuid}` go on returning 404 and nothing logs why. The new integration test sets a single value with no spaces, so CI can't see it.

The `v.trim().length() > 0` filter shows the intent to tolerate whitespace was already there, it's just applied to the test and not to the value:

```suggestion
			return Arrays.stream(StringUtils.split(value, ',')).map(String::trim).filter(v -> !v.isEmpty())
			        .collect(Collectors.toList());
```

I applied that here, ran `formatter:format` (it leaves those two lines exactly as written) and `MedicationFhirResourceProviderIntegrationTest` still passes. Worth fixing before this merges. master carries the same line, so it wants the same change.

api/src/main/java/org/openmrs/module/fhir2/providers/r4/MedicationFhirResourceProvider.java
```diff
@@ -78,18 +76,19 @@ public Medication getMedicationByUuid(@IdParam @Nonnull IdType id) {
...
-		if (id == null || id.getIdPart() == null) {
-			throw new InvalidRequestException("id must be specified to update");
-		}
-		
-		medication.setId(id.getIdPart());
-		
-		return FhirProviderUtils.buildUpdate(fhirMedicationService.update(id.getIdPart(), medication));
-	}
+
+    @Override
```
dkayiwa, line 80: This block, and the new lines in `BaseFhirService`, are indented with spaces, so the OpenMRS formatter rewrites them on the next build. I ran `mvn clean verify -Pintegration-test` on this branch with JDK 8: it passes, but `formatter:format` rewrote both of those files as a side effect. On 2.5.x without this PR, the same run only touches `FhirDaoAop.java` and `FhirAutoProxyCreator.java`, which were already out of format. So whoever builds next finds two of your files modified and either commits the reformat or has to unpick it.

`mvn formatter:format impsort:sort` and amend clears it, and it puts both files back into the same shape as master's copy of this change. Worth doing before this merges, since it's one command.

omod/src/test/java/org/openmrs/module/fhir2/providers/r4/LocationFhirResourceProviderWebTest.java
```diff
@@ -444,7 +457,7 @@ public void updateLocation_shouldUpdateExistingLocation() throws Exception {
 		Location location = new Location();
 		location.setId(LOCATION_UUID);
 		
-		when(locationService.update(anyString(), any(Location.class))).thenReturn(location);
+		lenient().when(locationService.update(anyString(), any(Location.class))).thenReturn(location);
```
dkayiwa, line 460: The provider now calls `update(uuid, resource, requestDetails, createIfNotExists)`, so this stub sits on a method that no longer runs, and the `lenient()` is what keeps Mockito quiet about it. Two checks on this branch:

* deleting this line outright leaves `LocationFhirResourceProviderWebTest` green at 27/27;
* changing `lenient().when` back to `when` fails the class with `UnnecessaryStubbingException`.

The four-argument call then returns null from the mock, `FhirProviderUtils.buildUpdate(null)` gives a `MethodOutcome` with no resource, and `isOk()` still passes, so this test no longer shows the provider reaching the service at all. The Medication and Practitioner versions, and `updateTask_shouldUpdateExistingTask`, are in the same position; `updateTask_shouldErrorForNonexistentTask` is the one stub that did get moved to the new signature.

Stubbing the real signature restores the check:

```java
when(locationService.update(anyString(), any(Location.class), any(RequestDetails.class), anyBoolean()))
        .thenReturn(location);
```

with the `lenient()` dropped. It needs `org.mockito.ArgumentMatchers.anyBoolean` and `ca.uhn.fhir.rest.api.server.RequestDetails` imported, which is why I haven't put it in a suggestion block. I'd rather fix these four than carry them over: on a backport they're exactly the tests that would catch a wiring slip between provider and service.

integration-tests/src/test/java/org/openmrs/module/fhir2/providers/r4/MedicationFhirResourceProviderIntegrationTest.java
```diff
@@ -613,4 +634,56 @@ public void shouldReturnNotModifiedWhenRetrievingAnExistingMedicationWithAnEtag(
...
+	public void shouldCreateANewMedicationWithTheProvidedUuid() throws Exception {
+		GlobalProperty gp = adminService.getGlobalPropertyObject(GP_NAME_SUPPORTED_RESOURCES);
+		if (gp == null) {
+			gp = new GlobalProperty(GP_NAME_SUPPORTED_RESOURCES);
+			gp.setDatatypeClassname(FreeTextDatatype.class.getName());
+		}
+		gp.setValue("Medication");
+		adminService.saveGlobalProperty(gp);
+		final String uuid = "105bf9c4-8fef-11f0-b303-0242ac180002";
+		assertThat(conceptService.getDrug(uuid), nullValue());
```
dkayiwa, line 648: `ConceptService.getDrug(String)` takes a drug name or id, not a uuid, so this assertion holds whether or not the drug exists. I added a probe to this test straight after the PUT, when the drug is definitely there:

```
PROBE getDrug(uuid)=null
PROBE getDrugByUuid(uuid)=Drug[hashCode=8e36ba62,uuid=105bf9c4-8fef-11f0-b303-0242ac180002]
```

`getDrugByUuid` is the one that looks at uuids. Switching to it here and in `shouldFailToCreateANewMedicationWithTheProvidedUuidIfUpsertIsNotEnabled` below makes the precondition real, and asserting it again after the PUT would also confirm the drug was stored under the client's id rather than just echoed back in the response body. These two lines are the only uses of `getDrug(` anywhere in the repo. Not a blocker, but it is a one-word change.

dkayiwa (commented):
One question decides whether this goes onto 2.5.x at all, and the PR doesn't answer it: which release is it aimed at, and has putting a new feature into the patch line been agreed? The description is the unfilled template with the issue link left as a bare `.../browse/FM2-`, FM2-669 is still sitting in "Code Review (Initial)" in JIRA and says nothing about 2.5.x, and the branch name (`product-catalog-backport-new`) points at a driver that isn't recorded anywhere. #622 named both the target release and the requesting downstream in its title, which made that call easy to make. Could you put that in the description?

ibacher (approved):

ibacher:
Accepting this. While the defects @dkayiwa identifies are correct, they are also part of the code that this is backporting, so as a backport, this is correct.

PR #628, opened 2026-08-26
FM2-698: Add support for externally configured interceptors

mherman22:
https://openmrs.atlassian.net/browse/FM2-698

## What the ticket asked

`FhirRestServlet.refreshed()` called `unregisterAllInterceptors()` and then re-registered a fixed set
of interceptors, so an interceptor another module had registered was silently dropped — and that
happens on every OpenMRS context refresh, which is every module start or stop. The ticket asks for an
extension point that survives it, contributed through Spring the way resource providers already are,
without anyone having to subclass `FhirRestServlet`.

## What this does

`@FhirInterceptor` marks a Spring bean as a HAPI interceptor to register, beside the existing
`@R3Provider` / `@R4Provider`. `FhirRestServlet.registerInterceptors()` registers this module's own
interceptors and then every `@FhirInterceptor` bean in the application context, and both
`initialize()` and `refreshed()` call it.

The beans are re-resolved from the refreshed context on every call rather than captured once. The
ticket proposes the other shape — "the configured interceptors should be retained by the servlet" —
and that shape silently loses the case the ticket exists for, because a module that starts *after*
fhir2 has no beans in the context at `initialize()` time. Re-resolving picks it up. OpenMRS also
constructs a new servlet instance on module restart, so nothing held in a field would have survived
anyway. `refreshed_shouldRegisterAContributedInterceptorThatReachedTheContextAfterInitialize` pins
this; it was checked by building the captured-collection design and confirming it was the only test
of the eight to fail.

`getModuleApplicationContext()` is a seam over `FhirActivator`'s static context, read by
`registerInterceptors()`, `refreshed()` and `autoInject()`, and overridden by the tests.

## Three fixes beyond the ticket

Reviewing and verifying this change turned up three defects in the code it touches. Each is a
separate commit with its own regression test, and each was watched failing before the fix.

**1. Every module start or stop left the FHIR API returning 404 for every resource.**
`refreshed()` called `unregisterAllProviders()`, which destroys the method bindings a read routes on,
and then `setResourceProviders(...)`, whose whole body is validate / `clear()` / `addAll()` on a
field. Nothing rebound, because binding happens in `RestfulServer.init()`, which a module lifecycle
callback never reaches. `/metadata` kept returning 200 throughout, because the conformance provider is
bound separately — so the obvious health check stayed green while every read failed, and it did not
recover until the webapp was reloaded. Now `registerProviders(...)`, which adds to the field *and*
calls `findResourceMethods`, and which also re-runs the provider `@Initialize` methods that
`unregisterAllProviders()` had just destroyed. Present since `refreshed()` was introduced in
`4289a669` (FM2-163, April 2021); reproduced on unmodified 4.2.0, so not caused by this PR.

**2. A request arriving mid-refresh was served with no authentication.**
`refreshed()` unregistered every interceptor at the top of its block and did not put them back until
the bottom, with the provider rebuild and two uncached global-property reads in between. The servlet
keeps answering throughout, and a request carrying no credentials passes straight through
`AuthenticationFilter`, which rejects only a request whose Basic credentials fail. The window
predates this change, but moving the contributed-bean lookup to the end of the block widened it, so
the teardown and rebuild are now adjacent, with `RequireAuthenticationInterceptor` registered first
within the rebuild. Adjacency narrows the window to two statements rather than closing it; closing it
would mean populating a fresh `InterceptorService` and swapping it in with `setInterceptorService`,
which is a bigger change than this ticket justifies.

**3. One servlet's failure left the other un-refreshed.**
`FhirActivator` dispatched lifecycle events with `lifecycleListeners.forEach(...)`, which aborts at
the first listener that throws — and both FHIR servlets register as listeners. All five dispatch
sites now log and continue.

## The decision a reviewer should confirm

A contributed bean that cannot be supplied is deliberately not caught. That choice is inherited, not
new here, but this change is what first puts another module's code on the `initialize()` path, and the
blast radius there is larger than it looks: at server startup the `ModuleException` escapes
`Listener.performWebStartOfModules`, whose tail loop is unguarded, and aborts the whole OpenMRS boot
rather than just this module. Only a bean whose creation is deferred past the context refresh — lazy,
prototype or scoped — can reach that path; a plain singleton fails the context refresh first. If that
trade is not wanted, the fix is to catch and log around the contributed lookup in `initialize()` only.

Two facts a module author may want that are deliberately not in the annotation's javadoc, because
nothing can falsify them there: a contributed bean declaring a negative `@Interceptor(order)` runs
ahead of this module's own interceptors, authentication included, since those all leave order at 0;
and on `Pointcut.SERVER_INCOMING_REQUEST_PRE_PROCESSED`, which is handed no `RequestDetails`,
`FhirVersionUtils.getFhirVersion(request)` is how to tell R3 from R4.

## How it was verified

Every test here was watched failing on the pre-change code and then checked by mutation: revert the
behaviour it covers and confirm that test, and no other, goes red. To check one yourself, mutate the
line it covers and read which test fails. `mvn clean install` is green from the repository root.

Fix 1 was then verified on a running OpenMRS 2.8.8 reference-application backend, against a control.
Both builds were pinned by sha256 at every path the module can load from, and the module's own version
was read back over REST rather than inferred from the filename.

On the stock shipped 4.2.0, an authenticated `GET /ws/fhir2/R4/Patient?_count=1` returned a Bundle of
50. Restarting one unrelated started module (`teleconsultation`) via `POST /ws/rest/v1/moduleaction`
returned 201, and the log shows the context refresh running to completion. Reading the same URL
afterwards returned **404**, body `HAPI-0302: Unknown resource type 'Patient' - Server knows how to
handle: [OperationDefinition]`, and the R3 equivalent likewise. `/R4/metadata` stayed 200 throughout,
which is what makes the outage invisible to a health check.

On this branch the identical sequence returns **200** afterwards, on both R4 and R3, and `HAPI-0302`
does not appear in the log at all.

Authentication was checked on the same deployment: unauthenticated `/R4/Patient` is 401, authenticated
is 200, and unauthenticated `/R4/metadata` and `/R4/.well-known/smart-configuration` are both 200. That
split is the discriminator, since a servlet filter would have rejected all three, so the 401 is
produced inside the interceptor chain. The contributed-bean lookup ran against a real context holding
no `@FhirInterceptor` beans and left the boot clean.

Two facts a module author may want, kept here rather than in the annotation's javadoc because nothing
in the code can falsify them: a contributed bean declaring a negative `@Interceptor(order)` runs ahead
of this module's own interceptors, authentication included, since those all leave order at 0; and a
contributed bean whose constructor throws is deliberately not caught, so from `initialize()` at server
startup it aborts OpenMRS boot rather than coming up with that interceptor quietly missing.

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/annotations/FhirInterceptor.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/annotations/FhirInterceptor.java
@@ -18,24 +18,13 @@ import java.lang.annotation.Target;
 import org.springframework.beans.factory.annotation.Qualifier;
 
 /**
- * Marks a Spring bean as a HAPI interceptor the FHIR REST servlets should register, so a module
- * other than this one can add a cross-cutting concern -- authorization, audit, consent -- in the
- * way {@link R4Provider} marks a resource provider.
+ * Marks a Spring bean as a HAPI interceptor every FHIR REST servlet registers, the way
+ * {@link R4Provider} marks a resource provider.
  * <p>
- * The bean must carry at least one {@code Hook} method: HAPI logs "Interceptor registered with no
- * valid hooks" and drops one that does not, so it is absent from the request path with only a
- * warning to say so.
+ * Carry at least one {@code Hook} method; HAPI drops an interceptor without one.
  * <p>
- * Registration order is only HAPI's tie-break -- it sorts each hook by {@code Interceptor(order)},
- * which a non-zero {@code Hook(order)} overrides per method. This module's own interceptors all
- * leave that at 0, so a contributed bean with a negative order runs ahead of them, authentication
- * included.
- * <p>
- * Every version-specific servlet registers it, so one meant for a single FHIR version must work out
- * which it is serving: {@code getFhirContext().getVersion().getVersion()} on a hook handed
- * {@code RequestDetails}, and on {@code Pointcut.SERVER_INCOMING_REQUEST_PRE_PROCESSED}, which is
- * passed only the request and the response,
- * {@code org.openmrs.module.fhir2.web.util.FhirVersionUtils.getFhirVersion(request)}.
+ * On a hook handed no {@code RequestDetails}, read the FHIR version from
+ * {@code FhirVersionUtils.getFhirVersion(request)} rather than the servlet path.
  */
 @Target({ ElementType.CONSTRUCTOR, ElementType.FIELD, ElementType.METHOD, ElementType.TYPE })
 @Retention(RetentionPolicy.RUNTIME)
--- a/api/src/main/java/org/openmrs/module/fhir2/api/annotations/FhirInterceptor.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/annotations/FhirInterceptor.java
@@ -18,25 +18,23 @@ import java.lang.annotation.Target;
 import org.springframework.beans.factory.annotation.Qualifier;
 
 /**
- * Marks a Spring bean as a HAPI interceptor the FHIR REST servlets should register, so that a
- * module other than this one can add a cross-cutting concern to the FHIR API -- authorization,
- * audit, consent. The bean carries HAPI's own {@code Hook} methods; this annotation only says it
- * should be picked up, in the way {@link R4Provider} says so for a resource provider.
+ * Marks a Spring bean as a HAPI interceptor the FHIR REST servlets should register, so a module
+ * other than this one can add a cross-cutting concern -- authorization, audit, consent -- in the
+ * way {@link R4Provider} marks a resource provider.
  * <p>
- * The bean must carry at least one {@code Hook} method. HAPI refuses one that does not, logging
- * "Interceptor registered with no valid hooks" and continuing, so an annotated bean without hooks
- * is absent from the request path with nothing but a warning to say so.
+ * The bean must carry at least one {@code Hook} method: HAPI logs "Interceptor registered with no
+ * valid hooks" and drops one that does not, so it is absent from the request path with only a
+ * warning to say so.
  * <p>
- * Registration order is only HAPI's tie-break: on each hook it sorts by
- * {@code Interceptor(order = ...)}, which {@code Hook(order = ...)} overrides per method when
- * non-zero. Every interceptor this module owns leaves that at the default 0, so a contributed bean
- * with a negative order runs ahead of them -- authentication included.
+ * Registration order is only HAPI's tie-break -- it sorts each hook by {@code Interceptor(order)},
+ * which a non-zero {@code Hook(order)} overrides per method. This module's own interceptors all
+ * leave that at 0, so a contributed bean with a negative order runs ahead of them, authentication
+ * included.
  * <p>
- * They are registered on every version-specific servlet, so an interceptor meant for one FHIR
- * version has to work out which it is serving. On a hook handed {@code RequestDetails} that is
- * {@code getFhirContext().getVersion().getVersion()}. On
- * {@code Pointcut.SERVER_INCOMING_REQUEST_PRE_PROCESSED} there is no such handle -- it is passed
- * only the request and the response -- so the servlet path is the only discriminator.
+ * Every version-specific servlet registers it, so one meant for a single FHIR version must work out
+ * which it is serving: {@code getFhirContext().getVersion().getVersion()} on a hook handed
+ * {@code RequestDetails}, and on {@code Pointcut.SERVER_INCOMING_REQUEST_PRE_PROCESSED}, which is
+ * passed only the request and the response, the servlet path.
  */
 @Target({ ElementType.CONSTRUCTOR, ElementType.FIELD, ElementType.METHOD, ElementType.TYPE })
 @Retention(RetentionPolicy.RUNTIME)
--- a/api/src/main/java/org/openmrs/module/fhir2/api/annotations/FhirInterceptor.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/annotations/FhirInterceptor.java
@@ -29,8 +29,8 @@ import org.springframework.beans.factory.annotation.Qualifier;
  * silently stopped.
  * <p>
  * The bean must carry at least one {@code Hook} method. HAPI refuses one that does not, logging
- * "Interceptor registered with no valid hooks" and continuing, so an annotated bean without hooks is
- * absent from the request path with nothing but a warning to say so.
+ * "Interceptor registered with no valid hooks" and continuing, so an annotated bean without hooks
+ * is absent from the request path with nothing but a warning to say so.
  * <p>
  * Beans found here are registered after the interceptors this module owns, so authentication has
  * already run. Order among contributed interceptors is HAPI's own, from
--- a/api/src/main/java/org/openmrs/module/fhir2/api/annotations/FhirInterceptor.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/annotations/FhirInterceptor.java
@@ -18,13 +18,9 @@ import java.lang.annotation.Target;
 import org.springframework.beans.factory.annotation.Qualifier;
 
 /**
- * Marks a Spring bean as a HAPI interceptor every FHIR REST servlet registers, the way
- * {@link R4Provider} marks a resource provider.
- * <p>
- * Carry at least one {@code Hook} method; HAPI drops an interceptor without one.
- * <p>
- * On a hook handed no {@code RequestDetails}, read the FHIR version from
- * {@code FhirVersionUtils.getFhirVersion(request)} rather than the servlet path.
+ * Marks a Spring bean as a HAPI interceptor the FHIR REST servlets should register, so a module
+ * other than this one can add a cross-cutting concern -- authorization, audit, consent. It must
+ * declare a {@code Hook} method or HAPI drops it.
  */
 @Target({ ElementType.CONSTRUCTOR, ElementType.FIELD, ElementType.METHOD, ElementType.TYPE })
 @Retention(RetentionPolicy.RUNTIME)
--- a/api/src/main/java/org/openmrs/module/fhir2/api/annotations/FhirInterceptor.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/annotations/FhirInterceptor.java
@@ -25,18 +25,26 @@ import org.springframework.beans.factory.annotation.Qualifier;
  * <p>
  * Registering one from outside used not to last. The servlet unregisters every interceptor when the
  * module context refreshes, and that happens whenever any OpenMRS module starts or stops, so an
- * interceptor added at runtime worked until the next module event and then silently stopped.
+ * interceptor added from a Spring bean or from a module's {@code started()} worked until the next
+ * module event and then silently stopped. A module that re-registered from its own
+ * {@code contextRefreshed()} did survive, being called after this one's.
  * <p>
  * The bean must carry at least one {@code Hook} method. HAPI refuses one that does not, logging
  * "Interceptor registered with no valid hooks" and continuing, so an annotated bean without hooks
  * is absent from the request path with nothing but a warning to say so.
  * <p>
  * Beans found here are registered after the interceptors this module owns, but registration order
- * is only HAPI's tie-break. On any one hook it sorts by {@code Interceptor(order = ...)} first, and
- * every interceptor this module owns leaves that at its default of 0 -- so a contributed bean
- * declaring a negative order runs before them on the hooks it shares with them, authentication
- * among those. They are registered on every version-specific servlet, so an interceptor meant for
- * one FHIR version should check the version it is handed.
+ * is only HAPI's tie-break. On any one hook it sorts by {@code Interceptor(order = ...)} first --
+ * or {@code Hook(order = ...)}, which overrides it per method -- and every interceptor this module
+ * owns leaves that at its default of 0, so a contributed bean declaring a negative order runs
+ * before them on the hooks it shares with them, authentication among those.
+ * <p>
+ * They are registered on every version-specific servlet, so an interceptor meant for one FHIR
+ * version has to work out which it is serving. On a hook handed {@code RequestDetails} that is
+ * {@code getFhirContext().getVersion().getVersion()}. On
+ * {@code Pointcut.SERVER_INCOMING_REQUEST_PRE_PROCESSED} -- the earliest, and the one this module's
+ * own authentication interceptor uses -- there is no such handle: it is passed only the request and
+ * the response, so the servlet path is the only discriminator.
  */
 @Target({ ElementType.CONSTRUCTOR, ElementType.FIELD, ElementType.METHOD, ElementType.TYPE })
 @Retention(RetentionPolicy.RUNTIME)
--- a/api/src/main/java/org/openmrs/module/fhir2/api/annotations/FhirInterceptor.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/annotations/FhirInterceptor.java
@@ -23,28 +23,20 @@ import org.springframework.beans.factory.annotation.Qualifier;
  * audit, consent. The bean carries HAPI's own {@code Hook} methods; this annotation only says it
  * should be picked up, in the way {@link R4Provider} says so for a resource provider.
  * <p>
- * Registering one from outside used not to last. The servlet unregisters every interceptor when the
- * module context refreshes, and that happens whenever any OpenMRS module starts or stops, so an
- * interceptor added from a Spring bean or from a module's {@code started()} worked until the next
- * module event and then silently stopped. A module that re-registered from its own
- * {@code contextRefreshed()} did survive, being called after this one's.
- * <p>
  * The bean must carry at least one {@code Hook} method. HAPI refuses one that does not, logging
  * "Interceptor registered with no valid hooks" and continuing, so an annotated bean without hooks
  * is absent from the request path with nothing but a warning to say so.
  * <p>
- * Beans found here are registered after the interceptors this module owns, but registration order
- * is only HAPI's tie-break. On any one hook it sorts by {@code Interceptor(order = ...)} first --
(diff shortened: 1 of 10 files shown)


api/src/main/java/org/openmrs/module/fhir2/api/annotations/FhirInterceptor.java
```diff
@@ -0,0 +1,46 @@
...
+ * leave that at 0, so a contributed bean with a negative order runs ahead of them, authentication
+ * included.
+ * <p>
+ * Every version-specific servlet registers it, so one meant for a single FHIR version must work out
+ * which it is serving: {@code getFhirContext().getVersion().getVersion()} on a hook handed
+ * {@code RequestDetails}, and on {@code Pointcut.SERVER_INCOMING_REQUEST_PRE_PROCESSED}, which is
+ * passed only the request and the response,
+ * {@code org.openmrs.module.fhir2.web.util.FhirVersionUtils.getFhirVersion(request)}.
+ */
+@Target({ ElementType.CONSTRUCTOR, ElementType.FIELD, ElementType.METHOD, ElementType.TYPE })
```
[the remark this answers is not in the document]
mherman22: I don't think this holds — checked against Spring 5.3.30, which is what this module resolves.

`getBeansWithAnnotation` is `getBeanNamesForAnnotation` plus a `getBean` per hit, and that resolves through `findMergedAnnotationOnBean` (`DefaultListableBeanFactory.java:748-781`), which looks in three places rather than only at the bean type:

1. the bean type, `SearchStrategy.TYPE_HIERARCHY`;
2. the raw bean class, for the proxy case;
3. `bd.getResolvedFactoryMethod()` — Spring's own comment on that branch is *"Check annotations declared on factory method, if any."*

The third branch is exactly the `@Bean`-method case.

Ran it rather than only reading it — a `@Configuration` with a `@Bean @FhirInterceptor` method returning a class that deliberately does **not** carry the annotation on its type:

```
spring version              : 5.3.30
bean class carries @FhirInterceptor? false
getBeanNamesForAnnotation   : [viaBeanMethod]
getBeansWithAnnotation keys : [viaBeanMethod]
resolved instance type      : ...$ContributedByFactoryMethod
```

The bean is discovered by the method annotation alone, so it is registered rather than silently omitted.

Your summary says the validation harness could not run — no `mvn`, `java` or `javac` in that environment — so this reads as inferred rather than measured. If you can run it and get a different result I'd like to see it.

One part of the underlying instinct is fair, though: `@Target` also allows `FIELD` and `CONSTRUCTOR`, and those genuinely are inert here, because nothing injects by this qualifier — it is only ever scanned for. The target set is copied verbatim from the existing `@R3Provider`/`@R4Provider`, and I kept it consistent with them rather than narrowing one of the three annotations in the package.

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
```diff
@@ -171,6 +184,132 @@ protected void initialize() {
...
+	 */
+	private void registerContributedInterceptors(List<Object> current) {
+		ConfigurableApplicationContext ctx = getModuleApplicationContext();
+		if (ctx == null) {
+			return;
+		}
+		
+		Map<String, Object> contributed;
+		try {
+			contributed = ctx.getBeansWithAnnotation(FhirInterceptor.class);
```
mherman22: This needs fixing before merge. `getBeansWithAnnotation` resolves every annotated bean in a single call, so it throws as a whole when Spring cannot supply any one of them, and the catch below then returns having registered nothing. One module contributing a bean Spring can't build at refresh time (a lazy bean whose constructor throws, or one in a scope that isn't active while a context refresh runs) silently leaves every *other* module's contributed interceptor unregistered, SMART on FHIR scope enforcement included, while the servlet carries on answering 200s. The only trace is the one ERROR line.

I ran this on the branch: a context holding a lazy `@FhirInterceptor` bean whose constructor throws plus a healthy one, then `registerInterceptors()`, and the healthy bean is absent from `getAllRegisteredInterceptors()` while all five built-ins are there.

It also isn't what the javadoc on `registerContributedInterceptors` promises, that "one that cannot be built or scanned is reported and skipped". Scanning is isolated per bean; building is not.

`ctx.getBeanNamesForAnnotation(FhirInterceptor.class)` returns both names without throwing (I checked that on the same fixture), so switching the lookup to that and moving `ctx.getBean(beanName)` inside the per-bean `try` gets the isolation the javadoc describes. The catch at the bottom of the loop then needs to stop dereferencing `interceptor`, since a bean that failed to build hasn't got one.

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
```diff
@@ -171,6 +184,132 @@ protected void initialize() {
...
+	 * contributes with {@link FhirInterceptor}. Safe to call on a servlet that is already serving,
+	 * which is what {@link #refreshed()} does - the new set is registered before the previous one is
+	 * dropped, so a concurrent request never finds the servlet without
+	 * {@link RequireAuthenticationInterceptor}.
+	 * <p>
+	 * Synchronized because it reads and rewrites {@code registeredInterceptors} around calls that
+	 * mutate HAPI's registry: {@link #initialize()} runs on the container's init thread and
+	 * {@link #refreshed()} on whichever thread drove the context refresh.
+	 */
+	protected synchronized void registerInterceptors() {
```
mherman22: This can't merge as it stands. SonarCloud raises S2654, "JEE applications should not use threads or synchronization", on this method as a CRITICAL bug, and that single issue is the whole of the D reliability rating on new code that fails the quality gate; the PR's merge state is currently BLOCKED. The last twelve PRs merged into master all carried SonarCloud Code Analysis green, so this isn't a check the project waves through.

The javadoc here makes a fair case for the lock, so I'd keep the mutual exclusion and add `@SuppressWarnings("java:S2654")` to the method rather than trade thread safety for a rating. Nothing in the repo suppresses a Sonar rule today, so if the annotation turns out not to take effect in this setup, the fallback is marking the issue Won't Fix in SonarCloud.

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
```diff
@@ -171,6 +184,132 @@ protected void initialize() {
...
+		for (Map.Entry<String, Object> entry : contributed.entrySet()) {
+			String beanName = entry.getKey();
+			Object interceptor = entry.getValue();
+			
+			try {
+				String orderViolation = describeOrderViolation(interceptor);
+				
+				if (orderViolation != null) {
+					log.error(
+					    "Not registering contributed FHIR interceptor bean {} ({}): {}, which would run it ahead of this "
```
mherman22: The middle clause of this message isn't true, and it's the part a module author will act on. The built-ins sit at -1000, so a contributed interceptor anywhere from -999 to -1 still dispatches after them, authentication included. I registered an `@Interceptor(order = -1)` bean first and a -1000 one second on an `InterceptorService` and drove the pointcut: the -1000 one ran first.

The `>= DEFAULT_ORDER` rule itself is fine, and it's what the annotation's javadoc documents, so I'd leave the check alone and drop the "which would run it ahead of ..." clause, keeping the sentence after it that already states the contract.

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
```diff
@@ -171,6 +184,132 @@ protected void initialize() {
...
+	 * Synchronized because it reads and rewrites {@code registeredInterceptors} around calls that
+	 * mutate HAPI's registry: {@link #initialize()} runs on the container's init thread and
+	 * {@link #refreshed()} on whichever thread drove the context refresh.
+	 */
+	protected synchronized void registerInterceptors() {
+		List<Object> previous = new ArrayList<>(registeredInterceptors);
+		List<Object> current = new ArrayList<>();
+		
+		try {
+			// these carry FhirConstants.BUILT_IN_INTERCEPTOR_ORDER, so they dispatch ahead of a contributed
```
mherman22: `loggingInterceptor` is HAPI's `LoggingInterceptor`, which carries a bare `@Interceptor` and so sits at order 0, not at `BUILT_IN_INTERCEPTOR_ORDER`. It keeps its place ahead of a contributed bean at order 0 only because HAPI's sort is stable and it was registered first, which is the thing this comment says doesn't matter. Worth narrowing the claim to the four interceptors this module owns.

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
mherman22: One bean Spring cannot supply makes this whole call throw, so a single unbuildable `@FhirInterceptor` bean silently drops every other module's interceptor too: use `getBeanNamesForAnnotation` and move `ctx.getBean(beanName)` inside the per-bean `try` before this merges.

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
mherman22: Sonar reads this as S2654 and rates it a critical bug, which is the entire D reliability rating failing the quality gate and holding the merge; keep the lock and add `@SuppressWarnings("java:S2654")`.

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
mherman22: Untrue for -999 to -1, which still dispatch behind the built-ins at -1000, so drop this clause and let the sentence after it state the contract.

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
mherman22: `loggingInterceptor` is HAPI's own and sits at order 0, not `BUILT_IN_INTERCEPTOR_ORDER`, so narrow the claim to the four this module owns.

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
mherman22: I think a bean Spring can't supply here takes the others down with it: the lookup throws as a whole, so every other module's interceptor ends up unregistered too (I hit this with a lazy bean whose constructor throws). Maybe `getBeanNamesForAnnotation` plus `getBean` inside the per-bean `try`, so a bad bean only loses itself?

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
mherman22: I think this is what's failing the quality gate: Sonar rates the `synchronized` as S2654 critical, which is the D rating on new code and is currently blocking the merge. Maybe keep the lock and add `@SuppressWarnings("java:S2654")`, unless you'd rather drop the lock because the lifecycle callbacks are serialised upstream anyway?

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
mherman22: I think the problem here is that `getBeansWithAnnotation(...)` resolves the beans as part of the lookup. If one lazy bean can't be instantiated (its constructor throws, say), the lookup itself fails and none of the other annotated beans get registered. That bites hardest on a refresh: the previous set is unregistered earlier in the rebuild, so a contributed interceptor that was working beforehand is removed and never comes back. The built-ins survive, but every contributed interceptor goes, whichever module it came from. Calling `getBeanNamesForAnnotation(...)` first and resolving each bean inside its own try/catch would isolate the failure, so a broken bean only loses itself.

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
mherman22: I think this is what's failing the quality gate: Sonar rates the `synchronized` as S2654 critical, which is the D rating on new code and is currently blocking the merge. Maybe keep the lock and add `@SuppressWarnings("java:S2654")`, unless you'd rather drop the lock because the lifecycle callbacks are serialised upstream anyway?

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
mherman22: I think the problem here is that `getBeansWithAnnotation(...)` resolves the beans as part of the lookup. If one lazy bean can't be instantiated (its constructor throws, say) the lookup itself fails, and none of the other annotated beans get registered. That bites hardest on a refresh: the previous set is unregistered earlier in the rebuild, so a contributed interceptor that was working beforehand is removed and never comes back. The built-ins survive, but every contributed interceptor goes, whichever module it came from.

I checked it on a running server with a healthy contributed bean alongside one that can't be built: as it stands, every FHIR response comes back without the healthy interceptor and the only trace is `Could not read the contributed FHIR interceptors from the Spring context; none of them will run`. Swapping in a per-name lookup brings it back, and swapping the old omod in again removes it, so it's that call and not something else in the environment.

`getBeanNamesForAnnotation(...)` first, then `ctx.getBean(beanName)` inside the per-bean `try`, would isolate it so a broken bean only loses itself. The catch below would need to stop dereferencing `interceptor`, since a bean that failed to build hasn't got one.

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
```diff
@@ -171,6 +184,132 @@ protected void initialize() {
...
+				registerInterceptor(interceptor);
+				current.add(interceptor);
+			}
+			
+			// drop the previous set only now that its replacement is in place, and only the entries the new
+			// set did not carry over by identity - unregistering one HAPI has kept would leave a hole
+			Set<Object> retained = Collections.newSetFromMap(new IdentityHashMap<>());
+			retained.addAll(current);
+			previous.stream().filter(interceptor -> !retained.contains(interceptor))
+			        .forEach(getInterceptorService()::unregisterInterceptor);
```
mherman22: I think there's a narrower version of the same thing here, and this one doesn't need a broken bean. The previous set is dropped on these two lines, but the contributed beans aren't put back until `registerContributedInterceptors` below, so on every refresh the servlet spends that whole call with authentication registered and the contributed hooks absent. The window is one `ctx.getBean` per contributed bean, so it covers bean construction, and `service()` isn't synchronised, so it's concurrent with request handling.

In the distro that's `SmartScopeInterceptor`, so a request landing in the window is a normal 200 with the scopes unenforced, and nothing throws or logs.

Registering the contributed beans before this unregister closes it, though it needs a second piece: a refresh that reuses the Spring context hands back the same singleton, HAPI's `registerInterceptor` returns false for an instance it already holds, and it would then be dropped from `current` and unregistered here. Treating already-registered as retained fixes that, and it also stops the false "declares no @Hook method" error for it. I measured the gap with an InterceptorService that snapshots the surviving interceptors at every unregister, and it goes away with the reorder.

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
mherman22: I think this is what's failing the quality gate: Sonar rates the `synchronized` as S2654 critical, which is the D rating on new code and is currently blocking the merge. Maybe keep the lock and add `@SuppressWarnings("java:S2654")`, unless you'd rather drop the lock because the lifecycle callbacks are serialised upstream anyway?

omod/src/main/java/org/openmrs/module/fhir2/web/servlet/FhirRestServlet.java
mherman22: I think the problem here is that `getBeansWithAnnotation(...)` resolves the beans as part of the lookup. If one lazy bean can't be instantiated (its constructor throws, say) the lookup itself fails, and none of the other annotated beans get registered. That bites hardest on a refresh: the previous set is unregistered earlier in the rebuild, so a contributed interceptor that was working beforehand is removed and never comes back. The built-ins survive, but every contributed interceptor goes, whichever module it came from.

I checked it on a running server with [a throwaway module](https://github.com/mherman22/openmrs-module-fhir2verify) that contributes a healthy `@FhirInterceptor` bean alongside one that can't be built. The healthy bean stamps a response header, so whether it is registered is visible on the wire rather than only in a log. Every FHIR response comes back without that header, and the only trace is `Could not read the contributed FHIR interceptors from the Spring context; none of them will run`. Swapping in a per-name lookup brings it back and swapping the old build in again removes it, so it's that call rather than something else in the environment. On the distro the same shape applies to `SmartScopeInterceptor`: a module whose interceptor can't be built would take another module's scope enforcement down with it, and the request still returns 200.

`getBeanNamesForAnnotation(...)` first, then `ctx.getBean(beanName)` inside the per-bean `try`, would isolate it so a broken bean only loses itself. The catch below would need to stop dereferencing `interceptor`, since a bean that failed to build hasn't got one. It would also make the javadoc above true: "cannot be built or scanned is reported and skipped" currently holds for the scanning half only.

mherman22, line 242: @ibacher, I think this resolves each bean as part of the lookup, so what happens to a `@Lazy` bean that can't be instantiated? Reading it, the whole call throws and none of the other annotated beans get registered, so one broken bean would take every other module's contributed interceptor with it. I poked at it with [a throwaway module](https://github.com/mherman22/openmrs-module-fhir2verify) and that's what I saw. Would `getBeanNamesForAnnotation(...)` first, then `ctx.getBean(beanName)` inside the per-bean `try`, be the shape you'd want?

ibacher (approved):

PR #629, opened 2026-08-27
FM2-700: a Task reference filter finds an id stored without a resource-type prefix

dkayiwa:
https://openmrs.atlassian.net/browse/FM2-700

## The problem

`GET /ws/fhir2/R4/Task?based-on=ServiceRequest/<uuid>` returns `"total": 0` while the unfiltered
`GET /Task` returns a Task whose `basedOn` is that ServiceRequest.

Reproduced on a reference application 3.7.1 standalone running `fhir2-4.2.0.omod`:

```
GET /Task?based-on=ServiceRequest/d9abe171-5444-4bfd-a252-5e52405b8332  ->  "total": 0
GET /Task?_summary=count                                               ->  "total": 331
GET /Task/6ec73b25-d984-4f95-9d4d-b651ff29def0                         ->  basedOn d9abe171-...
```

## Cause

`FhirTaskDaoImpl#handleReference` matched the search id against `fhir_reference.target_uuid` only.
`FhirReference#setReference` writes that column with the id it can parse out of the reference
string, and a bare resource id yields nothing to parse — so `target_uuid` stays null whenever a
client sends `Reference.type` alongside a bare id. On that install every one of the 331
`fhir_reference` rows is exactly that shape: `target_type='ServiceRequest'`,
`reference='<bare uuid>'`, `target_uuid=NULL`. `target_uuid` has one consumer in the module, that
predicate, so the filter could not match anything.

The predicate moved from `reference` to `targetUuid` in a2bc7a9a (FM2-682, #607), which is tagged
4.1.0 and 4.2.0 — `git log -S targetUuid -- .../FhirTaskDaoImpl.java` returns that one commit, and
the JPA-criteria migration before it (ee174cd3) still matched on `reference`. The same commit added
`target_uuid="<the bare reference>"` to every `fhir_reference` row in both shared fixtures, which is
why the existing search tests stayed green: with the two columns equal, the two predicates are
indistinguishable on that data.

`based-on`, `owner`, `subject` and `focus` all share `handleReference`, so all of them were
affected; the R3 provider routes its own reference params (`based-on`, `owner`, `subject`) through
the same method.

## The change

Match the id against either column, still scoped by the stored type:

```java
(bo.targetUuid = :idPart OR bo.reference = :idPart) AND bo.type = :resourceType
```

A strict superset of the previous predicate, so everything `target_uuid` was introduced to serve
keeps resolving through it, and ids stored bare resolve through `reference` as they did before
4.1.0. No schema change and no data migration — existing rows work as they are.

## Verification

Verified on the live 3.7.1 standalone, against the same 331-row database that returned 0, after
swapping in a build of this branch:

| request | on 4.2.0 | on this branch |
| --- | --- | --- |
| `?based-on=ServiceRequest/d9abe171-…` | 0 | **1** (Task `6ec73b25-…`) |
| `?based-on:ServiceRequest=d9abe171-…` | 0 | **1** |
| `?based-on=ServiceRequest/00000000-…` (unknown id) | 0 | 0 |
| `?based-on=Patient/d9abe171-…` (wrong type) | 0 | 0 |
| `?_summary=count` | 331 | 331 |

In the test suite:

- New fixture `FhirTaskDaoImplTest_bare_reference_data.xml` whose `fhir_reference` rows have **no**
  `target_uuid` — the shape measured on that install, and one no existing fixture has.
- `TaskSearchQueryTest`: `based-on` and `owner` searches over that fixture, plus two cases that
  search an id under a different resource type and expect no match — one over a row whose
  `target_uuid` is null, one over a row where it is populated.
- `TaskFhirResourceIntegrationTest` (R4): POSTs a Task whose `basedOn` carries `type` plus a bare id,
  asserts it really was stored unprefixed, then issues the ticket's own request over REST.
- Every clause of the predicate is discriminated — mutate a clause and read the failures.
  - Drop the `reference` arm: `TaskSearchQueryTest` fails 2 of 37.
  - Drop the `targetUuid` arm: the api module fails 1 of 4350,
    `FhirTaskDaoImplTest.searchForTasks_shouldReturnTasksByFocusReference`, which saves
    `Observation/<uuid>` programmatically and is therefore the only search in the module that
    reaches a row where `target_uuid` and `reference` differ.
  - Scope `type` to the `targetUuid` arm alone: both wrong-resource-type cases fail. Scope it to the
    `reference` arm alone: `searchForTasks_shouldNotReturnTasksWhenATargetUuidMatchIsSearchedUnderAnotherResourceType` fails.
  - Swap `getIdPart()` for `getValue()`: `TaskSearchQueryTest` fails 2 of 37. This is caught in the
    api module, not only over REST, because the new cases build their `ReferenceParam` through the
    single-argument constructor, which leaves `getValue()` as `ServiceRequest/<id>`, rather than
    through `setValue`, which re-parses it to a bare id and makes the two indistinguishable.
- Root `mvn clean install`: green apart from `ObservationNarrativeTest`, which fails identically on
  an unmodified `origin/master` checkout on this machine — a JDK locale renders `pm` where the
  expectation says `PM`.

## Not fixed here

A reference parameter with no resource type at all — `GET /Task?based-on=<bare uuid>` — is dropped
rather than applied, because `validReferenceParam` requires a non-null resource type and
`handleReference` then contributes no predicate, so the search returns every Task (measured: 331 of
331 on that install). The ticket reproduces with the typed form, so this is left alone here and
deserves its own issue.

### DIFF
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
@@ -118,7 +118,7 @@ public class FhirTaskDaoImpl extends BaseFhirDao<FhirTask> implements FhirTaskDa
 				List<Optional<? extends Predicate>> predicateList = new ArrayList<>();
 				// reference holds the reference string as the client wrote it; targetUuid holds the id
 				// parsed out of that string, which is null when the client wrote a bare id. Match the
-				// id against both columns. FM2-700.
+				// id against both columns.
 				predicateList.add(Optional.of(criteriaContext.getCriteriaBuilder().or(
 				    criteriaContext.getCriteriaBuilder().equal(taskAliasJoin.get("targetUuid"), param.getIdPart()),
 				    criteriaContext.getCriteriaBuilder().equal(taskAliasJoin.get("reference"), param.getIdPart()))));
--- a/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
@@ -116,8 +116,12 @@ public class FhirTaskDaoImpl extends BaseFhirDao<FhirTask> implements FhirTaskDa
 				Join<?, ?> taskAliasJoin = criteriaContext.addJoin(property, alias);
 				
 				List<Optional<? extends Predicate>> predicateList = new ArrayList<>();
-				predicateList.add(Optional
-				        .of(criteriaContext.getCriteriaBuilder().equal(taskAliasJoin.get("targetUuid"), param.getIdPart())));
+				// reference holds the reference string as the client wrote it; targetUuid holds the id
+				// parsed out of that string, which is null when the client wrote a bare id. Match the
+				// id against both columns. FM2-700.
+				predicateList.add(Optional.of(criteriaContext.getCriteriaBuilder().or(
+				    criteriaContext.getCriteriaBuilder().equal(taskAliasJoin.get("targetUuid"), param.getIdPart()),
+				    criteriaContext.getCriteriaBuilder().equal(taskAliasJoin.get("reference"), param.getIdPart()))));
 				predicateList.add(Optional
 				        .of(criteriaContext.getCriteriaBuilder().equal(taskAliasJoin.get("type"), param.getResourceType())));
 				return Optional.of(criteriaContext.getCriteriaBuilder().and(toCriteriaArray(predicateList)));
--- a/api/src/test/java/org/openmrs/module/fhir2/api/search/TaskSearchQueryTest.java
+++ b/api/src/test/java/org/openmrs/module/fhir2/api/search/TaskSearchQueryTest.java
@@ -66,7 +66,7 @@ public class TaskSearchQueryTest extends BaseFhirContextSensitiveTest {
 	
 	private static final String BARE_REFERENCE_TASK_UUID = "2ba49b19-6d51-4de9-8bb0-b2d3c86ab0a8";
 	
-	private static final String BARE_REFERENCE_ORDER_UUID = "6d0ae116-707a-4629-9850-f15206e63ab0";
+	private static final String BARE_REFERENCE_ORDER_UUID = "05caaf38-35a1-4c06-a229-44b4ea608b34";
 	
 	private static final String BARE_REFERENCE_PRACTITIONER_UUID = "8f1a6b47-1f0d-4a5e-8f36-3f5f2c9a4d21";
 	
@@ -200,8 +200,8 @@ public class TaskSearchQueryTest extends BaseFhirContextSensitiveTest {
 	public void searchForTasks_shouldReturnTasksByBasedOnWhenTheStoredReferenceHasNoResourceTypePrefix() throws Exception {
 		executeDataSet(TASK_DATA_BARE_REFERENCE_XML);
 		
-		ReferenceParam basedOnReference = new ReferenceParam();
-		basedOnReference.setValue(FhirConstants.SERVICE_REQUEST + "/" + BARE_REFERENCE_ORDER_UUID);
+		ReferenceParam basedOnReference = new ReferenceParam(
+		        FhirConstants.SERVICE_REQUEST + "/" + BARE_REFERENCE_ORDER_UUID);
 		
 		ReferenceAndListParam ref = new ReferenceAndListParam().addAnd(new ReferenceOrListParam().add(basedOnReference));
 		
@@ -223,8 +223,7 @@ public class TaskSearchQueryTest extends BaseFhirContextSensitiveTest {
 	        throws Exception {
 		executeDataSet(TASK_DATA_BARE_REFERENCE_XML);
 		
-		ReferenceParam wrongType = new ReferenceParam();
-		wrongType.setValue(FhirConstants.PATIENT + "/" + BARE_REFERENCE_ORDER_UUID);
+		ReferenceParam wrongType = new ReferenceParam(FhirConstants.PATIENT + "/" + BARE_REFERENCE_ORDER_UUID);
 		
 		ReferenceAndListParam ref = new ReferenceAndListParam().addAnd(new ReferenceOrListParam().add(wrongType));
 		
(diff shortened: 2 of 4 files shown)


api/src/test/java/org/openmrs/module/fhir2/api/search/TaskSearchQueryTest.java
```diff
@@ -183,6 +191,49 @@ public void searchForTasks_shouldReturnTasksByBasedOn() {
...
+	 * A reference whose id is stored without a "Type/" prefix - the shape written whenever a client
+	 * sends Reference.type alongside a bare resource id - leaves fhir_reference.target_uuid unset.
+	 * Searching by such a reference must still find the Task. See FM2-700.
+	 */
+	@Test
+	public void searchForTasks_shouldReturnTasksByBasedOnWhenTheStoredReferenceHasNoResourceTypePrefix() throws Exception {
+		executeDataSet(TASK_DATA_BARE_REFERENCE_XML);
+		
+		ReferenceParam basedOnReference = new ReferenceParam();
+		basedOnReference.setValue(FhirConstants.SERVICE_REQUEST + "/" + BARE_REFERENCE_ORDER_UUID);
```
dkayiwa: `setValue("ServiceRequest/<uuid>")` re-parses the string, so `getValue()` here comes back as the bare uuid and is indistinguishable from `getIdPart()`. Over HTTP the two differ: HAPI builds the param through `setValueAsQueryToken`, which leaves `getValue()` as the full `ServiceRequest/<uuid>`. The single-argument constructor gives you that same shape.

```suggestion
		ReferenceParam basedOnReference = new ReferenceParam(
		        FhirConstants.SERVICE_REQUEST + "/" + BARE_REFERENCE_ORDER_UUID);
```

Why it's worth changing: swap `getIdPart()` for `getValue()` in the new predicate and the entire api module still passes, all 4350 tests. The only test that catches the swap is `TaskFhirResourceIntegrationTest.shouldSearchTasksByBasedOnReferenceStoredWithoutAResourceTypePrefixAsJson`, and that class doesn't run in CI: surefire in the root pom excludes `**/*IntegrationTest`, failsafe is bound only inside the `integration-test` profile, and the workflow runs a plain `clean install`. The class name appears nowhere in this PR's build log, while `TaskSearchQueryTest` does. With these two lines changed the DAO test catches the swap as well, inside the gate that actually runs.

I ran the integration tests by hand (`-Dtest=TaskFhirResourceIntegrationTest`, r3 and r4, 54 green), so nothing is broken as it stands and this is optional.

test-data/src/main/resources/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImplTest_bare_reference_data.xml
```diff
@@ -0,0 +1,23 @@
...
+    graphic logo is a trademark of OpenMRS Inc.
+-->
+<dataset>
+	<!--
+		A reference stored the way a client that sends Reference.type alongside a bare resource id
+		writes one: the type lands in target_type and the id in reference, and target_uuid is left
+		unset because FhirReference#setReference could not parse a resource id out of a bare id.
+		target_uuid is deliberately absent from these rows - do not add it.
+	-->
+	<fhir_reference reference_id="300" name="ServiceRequest/6d0ae116-707a-4629-9850-f15206e63ab0" target_type="ServiceRequest" reference="6d0ae116-707a-4629-9850-f15206e63ab0" uuid="3f3ecb0a-2ca2-4e21-a5cd-bb1af1ad1e94" creator="1" date_created="2026-01-01 00:00:00.0" retired="false"/>
```
dkayiwa: `6d0ae116-707a-4629-9850-f15206e63ab0` is already spoken for: it's the `drug_order` at `orders.order_id=1005` in `FhirMedicationRequestDaoImpl_initial_data.xml`, and three other test classes hold it as their MedicationRequest / DrugOrder uuid. Nothing loads both fixtures in one test today, so it breaks nothing, but this row does declare a known MedicationRequest id as a `ServiceRequest`. A uuid nobody else has claimed would save the next reader the double take.

api/src/test/java/org/openmrs/module/fhir2/api/search/TaskSearchQueryTest.java
dkayiwa: Taken, and confirmed in the HAPI 5.7.9 source: `setValue` splits its argument into a `:ServiceRequest` qualifier plus a bare id before calling `setValueAsQueryToken`, so `myValue` and `myIdPart` both end up as the bare id; the single-argument constructor passes the string through with no qualifier and `myValue` stays `ServiceRequest/<id>`, which is the shape the HTTP path produces.

Applied to all four new cases in this class, not just this one, so they all describe the request a client actually sends.

The discrimination argument holds under measurement: with `getIdPart()` mutated to `getValue()` in the predicate, `TaskSearchQueryTest` now fails 2 of 37, where before the whole api module stayed green.

Fixed in 016331d2. Verified: `TaskSearchQueryTest` 37/37, the api module 4350/4350, `TaskFhirResourceIntegrationTest` r3 + r4 54/54, `spotless:check` clean.

test-data/src/main/resources/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImplTest_bare_reference_data.xml
dkayiwa: You're right — `6d0ae116-707a-4629-9850-f15206e63ab0` is the `drug_order` at `orders.order_id=1005` in `FhirMedicationRequestDaoImpl_initial_data.xml`, and four other classes hold it as their MedicationRequest / DrugOrder uuid. Nothing loads both fixtures together, but the row did declare a known MedicationRequest id as a `ServiceRequest`.

Replaced with `05caaf38-35a1-4c06-a229-44b4ea608b34`, which no other fixture or test claims anywhere in the repo, in both this fixture and `TaskFhirResourceIntegrationTest`.

Fixed in 016331d2.

dkayiwa (commented):
The verification bullet still says `getIdPart()` vs `getValue()` is caught by the integration test, with no unit test able to catch it because `ReferenceParam.setValue` re-parses. 016331d2 changed that: with the four new cases built through the single-argument constructor, swapping `getIdPart()` for `getValue()` in the predicate now fails `TaskSearchQueryTest` 2 of 37 (`searchForTasks_shouldReturnTasksByBasedOnWhenTheStoredReferenceHasNoResourceTypePrefix` and `searchForTasks_shouldReturnTasksByOwnerWhenTheStoredReferenceHasNoResourceTypePrefix`), under surefire, which is the gate CI actually runs.

As it stands the bullet tells the next reader the opposite of what the suite now guards, and it contradicts the reply already on that thread. Worth an edit to the description; it does not hold up the merge.

api/src/main/java/org/openmrs/module/fhir2/api/dao/impl/FhirTaskDaoImpl.java
```diff
@@ -116,8 +116,12 @@ private <U> Optional<Predicate> handleReference(@Nonnull OpenmrsFhirCriteriaCont
 				Join<?, ?> taskAliasJoin = criteriaContext.addJoin(property, alias);
 				
 				List<Optional<? extends Predicate>> predicateList = new ArrayList<>();
-				predicateList.add(Optional
-				        .of(criteriaContext.getCriteriaBuilder().equal(taskAliasJoin.get("targetUuid"), param.getIdPart())));
+				// reference holds the reference string as the client wrote it; targetUuid holds the id
+				// parsed out of that string, which is null when the client wrote a bare id. Match the
+				// id against both columns. FM2-700.
```
ibacher:
```suggestion
				// reference holds the reference string as the client wrote it; targetUuid holds the id
				// parsed out of that string, which is null when the client wrote a bare id. Match the
				// id against both columns.
```

ibacher (approved):
Other than some comment clean-up LGTM.

PR #571, opened 2025-07-23
FM2-667: Add Diagnosis resource to condition fhir resource provider

parthfloyd:
## Description of what I changed
Added DiagnosisTranslator and integrated with ConditionFhirResourceProvider.

Overall adding a tag search parameter to differentiate between Condition and diagnosis in Condition(FHIR) requests.

## Issue I worked on
[Jira Link: FM2-667](https://openmrs.atlassian.net/browse/FM2-667)
## Checklist: I completed these to help reviewers :)
- [x] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

  No? Unsure? -> [configure your IDE](https://wiki.openmrs.org/display/docs/How-To+Setup+And+Use+Your+IDE), format the code and add the changes with `git add . && git commit --amend`

- [x] I have **added tests** to cover my changes. (If you refactored
  existing code that was well tested you do not have to add tests)

  No? -> write tests and add them to this commit `git add . && git commit --amend`

- [x] I ran `mvn clean package` right before creating this pull request and
  added all formatting changes to my commit.

  No? -> execute above command

- [x] All new and existing **tests passed**.

  No? -> figure out why and add the fix to your commit. It is your responsibility to make sure your code works.

- [x] My pull request is **based on the latest changes** of the master branch.

  No? Unsure? -> execute command `git pull --rebase upstream master`

PR #630, opened 2026-09-04
Implement extensible FhirResourceHandlers

ibacher:

Prior to this PR, the FHIR2 module divided things into four layers:

1. `ResourceProvider`s, our interface with HAPI
2. `Service`s, which provide the API for `ResourceProvider`s
3. `Translator`s, which translate between OpenMRS and FHIR
4. `DAO`s, which mostly interact with the OpenMRS service layer and the DB for search queries

Along the way, we've implemented a pattern for handling a service backed by up to two different resource types to handle things like mapping both OpenMRS Encounters and Visits to FHIR Encounters. However, there is no easy way to either extend that in an external module or to add another "backer" for a specific FHIR Resource, which makes it harder to customize with the FHIR2 module than would be ideal.

This PR adds a new concept that sits below `Service`s, a `FhirResourceHandler`. A `FhirResourceHandler` implements an interface similar to the previous `FhirService`, but is intended to allow an arbitrary number of `FhirResourceHandler`s to be registered per FHIR Resource. Each `FhirResourceHandler` can produce a FHIR resource in whatever way makes sense to it and the services (which the resource providers interact with) have been updated to delegate to `FhirResourceHandler`s as discussed later. 

A `FhirResourceHandler` has four hooks which define how it interacts with the top-level `FhirService`:

* `getImplicitProfile()` — This is the one thing that every `FhirResourceHandler` _must_ provide. It is the URL identifier of a profile that the `FhirResourceHandler` implements. This profile is intended as a _marker_ profile and is used to provide a "fast-path" for resources that are known to need to go to a specific handler. The idea is that we can use this to distinguish between, e.g., a FHIR Encounter that's really backed by a Visit and one that's backed by an Encounter.
* `canHandle()` — For FHIR Read, Update, Patch, Delete, and Create operations (as well as, where supported, their conditional varieties), this is used to resolve which handler can handle a specific resource if that resource does not have a profile associated with it. `FhirResourceHandler`s are welcome to use whatever detection method they want.
* `acceptsSearch()` — For search operations, if the client does not provide a profile for the `FhirResourceHandler` they expect to answer, we fallback to a "fan-out" style where we run searches through all participating `FhirResourceHandler`s. This is the callback that we use to determine if this `FhirResourceHandler` should participate in the search. For example, with the `FhirEncounterServiceImpl`, we previously fanned out to either or both of the `Encounter` service or `Visit` service depending on the `_tag` property submitted from the search. `acceptsSearch()` is how we implement this mechanism, allowing, e.g., the default `EncounterBackendEncounterResourceHandler` to skip any searches that only targeted visits. Again, `FhirResourceHandler`s are welcome to use whatever checks they like to determine whether they should respond to a search query.
* `search()` — Assuming the `FhirResourceHandler` participates in the search, this is the entry point for actually performing the search. `FhirResourceHandler`s have full control over how they handle the search, but for those defined in this module, the implementation matches the corresponding search they previously did.

All `FhirResourceHandler`s define the order they are tried in using the Spring `@Ordered` annotation, defaulting to Spring's default order level. All `FhirResourceHandler`s implemented by this module are intentionally set to the lowest possible precedence so other `FhirResourceHandler`s will default to overriding them _except in search_ because this is done as a fan-out as described above.

A new base class, `BaseCompositeFhirService`, is provided to handle all of the orchestration between a base `FhirService` and the `FhirResoruceHandler`s that implement it. Consequently, after the PRs, most of these services look like this:

```java
@Component
public class FhirLocationServiceImpl extends BaseCompositeFhirService<Location> implements FhirLocationService {
	
	@Override
	public IBundleProvider searchForLocations(LocationSearchParams locationSearchParams) {
		return doSearch(locationSearchParams.toSearchParameterMap());
	}
}
```

With the only function left being `searchFor<Resource>` instead of `BaseCompositeFhirService`'s `doSearch()`.

By registering a custom `FhirResourceHandler`, a module can provide it's own version of `Encounter`s, `Immunization`s or any other arbitrary FHIR resources where the `ResourceProvider` is backed by a `BaseCompositeFhirService` (as all of the built-in services are).

I've marked this as a breaking change, but it should be backwards-compatible with most uses of this module. If you override a `FhirService` with a custom implementation, your custom implementation continues to control. If you override a `FhirTranslator` with a custom implementation, the new default `FhirResourceHandler` will defer to that translator, etc.

Note that now all FHIR Resources generated by this module are stamped by the implicit profile of `FhirResourceHandler` that generated their contents. The intention is to make it easy for the client to echo this _back_ so that patches or updates get routed to the handler that create the resource, since this is most likely to be successful.

## Issue I worked on

  No? -> execute above command

## Anything else?

I understand that this PR is not easy to review; it's over 20,000 changed lines. Consequently, I've left this in fairly sizeable, but readable commit sizes. Most commits here implement 1 FHIR Resource at a time. There are several commits that implement the baseline functionality those depend on: [754c730](https://github.com/openmrs/openmrs-module-fhir2/pull/630/commits/754c7309bf8426c53d8259c3887c7813a20895be), [6524a3e](https://github.com/openmrs/openmrs-module-fhir2/pull/630/commits/6524a3e8c6cab9c9ff587ccde9b67691e35d4331), [58f0207](https://github.com/openmrs/openmrs-module-fhir2/pull/630/commits/58f0207ae4ac6c1630608d76dc689dea144e2e50), and [fb2992d](https://github.com/openmrs/openmrs-module-fhir2/pull/630/commits/fb2992d97cdeab8fdbb1fbdc2dca48b9fd67be84). Note that commits prior to [fb2992d](https://github.com/openmrs/openmrs-module-fhir2/pull/630/commits/fb2992d97cdeab8fdbb1fbdc2dca48b9fd67be84) were using a more complicated model which I realized duplicated what we could do with the implicit profile. Initially, there was a "backingKey" concept that could be passed to search as a `_tag` to select a single `FhirResourceHandler`, but I realized that FHIR search allows a `_profile`


api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/DiagnosisTranslatorImpl.java
```diff
@@ -206,7 +206,7 @@ public Diagnosis toOpenmrsType(@Nonnull Diagnosis existingDiagnosis, @Nonnull Co
 		
 		// Set rank from extension
 		Optional<Extension> rankExtension = Optional
-		        .ofNullable(condition.getExtensionByUrl(FhirConstants.DIAGNOSIS_RANK_EXTENSION_URI));
+		        .ofNullable(condition.getExtensionByUrl(FhirConstants.OPENMRS_FHIR_EXT_DIAGNOSIS_RANK));
```
mherman22: This one needs sorting before merge. The rank extension URI moves from `http://fhir.openmrs.org/R4/StructureDefinition/diagnosis-rank` to `http://fhir.openmrs.org/ext/diagnosis-rank`, and the old one shipped: it came in with FM2-667 (#571), and the 4.0.0, 4.1.0 and 4.2.0 tags all contain that commit.

This is the only URL the inbound path looks up, so a client POSTing a Condition carrying the shipped extension has its rank silently dropped, and gets a 201 back with no rank and no error. Reads are the mirror image: a client looking for rank at the old URI simply stops finding it. Nothing throws in either direction, so there's no signal until someone notices the data is gone.

`DiagnosisTranslatorImplTest` was retargeted to the new constant in the same commit, so it stays green and doesn't speak to the compatibility question.

Falling back to the old URI on the way in, while only ever emitting the new one, would cost a couple of lines and keep existing clients working. If a clean break is what you want for 5.0.0, it's worth naming in the description, which currently scopes breaking changes to `FhirService` and `FhirTranslator` overrides. Either way it's a wire change with nothing to do with `FhirResourceHandler`s, so it would be far easier to reason about on its own.

api/src/main/java/org/openmrs/module/fhir2/api/impl/BaseCompositeFhirService.java
```diff
@@ -0,0 +1,539 @@
...
+		
+		IBaseMetaType meta = resource.getMeta();
+		if (meta == null) {
+			return;
+		}
+		
+		boolean already = meta.getProfile().stream().map(IPrimitiveType::getValue).filter(java.util.Objects::nonNull)
+		        .anyMatch(url::equals);
+		if (!already) {
+			meta.addProfile(url);
```
mherman22: Every resource the module returns now carries a `meta.profile`, and these URLs don't resolve. `http://fhir.openmrs.org/` answers 200 but `http://fhir.openmrs.org/StructureDefinition/openmrs-encounter` is a 404, and I can't find StructureDefinitions shipped or served anywhere in the module.

Canonical URLs are identifiers rather than locations, so that isn't a spec violation by itself. But a client that validates against declared profiles will start failing on responses it accepted before, and this fires on every resource rather than only where routing actually needed it. Is the expectation that a validating client holds the definitions locally, or is publishing them at those URLs something you'd want to land first?

ibacher:
@claude review

